PR #48, opened 2020-02-05
Fix enum options default not None

### LINKED ISSUE #46: Enum Option default is `None` [BUG]
### Describe the bug

If an enum option type is not given, it's value is `None` instead of the specified default value.  If the option is specified then it works as expected

### To Reproduce

Steps to reproduce the behavior with a minimum self-contained file.

Replace each part with your own scenario:

```Python
from enum import Enum

import typer


class NeuralNetwork(str, Enum):
    simple = "simple"
    conv = "conv"
    lstm = "lstm"


def main(
    network: NeuralNetwork = typer.Option(NeuralNetwork.simple, case_sensitive=False)
):
    typer.echo(f"Training neural network of type: {network.value}")


if __name__ == "__main__":
    typer.run(main)

```

* Call it with:

```bash
python main.py
```

* It outputs:

```
...
  File "tutorial002.py", line 15, in main
    typer.echo(f"Training neural network of type: {network.value}")
AttributeError: 'NoneType' object has no attribute 'value'
```

```
Training neural network of type: simple
```

### Expected behavior
I would expect that the specified default value of an enum would be passed to the function parameter instead of `None` when the option is not provided.

### Environment

* OS: [macOS]
* Typer Version [0.0.8]
* Python version [3.7.6]

sullivancolin:
fixes #46 

* converter checks if value is Enum type and gets enum.value, or is string 
* New test to check default is correctly selected when option not provided

### DIFF
--- a/tests/test_tutorial/test_parameter_types/test_enum/test_tutorial001.py
+++ b/tests/test_tutorial/test_parameter_types/test_enum/test_tutorial001.py
@@ -32,6 +32,12 @@ def test_invalid():
     )
 
 
+def test_no_options_is_default():
+    result = runner.invoke(app, [])
+    assert result.exit_code == 0
+    assert "Training neural network of type: simple" in result.output
+
+
 def test_script():
     result = subprocess.run(
         ["coverage", "run", mod.__file__, "--help"],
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for your contribution @sullivancolin ! :rocket: :cake:

I wasn't able to reproduce the error from the issue with the latest Typer, so I think the underlying bug might have been fixed by Click 7.1, released recently.

Could you confirm?

sullivancolin:
Confirmed fixed by new release of Click! 🎉

tiangolo:
Great! Thanks for reporting back and closing the issue. :+1:

PR #63, opened 2020-03-11
:pencil2: fix typo in progressbar/tutorial004.py

ValentinCalomme:
In the tutorial, the length of the progress bar is set to 1000 and the update was mistakenly set to 2500. The update should be 1/4 of the length given that the loop runs for 4 iterations. This commit sets the update amount to 250 which better showcases the feature.


tiangolo:
Ah! Good catch, thanks @ValentinCalomme ! :rocket: :cake:

PR #73, opened 2020-03-19
[TYPO] Change bashrc to zshrc in example for zsh completion

StephenBrown2:
Looks like there was a typo in the example for `typer-cli`. zsh completion wouldn't be installed in `~/.bashrc`.

### DIFF
--- a/docs/tutorial/package.md
+++ b/docs/tutorial/package.md
@@ -77,7 +77,7 @@ Package operations: 15 installs, 0 updates, 0 removals
   - Installing py (1.8.1)
   - Installing shellingham (1.3.2)
   - Installing wcwidth (0.1.8)
-  - Installing pytest (5.4.1)
+  - Installing pytest (5.3.5)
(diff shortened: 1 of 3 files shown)


tiangolo:
So, it will install completion in a different way depending on your shell, it supports Bash, Zsh, Fish, and PowerShell.

It would be installed in `~/.zshrc` when using Zsh. That's just an example of how it would look like you were using Zsh.

So, I'll pass on this, but thanks for your contribution and effort! :cake:

StephenBrown2:
So, I'm confused then. Is:
```
zsh completion installed in /home/user/.bashrc.
```
correct, then? If so, that's terribly confusing, as my zsh shell does not read `~/.bashrc`, and the completions would be broken.

The pertinant change is in https://github.com/tiangolo/typer/pull/73/commits/eacaaff88db34efd3d67d70aa5675c6f3898f1bb, the rest of it is to get the tests to run successfully.

PR #92, opened 2020-04-26
Make installed completion lazy load

tiangolo:
This enables having a Typer script/program that is only available under a virtual environment that is loaded later, after the shell is already started, and still have completion support there.

This would also improve shell startup performance as it no longer has to call Python during startup, only later when requesting completion.

The way it works is that completion for a program can be installed "globally" for the whole shell, but it won't call the program until a completion is requested. This also means that it would be even possible to have different versions of a program under different environments (or 2 programs with the same name), and after activating the environment, when triggering a completion for that program name, the one inside the environment will be the one used to request the completion.


tiangolo:
This should fix / related to https://github.com/tiangolo/typer/issues/75

Issue #94, opened 2020-04-30
[FEATURE] Support datetime.date in addition to datetime.datetime


lbjarre:
### Is your feature request related to a problem

Typer currently allows for `datetime.datetime`-objects as arguments,
but does not support the less specific `datetime.date`-object:
```python
# example.py
import datetine
import typer

def do_thing_with_date(date: datetime.date):
    typer.echo(f"Year is: {date.year}")

if __name__ == "__main__":
    typer.run(do_thing_with_date)

```
This code produces:
```shell
$ python example.py 2020-04-01
Traceback (most recent call last):
  File "dateapp.py", line 12, in <module>
    typer.run(do_thing_with_date)
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 743, in run
    app()
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 215, in __call__
    return get_command(self)()
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 241, in get_command
    click_command = get_command_from_info(typer_instance.registered_commands[0])
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 423, in get_command_from_info
    ) = get_params_convertors_ctx_param_name_from_function(command_info.callback)
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 404, in get_params_convertors_ctx_param_name_from_function
    click_param, convertor = get_click_param(param)
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 655, in get_click_param
    parameter_type = get_click_type(
  File "/usr/local/lib/python3.8/site-packages/typer/main.py", line 586, in get_click_type
    raise RuntimeError(f"Type not yet supported: {annotation}")  # pragma no cover
RuntimeError: Type not yet supported: <class 'datetime.date'>
```

### The solution you would like

Allow for `datetime.date` as input arguments/options to Typer.

### Describe alternatives you've considered

It is true that it is possible to get just the date from `datetime.datetime` in the current implementation:
- input parsing supports the format `%Y-%m-%d` without needing to specify a time
- you can get a `datetime.date`-object with `.date()`

However, I think that allowing for just the date part is better in some situations:
- No need to explicitly extract the date part with `.date()`
- It provides some intention to the end user that only the date part is considered
[...]

PR #116, opened 2020-06-11
➕ Added support for datetime.date CLI argument

IamCathal:
Adding support for arguments with type `datetime.date` as requested from [Issue #94](https://github.com/tiangolo/typer/issues/94)

By changing the import from `from datetime import datetime` to `import datetime` ( [see this stack overflow thread on it](https://stackoverflow.com/questions/16151402/python-how-can-i-check-whether-an-object-is-of-type-datetime-date) ) and by checking if the annotation type was either `datetime.datetime` or `datetime.date` we can support the new type. 

Not sure if this was a design decision to not include datetime.date but if it wasn't then I hope I've helped out even if it was only 2 lines. If I did anything drastically wrong please let me know.

All tests (`bash scripts/test-cov-html.sh`) pass with 100% coverage so nothing seems broken

This script which was given in the issue now works: 
```python
# example.py
import datetime
import typer

def do_thing_with_date(date: datetime.date):
    typer.echo(f"Year is: {date.year}")

if __name__ == "__main__":
    typer.run(do_thing_with_date)
```

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -1,5 +1,5 @@
+import datetime
 import inspect
-from datetime import datetime
 from enum import Enum
 from functools import update_wrapper
 from pathlib import Path
@@ -528,7 +528,7 @@ def get_click_type(
         return click.BOOL
     elif annotation == UUID:
         return click.UUID
-    elif annotation == datetime:
+    elif annotation == datetime.datetime or annotation == datetime.date:
         return click.DateTime(formats=parameter_info.formats)
     elif (
         annotation == Path


tiangolo:
Thanks for your contribution @IamCathal ! :rocket: 

So, here's the problem, Click will actually pass a value of type `datetime`, not `date` to the function.

Here's a simple test that could be added:

```Python
from datetime import date

import typer
from typer.testing import CliRunner

runner = CliRunner()

app = typer.Typer()


@app.command()
def get_date(d: date):
    assert type(d) == date
    typer.echo(f"Type: {type(d)} - Value: {d}")
    typer.echo("All good!")


def test_date():
    result = runner.invoke(app, ["2020-06-22"])
    assert result.exit_code == 0
    assert "Type: <class 'datetime.date'> - Value: 2020-06-22" in result.output
    assert "All good!" in result.output
```

---

Now, thinking about it, I'm checking, and the docs are currently not accurate about that: https://typer.tiangolo.com/tutorial/parameter-types/datetime/

The docs say that by default it will take an ISO string, but in reality, they will take any of `"%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M:%S"`. So, declaring `datetime` still allows passing a simple date.

There's actually no need to support the independent `date` type. :thinking: 

@IamCathal would you like to take a stab at updating the docs to clarify that by default it will take a datetime **or date** in ISO format?

IamCathal:
Sure I'll go ahead with updating the docs. I'll do it on a new branch so the date support commit isn't there

tiangolo:
Thanks! I just merged the other PR, so I'm gonna close this one.

Issue #125, opened 2020-06-25
Use Optional in docs


tiangolo:
As part of https://github.com/tiangolo/typer/issues/124, I updated the initial docs that introduce `Optional` for values that default to `None`.

We should have all the docs that have *CLI arguments* or *CLI options* with a default value of `None` to be declared using `Optional[some_type]`, that way the editor will be able to help to detect unsafe operations that assume that the value is not `None`.

The quick intro to `Optional` is now here: https://typer.tiangolo.com/tutorial/#python-types

rkbeatss:
will be working on it 🥰

tiangolo:
Thanks @rkbeatss ! :rocket: :cake:

PR #131, opened 2020-07-01
Support for Optional added to docs for CLI Argument and Options

rkbeatss:
Addressing #125 : all CLI arguments / options with default value of None declared using Optional generics


rkbeatss:
not sure why the build is stuck at the queued state

tiangolo:
Awesome! Thank you @rkbeatss ! :rocket: :cake: :tada: 

And thanks @zhezhel for the review :coffee:

PR #135, opened 2020-07-07
Update docs with the new --help argument

### LINKED ISSUE #124: Update docs with new Arguments help section
The new support for `help` parameters in *CLI arguments* from https://github.com/tiangolo/typer/pull/123 changes how the help text is generated with *CLI arguments*.

It also now shows default values by default.

So, many of the demos showing how it would look like in the terminal have to be updated to show the new real output.

* Any example that has *CLI arguments* and uses `--help` will change.
* And any example that has a default value for *CLI arguments* or *CLI options* and uses `--help`, will change too.

The task here is to read all the docs, page by page, and copy the code in the example, run the same commands as in the demonstration and update the output (copy it from the terminal) to the docs when it changes.

In most/all of the cases, it should require just to copy the example code, run it with the example commands, and copy the output back in the docs.

But it requires a lot of attention to the details to make sure not to skip/forget any example that also changes, to preserve the demonstration comments and guides, and to recognize when there are changes.

And the changes are probably a lot, spread through all the docs :sweat_smile:

victorphoenix3:
Fixes #124

### DIFF
--- a/docs/tutorial/multiple-values/arguments-with-multiple-values.md
+++ b/docs/tutorial/multiple-values/arguments-with-multiple-values.md
@@ -46,7 +46,7 @@ $ python main.py --help
 Usage: main.py [OPTIONS] [NAMES]...
 
 Arguments:
-  [NAMES]...  [default: Harry, Hermione, Ron]
+  [NAMES]...  Select 3 characters to play with  [default: Harry, Hermione,Ron]
 
 Options:
   --install-completion  Install completion for the current shell.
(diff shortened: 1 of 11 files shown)


victorphoenix3:
@tiangolo @kludex This is ready for review.
The PR was initially failing build, had to run `format.sh` hence the large size of the PR.

docs/tutorial/multiple-values/arguments-with-multiple-values.md
```diff
@@ -46,7 +46,7 @@ $ python main.py --help
 Usage: main.py [OPTIONS] [NAMES]...
 
 Arguments:
-  [NAMES]...  [default: Harry, Hermione, Ron]
+  [NAMES]...  Select 3 characters to play with  [default: Harry, Hermione,Ron]
```
Kludex, line 49:
```suggestion
  [NAMES]...  Select 3 characters to play with  [default: Harry, Hermione, Ron]
```

tiangolo:
Thanks @victorphoenix3 ! :rocket: :cake: 

And thanks @Kludex for the review! :coffee:

PR #143, opened 2020-07-16
🐛 Fix type conversion for `List` and `Tuple` and their internal types

### LINKED ISSUE #127: [BUG] Multiple value option gives tuple and not list
### Describe the bug

The standard way to configure an option that expects multiple values is to declare with `List`, as [described in the docs](https://typer.tiangolo.com/tutorial/multiple-values/multiple-options/).

However, this provides a tuple of values, and not a list of values, as one would expect based on the type hint declaration. It also contradicts the [documentation](https://typer.tiangolo.com/tutorial/multiple-values/multiple-options/), which states "You will receive the values as you declared them, as a list of str".

### To Reproduce

Steps to reproduce the behavior with a minimum self-contained file.

```Python
from typing import List
import typer


def my_command(opt: List[str] = ["foo", "bar"]):
    typer.echo(opt)
    typer.echo(type(opt))


if __name__ == "__main__":
    typer.run(my_command)
```

* Call it with:

```bash
python main.py
```

  * It outputs:

```
('foo', 'bar')
<class 'tuple'>
```

* Call it with:

```bash
python main.py --opt baz
```

  * It outputs:

```
('baz',)
<class 'tuple'>
```

### Expected behavior

I would expect to get a container of values that is class `list` as declared. 

### Environment

* OS: macOS 10.15.4
* Python 3.6.10
* Typer 0.3.0

hellowhistler:
Resolves #127 and an unticketed but related issue I found when digging through the code.

Fixes:
- `List` parameters are properly converted to a list. Previously this worked for lists of `Path` and `Enum` but all other types of lists were converted to tuples.
- `Tuple` parameters with `Path` or `Enum` elements will have those elements properly converted to their respective types. Previously this worked for all other types, but these two types were retained as strings. I removed a `TODO` item about recursive conversion which sounds like what I've done, but if that was referring to something else, I can re-add it.


hellowhistler:
Hi @tiangolo - would love to get your thoughts on this. I just pulled in the latest commits, made the code compliant with `mypy --strict`, and reworked the tests a little. Instead of putting them into the tutorial (which felt a bit like forcing a square peg), I moved them along with some other type conversion tests into a new file.

razy69 (approved):
Hi,

version: Typer==0.4.0

I experienced this issue in lazydocs project.
https://github.com/ml-tooling/lazydocs/blob/main/src/lazydocs/_cli.py#L33

Before this patch i had this error while trying to generate documentation.

```python
'tuple' object has no attribute 'append'
  File "/Users/razy69/.pyenv/versions/3.9.6/envs/test39/lib/python3.9/site-packages/lazydocs/_cli.py", line 49, in generate
    generate_docs(
  File "/Users/razy69/.pyenv/versions/3.9.6/envs/test39/lib/python3.9/site-packages/pysnooper/tracer.py", line 263, in simple_wrapper
    return function(*args, **kwargs)
  File "/Users/razy69/.pyenv/versions/3.9.6/envs/test39/lib/python3.9/site-packages/lazydocs/generation.py", line 966, in generate_docs
    ignored_modules.append(module_name)
```

I used pysnooper to dig the issue, and it is clearly related to Typer arguments conversion.

```bash
$ lazydocs ./app --ignored-modules "main" --ignored-modules "tests" --overview-file README.md
Source path:... /Users/razy69/.pyenv/versions/3.9.6/envs/test39/lib/python3.9/site-packages/lazydocs/generation.py
Starting var:.. ignored_modules = ('main', 'tests')
```

I applied this patch, and it successfully fix issue.

```bash
$ lazydocs ./app --ignored-modules "main" --ignored-modules "tests" --overview-file README.md
Source path:... /Users/razy69/.pyenv/versions/3.9.6/envs/test39/lib/python3.9/site-packages/lazydocs/generation.py
Starting var:.. ignored_modules = ['main', 'tests']
```

Michionlion:
@tiangolo Any update on possibly merging this fix? The workaround is minor (just add `list(param)` to get an actual list) but still an annoyance and potentially confusing. Seems like this PR is in pretty good working order. If there's any code changes that need to happen I'd be open to contributing myself, it'd be nice to have this merged.

aperiodic:
We were just bit by this bug ourselves. It would be nice to see this merged!

tiangolo:
Awesome, thanks @hellowhistler! :rocket: :cake: 

And thanks everyone for the discussion here! :coffee: 

This will be available in the next release, in some hours, Typer version `0.4.2` :tada:

PR #148, opened 2020-07-30
Add note on develop install. Add means of simply running subcommand program using typer.run

cottrell:
The linter is failing and I don't underestand it. Otherwise probably fine.

I was looking for something to replace `argh` library and thought this one was closest. It is very important to have the option to not touch the functions (no decoration on import) so that these libraries have the option of being completely unused in the case of no `if __name__ == '__main__':`.

You can even bury the import of typer in the ```if __name__``` block if you really wanted to make the CLI optional for some library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,27 @@ And that will:
 * Add two subcommands with `@app.command()`.
 * Execute the `app()` itself, as if it was a function (instead of `typer.run`).
 
+Alternatively, you can simply do
+
+```Python hl_lines="3  6  11  20"
+import typer
+
+
+def hello(name: str):
+    typer.echo(f"Hello {name}")
+
+
+def goodbye(name: str, formal: bool = False):
+    if formal:
+        typer.echo(f"Goodbye Ms. {name}. Have a good day.")
+    else:
+        typer.echo(f"Bye {name}!")
+
+
+if __name__ == "__main__":
+    typer.run(hello, goodbye)
+```
+
 ### Run the upgraded example
 
 <div class="termy">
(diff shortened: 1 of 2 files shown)


tiangolo:
Hey, thanks for your interest. So, the development instructions are all documented here: https://typer.tiangolo.com/contributing/

About the additional way to run applications with `typer.run()`, I prefer not to have it, I prefer `typer.run()` to work for the minimum scripts and examples. But for more complex applications that require having multiple sub-commands, I prefer to encourage a full application object and direct control of everything.

So I'll pass on this, but thanks for your effort. :coffee:

cottrell:
I think the main point is to not pollute the main code. 

I would encourage everyone to separate CLI from function code, doesn't need to go in the library as you say but basically a simple wrapper pattern to be used inside __main__. 

I've seen too many people write import and __main__ bombs with these kind of cli and they effectively end up blocking the ipython import workflow for debugging.

cottrell:
Probably good call to keep the library patterns simple. 

I would only emphasize that folks should aim to separate the CLI usage from the rest of their entry points. Firstly this allows easy swaps of CLI libraries. Secondly, not sure if this is applicable in this library, but we should be mindful not to block the "IPython + importlib.reload + %debug" workflow. I generally try to put all CLI stuff inside __main__ (assuming I am not debugging the CLI lib itself). 

So a re-usable pattern like this in my lib somewhere is what I revert to.

```
from typer import Typer
from typing import List, Callable, Any

def make_app(*functions: List[Callable]) -> Any:
    app = Typer()
    for fun in functions:
        app.command()(fun)
    return app
```

PR #152, opened 2020-08-07
use value.name as input value for enums

sapristi:
This is a pull request related to https://github.com/tiangolo/typer/issues/151

I did not change the value provided to the function (still an instance of the Enum class, not `value.value` as stated in the issue).
If you are ok with these changes I will try to update the documentation as well.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,8 +1,0 @@
-[run]
-
-source =
-    typer
-    tests
-    docs_src
-
-parallel = True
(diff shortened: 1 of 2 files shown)


sirex:
I think, this can be closed, I have a working version with green tests: https://github.com/tiangolo/typer/pull/224

tiangolo:
Thanks for the help here @sirex ! 👏  🙇 

Thanks for reporting back and closing the issue @sapristi  👍

PR #168, opened 2020-09-26
Improve installation of bash completion

lxmnk:
Editing `.bashrc` is no longer necessary now.

### DIFF
--- a/tests/test_completion/test_completion.py
+++ b/tests/test_completion/test_completion.py
@@ -1,7 +1,6 @@
 import os
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


amfarrell:
This PR also fixes the failures of `test_show_completion` and `test_install_completion` that occur while running fish. 🎉  🚀

tiangolo:
Thanks for the help and effort! :cookie: 

Do you have a reference about why that would no longer be needed? Where are you testing it? I'm imagining that this is based on some new release of Bash or some Linux distro... but what would happen for older systems?

PR #176, opened 2020-10-15
✏ Update `first-steps.md`, clarify distinction between parameter and argument

mccarthysean:
Made it clearer which we're talking about, parameter or argument.


tiangolo:
Good point, thanks @mccarthysean! :cake:

PR #206, opened 2020-12-16
✨ add custom logger handler

Faylixe:
The proposed logger handler use `echo` function as log record channel allowing to use traditional logging module with Typer.

More information in https://github.com/tiangolo/typer/issues/203

### DIFF
--- a/tests/test_others.py
+++ b/tests/test_others.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for the patience! :pray: 

I have a question, if you are using standard logging, which I would assume means the main output would be for logs, do you see any benefit in using Typer's (Click's) `echo` instead of the plain default logging module?

I think it would be preferable to use standard logging without going through this utility function.

Now I'm wondering if the docs sound like discouraging using anything else than `echo()` and if that makes it look like you should *not* use logging, because then that would be incorrect... :thinking:

Issue #195, opened 2020-11-21
[FEATURE] Add Extra Exception Handling


malthunayan:
### Is your feature request related to a problem

I want to be able to raise exceptions other than `BadParameter` because not all exceptions (at least with my app) fall under this type. The parameters and arguments could be technically correct, but the usage is wrong, or the file passed in could not be opened. Also, built-in python exceptions will give traceback errors with `typer`, which I do not want to expose in the final product.

### The solution you would like

I believe more imports from `click.exceptions` API in the `__init__.py` file of the `typer` package will give users access to different types of exceptions that `typer` can handle will solve the issue. For example:
```
"""Typer, build great CLIs. Easy to code. Based on Python type hints."""
...
from click.exceptions import Abort as Abort
+ from click.exceptions import BadArgumentUsage as BadArgumentUsage
+ from click.exceptions import BadOptionUsage as BadOptionUsage
from click.exceptions import BadParameter as BadParameter
+ from click.exceptions import ClickException as TyperException
from click.exceptions import Exit as Exit
+ from click.exceptions import FileError as FileError
+ from click.exceptions import NoSuchOption as NoSuchOption
+ from click.exceptions import UsageError as UsageError
from click.termui import clear as clear
...
```

This way users can import any of the available exceptions from `click` and raise them when applicable. Should they need to create a custom exception they can subclass from `TyperException` to create their own exception that can be handled by `typer`, because it is handled by `click` under the hood.

### Describe alternatives you've considered

Users can equally just import from `click` and use those exceptions, but it would be nice to have them under one import. Also, for those using high-level package managers (e.g., `Poetry`), and are specific about their dependencies, they will add `click` to their dependencies, to be explicit about its usage in the project.

### Additional context

I've import `UsageError` and `FileError` from `click` in my project, and tested their usage, it seems like `typer` handles them out of the box, because `click` can handle them.

malthunayan:
I would be happy to open a PR for this feature request, and equally happy to add to the documentation to clarify the exceptions available. This is similar to how `fastapi` imports some of `starlette`'s exceptions to be publicly available to users under `fastapi`, but at the same time users could import from `starlette` directly.

tiangolo:
Hey! Thanks for the interest. Checking the docs, it seems like most of these exceptions are handled by Click internally, they handle the data validation, and if they are raised, they are raised before your code is executed.

[...]

PR #209, opened 2020-12-17
Add exceptions from click

malthunayan:
Hello, @tiangolo,
Just a thought here that adding these exceptions from `click` helps give users more flexibility in terms of what exceptions they can raise.

[[FEATURE] Add Extra Exception Handling #195](https://github.com/tiangolo/typer/issues/195)

### DIFF
--- a/typer/__init__.py
+++ b/typer/__init__.py
@@ -3,8 +3,14 @@
 __version__ = "0.3.2"
 
 from click.exceptions import Abort as Abort
+from click.exceptions import BadArgumentUsage as BadArgumentUsage
+from click.exceptions import BadOptionUsage as BadOptionUsage
 from click.exceptions import BadParameter as BadParameter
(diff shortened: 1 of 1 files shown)


tiangolo:
Hey! Thanks for the interest and effort!

I replied to your issue here: https://github.com/tiangolo/typer/issues/195#issuecomment-1186444136 but I'll copy the message here as well for completeness.

---

Checking the docs, it seems like most of these exceptions are handled by Click internally, they handle the data validation, and if they are raised, they are raised before your code is executed.

The only one I see that seems to be useful in the app's code would be the `BadParameter` that can be raised from a callback, that's why that one is exported there and documented https://typer.tiangolo.com/tutorial/options/callback-and-context/#validate-cli-parameters

Is there a specific use case that is not covered by the default handling, where you would need to raise an exception, and that needs to be one of the Click internal exceptions?

PR #210, opened 2020-12-18
🐛 Fix `context_settings` for a Typer app with a single command

daddycocoaman:
Addresses https://github.com/tiangolo/typer/issues/208

When the design choice is to create an `typer.Typer()` object with `context_settings`, that context should be passed even when only a single command is registered. This is already handled when there are more than two commands since a Group is created in that scenario but should also exist for single commands.


malthunayan:
@daddycocoaman great work on your PR, maybe your predicate in line 241:
```
if not single_command.context_settings and typer_instance.info.context_settings:
    ...
```

could be:
```
if isinstance(single_command.context_settings, DefaultPlaceholder) and not isinstance(typer_instance.info.context_settings, DefaultPlaceholder):
    ...
```

That way if a user (for some odd reason) sets the context settings explicitly to `None` in their decorator it would no longer use the context settings instantiated in the `typer` instance. Also, I think it aligns more with how the codebase determines if a change is required. I'm not entirely sure, what do you think?

daddycocoaman:
That makes perfect sense. When I first tried using DefaultPlaceholder, I must have messed up the conditional and the local test failed. I'll make that change.

daddycocoaman:
@malthunayan It turns out that the `CommandInfo` and `Typer.command` doesn't use the DefaultPlaceholder class to represent default values, which is why it didn't work when I originally tried it. I tried adding it to the classes, but it broke some other things further along, like documentation generation. 

```
doc = <typer.models.DefaultPlaceholder object at 0x00000297B400C7F0>

    def cleandoc(doc):
        """Clean up indentation from docstrings.

        Any whitespace that can be uniformly removed from the second line
        onwards is removed."""
        try:
>           lines = doc.expandtabs().split('\n')
E           AttributeError: 'DefaultPlaceholder' object has no attribute 'expandtabs'

doc        = <typer.models.DefaultPlaceholder object at 0x00000297B400C7F0>

..\..\..\AppData\Local\Programs\Python\Python38\lib\inspect.py:631: AttributeError
```
Since the value is already set to `None`, looks like it shouldn't be an issue. I'd rather leave that much of a change up to @tiangolo (although I would be willing to do it) so for now I'll just use  
`if not single_command.context_settings and not isinstance(typer_instance.info.context_settings, DefaultPlaceholder)` to align with existing code and keep the PR simple. Although, I do think those parameters should eventually become `DefaultPlaceholder`.

malthunayan:
Hey, @daddycocoaman,
Thank you very much on your detailed explanation. I didn't actually try out the predicate I've suggested above, and I do agree with everything you have said.

davised:
I was just running into this issue and I'm glad this PR is here. @tiangolo do you plan on pulling this in?

tiangolo:
Awesome, thanks a lot for your contribution @daddycocoaman! :cake: 

And thanks for the discussion here everyone! :clap: :bow:

Issue #215, opened 2021-01-05
[QUESTION] Default datetime must be string (bug in older versions of Click)


aguinane:
### Describe the bug

When specifying a default value for an option of type datetime, it will throw a TypeError. 
You need to instead provide a string as click expects a string from the CLI and tries to convert it. 

### To Reproduce

Modifying the [datetime example](https://typer.tiangolo.com/tutorial/parameter-types/datetime/) slightly:

```Python
from datetime import datetime

import typer


def main(birth: datetime = datetime(2021, 1, 1)):
    typer.echo(f"Interesting day to be born: {birth}")
    typer.echo(f"Birth hour: {birth.hour}")


if __name__ == "__main__":
    typer.run(main)
```

* Call it with:

```bash
python main.py 1956-01-31T10:00:00
```

* It outputs:

```
 File "\venv\lib\site-packages\click\types.py", line 246, in convert
    dtime = self._try_to_convert_date(value, format)
  File "\venv\lib\site-packages\click\types.py", line 239, in _try_to_convert_date
    return datetime.strptime(value, format)
TypeError: strptime() argument 1 must be str, not datetime.datetime
```

### Environment

* OS: Windows
* Typer Version: 0.3.2
* Python version: 3.8

### Additional context

It works if you specify a string as the default value instead. But this causes mypy to be unhappy as you are assigning a value of the wrong type. It would be worth either calling this out in the documentation, or adding an additional check before sending the value to click for conversion to see if it's already a datetime.

alexreg:
This has been fixed in more recent versions of Click.

tiangolo:
Thanks for reporting it! :nerd_face: 

And thanks @alexreg for the help. :bow: 

It seems this is now fixed in Click. I just tried your example and it seems to work correctly now.  So it shouldn't need anything else here, right?

aguinane:
yep working now.

PR #216, opened 2021-01-05
Convert default values for datetime option to str

aguinane:
If you specify a default value for a DateTime Option of type datetime, click will throw an error.
This converts it to a string first, so that click can then convert it back to a datetime again. 

Fixes Issue #215 [BUG] Default datetime must be string

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -614,6 +614,9 @@ def get_click_param(
     else:
         default_value = param.default
         parameter_info = OptionInfo()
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest and effort! :cake: 

Checking the issue and discussion, it seems this was fixed in Click. So I think we can probably close this one, right?

aguinane:
No longer needed - fixed in click.

PR #224, opened 2021-01-21
✨ Add option to use Enum names instead of values on the commandline

### LINKED ISSUE #151: [FEATURE] Use enum.name as parameter, and enum.value as value received by the function

I would like to be able to unleash the power of Enum ! The first case I stumbled onto is that of logging, e.g have the following Enum usable in typer:

```python
class LogLevel(Enum):
    debug = 10
    info = 20
    warning = 30
```

Instead I have to use  this one, and then convert the string to the right value

```python
class LogLevel(Enum):
    debug = "debug"
    info = "info"
    warning = "warning"
```

I would like typer to interpret the `value.name` as arguments, and provide `value.value` to the called function, so that the following program would work as intended:

```python
from enum import Enum
import typer
import logging

logger = logging.getLogger(__name__)

class LogLevel(Enum):
    debug = logging.DEBUG
    info = logging.INFO
    warning = logging.WARNING

def main(
        level: LogLevel,
):
    logger.setLevel(level)
    ...

if __name__ == "__main__":
    typer.run(main)
```

sirex:
Fixes: https://github.com/tiangolo/typer/issues/151

Added `names` parameter, in order to user Enum names instead of values. Also for IntEnum, names are used by default, even if names is False.


Aerilius:
Thanks! This feature is so much missing when using enums with values of other types than strings.

However I got an error with this code:
```python
    transform: SymmetryTransformation = typer.Option(SymmetryTransformation.TRANSPOSE, names=True),
```
```
Error: Invalid value for '--transform': invalid choice: SymmetryTransformation.TRANSPOSE. (choose from IDENTITY, ROT90, ROT180, ROT270, FLIP_X, FLIP_Y, TRANSPOSE, ANTITRANSPOSE)
```
I had to change the default to the value that should be provided on the CLI, not as received in the application:
```python
    transform: SymmetryTransformation = typer.Option("TRANSPOSE", names=True),
```

mattchan-tencent:
Is there some update on whether this can be merged? This would make the Choices feature much more pythonic.

abondrnco:
Bumping this for interest, surprised this isn't the default behavior but understandable if all of the exploration done around this feature was with string enums

docs_src/parameter_types/enum/tutorial003.py
```diff
@@ -0,0 +1,18 @@
...
+import typer
+
+
+class LogLevel(enum.Enum):
+    debug = logging.DEBUG
+    info = logging.INFO
+    warning = logging.WARNING
+
+
+def main(log_level: LogLevel = typer.Option(LogLevel.warning, names=True)):
```
ryangalamb, line 13: `LogLevel.warning` didn't work as the default value when I tried it.

```console
$ python docs_src/parameter_types/enum/tutorial003.py
Usage: tutorial003.py [OPTIONS]
Try 'tutorial003.py --help' for help.

Error: Invalid value for '--log-level': invalid choice: LogLevel.warning. (choose from debug, info, warning)
```

Can you please fix this and add a test case that runs `runner.invoke(app)` to exercise the default value?

docs/tutorial/parameter-types/enum.md
```diff
@@ -66,3 +66,32 @@ Training neural network of type: lstm
...
+
+```console
+$ python main.py --log-level debug
+
+Log level set to DEBUG
+```
+
+</div>
+
+If `IntEnum` type is given, then enum names are used implicitly.
```
ryangalamb, line 93: I think this is reasonable, but I'm not sure about doing it implicitly. I'd rather go for consistency here.

Imagine a user who sees an `IntEnum` before seeing a str enum. They'll (reasonably) assume that enums use names instead of values by default. And then they'll be in for a surprise when they encounter the default behavior of all other enums.

Keeping it consistent seems best to me. Does anyone else have any thoughts on this?

docs_src/parameter_types/enum/tutorial004.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Access(enum.IntEnum):
+    private = 1
+    protected = 2
+    public = 3
+    open = 4
+
+
+def main(access: Access = typer.Option(Access.private)):
```
ryangalamb, line 13: The default is broken on this one too.

```console
$ python docs_src/parameter_types/enum/tutorial004.py
Usage: tutorial004.py [OPTIONS]
Try 'tutorial004.py --help' for help.

Error: Invalid value for '--access': invalid choice: 1. (choose from private, protected, public, open)
```

ryangalamb (dismissed):
This is a super useful feature, but I noticed some issues while reviewing this:

* The default values are broken. Should add a test case to cover that too. (Also pointed out by @Aerilius)
* The name `names` should be changed to indicate that it's specific to enums.
* The implicit behavior with `IntEnum` will likely cause confusion. I'd prefer consistency.

I'd also like to see a test case covering `Argument`. (I see that this is implemented for `Argument`, but a test case would help guard against accidentally breaking it.)

I know this has been open for a while, so no worries if you don't have the time/interest to keep working on this. Either way, thank you for getting this started!

docs/tutorial/parameter-types/enum.md
svlandeg: I agree - consistency is more important here.

docs_src/parameter_types/enum/tutorial003.py
svlandeg: Good catch! I pushed a fix for this.

docs_src/parameter_types/enum/tutorial004.py
svlandeg: Should be fixed now 🙏

typer/main.py
```diff
@@ -873,6 +886,11 @@ def get_click_param(
             annotation=main_type, parameter_info=parameter_info
         )
     convertor = determine_type_convertor(main_type)
+    if lenient_issubclass(main_type, Enum):
+        if parameter_info.enum_by_name:
+            convertor = generate_enum_name_convertor(main_type)
+        else:
+            convertor = generate_enum_convertor(main_type)
```
svlandeg, line 893: This bit is currently still hacky and not 100% functional. The `determine_type_convertor` is not yet adjusted and always returns `generate_enum_convertor` instead of potentially switching to `generate_enum_name_convertor`. In this code block, that is being overwritten by the additional `if lenient_issubclass(main_type, Enum)` check here, but there's an issue when you'd use an `Enum` inside of a `Tuple`, because `generate_tuple_convertor` will call `determine_type_convertor` directly and thus will not be able to convert the enum correctly.

This can probably be fixed by passing around more parameters, but it might look a bit ugly 🫤  

Let me contribute some of the examples with Tuples+enums I've been testing on, to showcase where the error is (the CI will go red again 😢 )

svlandeg: Update: pushed a (somewhat ugly) fix that address this 👇

docs_src/parameter_types/enum/tutorial006_an.py
```diff
@@ -0,0 +1,21 @@
...
+    f1 = "Eggs"
+    f2 = "Bacon"
+    f3 = "Cheese"
+
+
+def main(
+    groceries: Annotated[List[Food], typer.Option(enum_by_name=True)] = ["f1", "f3"],
+):
+    print(f"Buying groceries: {', '.join([f.value for f in groceries])}")
+
```
svlandeg, line 18: The linter is unhappy with this code, but does accept similar code in `tutorial003_an.py`:
```
def main(groceries: Annotated[List[Food], typer.Option()] = [Food.food_1, Food.food_3]):
```
🤔

svlandeg: Fixed!

docs_src/multiple_values/options_with_multiple_values/tutorial002.py
```diff
@@ -0,0 +1,25 @@
...
+import typer
+
+
+class Food(str, Enum):
+    f1 = "Eggs"
+    f2 = "Bacon"
+    f3 = "Cheese"
+
+
+def main(user: Tuple[str, int, bool, Food] = typer.Option((None, None, None, Food.f1))):
```
svlandeg, line 13: This file and the next (`tutorial002` and `tutorial003`) are important to test, because they encovered a bug. But it feels overkill to actually mention these in the documentation? So maybe we should move these tests to somewhere else?

svlandeg (commented):
Ok, to recap.

This PR is pretty old, but nevertheless the feature proposed here makes sense to me: being able to refer to an `Enum` with its names instead of values. 
This is particularly useful in cases where the values aren't strings, and it might be more intuitive for users to specify the enum names instead.

Thanks for your work on this @sirex, and thanks to @Aerilius and @ryangalamb for the reviews - all issues you both highlighted should now be addressed.

I spent some time on this PR adding additional tests and fixing a few issues + adding more documentation.

I'll leave this up to Tiangolo now for a final review. For his benefit, some points to highlight:
- The actual implementation of this requires the addition of the parameter `enum_by_name` everywhere + a custom `generate_enum_name_convertor`. 
- To work with tuples, in `main.py` we do need to cary around the `enum_by_name` parameter to the convertor functions as well. This is slightly ugly, but I don't see a better way right now.
- The main functionality is tested with the new tests in `parameter_types/enum` and described in `enum.md`
- The additional tests in `arguments_with_multiple_values` and `options_with_multiple_values` are important because they cover using the enum in a list/tuple, but these tests aren't yet covered in the documentation. It felt to me like that would become too complex, but happy to either mention them anyway, or move them out of the test_tutorial folder.

svlandeg:
Update:

- I merged in `master` and resolved the conflict
- I updated the docs to the new format of #1054
- I fixed a test where `stdout` was being checked instead of `stderr` (or just simply `output`)

docs/tutorial/parameter-types/enum.md
```diff
@@ -74,6 +74,37 @@ Training neural network of type: lstm
 
 </div>
 
+### Using Enum names instead of values
+
+Sometimes you want to accept `Enum` names from the command line and convert
+that into `Enum` values in the command handler. You can enable this by setting
+`enum_by_name=True`:
+
+{* docs_src/parameter_types/enum/tutorial007_an.py hl[14] *}
```
svlandeg: Note: throughout this file, the numbering of the tutorials is now pretty disorganized, because of the addition of new ones in between existing ones. I suggest to keep this as-is in this PR to not complicate the git history, but if we're annoyed by the out-of-sequence usage of tutorials here (001-002-007-005-003-006-004), we could fix it afterwards.

docs_src/multiple_values/options_with_multiple_values/tutorial002_py39.py
```diff
@@ -0,0 +1,25 @@
+from enum import Enum
```
svlandeg: A quick note that this tutorial file isn't actually referenced from anywhere within the tutorial markdown files.

docs_src/multiple_values/arguments_with_multiple_values/tutorial003_py39.py
```diff
@@ -0,0 +1,29 @@
+from enum import Enum
```
svlandeg: A quick note that this tutorial file isn't actually referenced from anywhere within the tutorial markdown files.

docs_src/multiple_values/options_with_multiple_values/tutorial003_py39.py
```diff
@@ -0,0 +1,29 @@
+from enum import Enum
```
svlandeg: A quick note that this tutorial file isn't actually referenced from anywhere within the tutorial markdown files.

svlandeg (commented):
I had another look at this PR, and brought it up-to-date with the latest docs formatting on `master`. I've also removed several unnecessary test files that weren't actually mentioned in the tutorial. 

I think the PR is now again ready for review 🙏

tiangolo:
Thanks for the work and interest on this! There are several refactors we needed to do, and some more we are still missing before tackling this.

I didn't want to close PRs before providing an alternative solution, but I ended up leaving too many PRs open for too long, sorry for that.

For now I'll close this one for cleanup, but we'll continue tracking this feature request in the discussion and internally. Thanks! :coffee:

PR #225, opened 2021-01-21
✏ Fix double "and" typo in first-steps tutorial

softwarebloat:


svlandeg (approved):
Thanks!

softwarebloat:
of course it's not an urgent PR but i was wondering why some status hasn't run :thinking:

svlandeg:
Good question which I'll have to defer to @tiangolo ...

tiangolo:
Thanks for the contribution @softwarebloat! :mag: 

The missing checks were probably because I added some new required CI checks that probably didn't exist when you created this PR, but after updating the branch now they are found and run. :rocket: 

Thanks @svlandeg as always! :cake: :bow:

PR #237, opened 2021-02-22
🐛 Fix `zsh` autocompletion installation

alexjurkiewicz:
I ran into issues installing completiong from a typer-enhanced program. Here are two fixes:

1. Don't set the completion style. This overrides user styling globally and is a purely personal, cosmetic preference.
2. Adjust the order of lines inserted into `.zshrc`. `compinit` must be run _after_ `fpath` is adjusted, for new completions to be loaded.

I tested this logic with my current zsh (using oh-my-zsh) and with a bare zshrc, and it works in both cases.


RuiLoureiro:
I was having problems with autocompletion on zsh (oh-my-zsh), and these changes fixed it.

svlandeg (approved):
I can confirm that the `zsh` completion wasn't working as expected. The second `"compinit"` line was never written to `.zshrc` due to the `if line not in completion_init_lines` check, as the string `"compinit"` was already part of the first line.

Additionally, as pointed out in the description of the PR and in [previous discussions](https://github.com/tiangolo/typer/issues/535#issuecomment-1604148651), the order of commands was wrong.

On `master`, my `.zshrc` would look like this:

```
autoload -Uz compinit
zstyle ':completion:*' menu select
fpath+=C:\Users\Sofie\.zfunc
```
and autocompletion wouldn't work.

With this PR, my `.zshrc` would look like this:

```
autoload -Uz compinit
fpath+=~/.zfunc
compinit
```
and autocompletion works 🎉 

Thanks for this contribution @alexjurkiewicz and apologies for the long delay in reviewing this!

tiangolo:
Great, thank you @alexjurkiewicz! :rocket: 

And thanks @svlandeg for the review! :bow: 

---

I've been studying today a bit about all this (I always have to re-study all the completion stuff when working on this :sweat_smile: ), these were good resources:

* https://thevaluable.dev/zsh-install-configure-mouseless/
* https://thevaluable.dev/zsh-completion-guide-examples/

I realized that all the Zsh completion installation could be a single line, which makes it easier to test for in the `if` block.

About the styling, as up to now the completion has installed the style configs, I won't remove this, at least not in this PR that is fixing a bug, maybe that could be considered a breaking change as it's a change in behavior, not sure. :thinking: 

Still, I updated it to checking if there's a `zstyle` config in the `~/.zshrc` and only set this default config if there's no other one.

My doubt about removing the `zstyle` config is that I don't expect many people to know how to tweak it, but they all would benefit from a default nice UX from the CLIs they install, that's the tradeoff I'm trying to balance. Maybe an option is to set the configs only for the current app, that would mean people would have the menu only for this program they are installing but not for the rest of their shell... still not sure. I added a working (commented) line there with that config so we can revisit it later.

For now, let's leave this PR at fixing completion installation for Zsh. :sunglasses: :rocket: 

Thank you for your contribution! :cake: 

This will be available in Typer `0.12.4` released in the next hours. :tada:

PR #255, opened 2021-03-29
Use expanduser() on Path arguments/options

### LINKED ISSUE #252: [BUG] Path option doesn't seem to expand user
### Describe the bug

When the path have `~` in the path it doesn't seem to resolve.

### To Reproduce

```Python
import typer

app = typer.Typer()

@app.command()
def login(
    config_path: Path = typer.Option(
        default=os.environ.get("PROFILE", "~/.aws/my_profile"),
        exists=True,
        file_okay=True,
        dir_okay=False,
        readable=True,
        resolve_path=True,
    )
):
  typer.echo(f"Hello {config_path}")

if __name__ == "__main__":
    app()
```

* Call it with:

```bash
python main.py
```

* It outputs:

```
Error: Invalid value for '--config-path': File '~/.aws/my_profile' does not exist.
```

```
Hello ~/.aws/my_profile 
```
or
```
Hello /User/username/.aws/my_profile
```

### Environment

* OS: macOS
* Typer Version: 0.3.2
* Python version: 3.6.12

pmav99:
Fixes #252

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,0 +1,11 @@
+import pathlib
+
+from typer import main
+
+
+def test_param_path_convertor() -> None:
+    assert main.param_path_convertor(None) is None
+    assert main.param_path_convertor("/foo") == pathlib.Path("/foo")
+    # check that expanduser has been called.
+    # If we start to run tests on windows, we will probably need to update this
+    assert main.param_path_convertor("~/foo").as_posix().startswith("/")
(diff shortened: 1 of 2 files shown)


curoky:
Thank you for your contribution, but please don't enable it by default, and there are also some discussions in https://github.com/pallets/click/issues/287.

YuriiMotov:
@pmav99, thank you for your interest and efforts!

I think we shouldn't make this change. I explained my opinion in the issue comment: https://github.com/fastapi/typer/issues/252#issuecomment-3301767899

PR #268, opened 2021-04-18
Improve testing stderr tip.

CharString:
It wasn't at all obvious to me that the error I got was not from my code or pytest, but  from Typer.
Preferrably I'd put something like 'pass mix_stderr=False to CliRunner' in the error message, and keep the tip short. But I couldn't find it quickly. I think it's upstream in the Click runner.


tiangolo:
Thank you! This was solved recently on another PR, so I'll close this one. But thanks for the effort! :coffee:

Issue #269, opened 2021-04-19
[FEATURE] treating keyword-only arguments as Required CLI Options


podhmo:
### Is your feature request related to a problem

There is a shorthand for the unrequired CLI option. 

```py
def command(verbose:bool=False):
    ...
```

But the shorthand for the required CLI option is not provided.

### The solution you would like

So, we can use using keyword-only arguments as a marker for the required CLI Option. This is my suggestion. 

following

```py
def command(*, verbose:bool):
    ...
```

Now, verbose is treated as the required flag.

### Describe alternatives you've considered

Currently, if we need the required CLI options, write the code as follows:

```py
def command(verbose:bool=typer.Option(...)):
    ...
```

https://typer.tiangolo.com/tutorial/options/required/

YuriiMotov:
Similar feature request was converted into discussion: #331

I think the drawback of such approach is that it might break the existing code bases.
It's common do declare all parameters as keyword-only in order to avoid having to place parameters with defaults after parameters without defaults. This feature might break such code..

```py
import typer

app = typer.Typer()

@app.command()
def hello(
    *,
    formal: bool = False,
    name: str,  #  <=  This will become option..
    color: str = "green",
):
    pass


if __name__ == "__main__":
    app()
```

tiangolo:
This is something I would consider but we need to handle a few internal refactors first. 

We'll track it internally. For now I'll close this one for cleanup (and to avoid ai slop bait 😅).

PR #270, opened 2021-04-20
keyword-only arguments as required Option

podhmo:
supports #269

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -610,7 +610,10 @@ def get_click_param(
             default_value = parameter_info.default
     elif param.default == Required or param.default == param.empty:
         required = True
-        parameter_info = ArgumentInfo()
+        if param.kind == inspect.Parameter.KEYWORD_ONLY:
+            parameter_info = OptionInfo()
+        else:
+            parameter_info = ArgumentInfo()
     else:
         default_value = param.default
         parameter_info = OptionInfo()
--- a/typer/models.py
+++ b/typer/models.py
@@ -425,7 +425,9 @@ class ParamMeta:
         name: str,
         default: Any = inspect.Parameter.empty,
         annotation: Any = inspect.Parameter.empty,
+        kind: inspect._ParameterKind = inspect.Parameter.POSITIONAL_OR_KEYWORD,
     ) -> None:
         self.name = name
         self.default = default
         self.annotation = annotation
+        self.kind = kind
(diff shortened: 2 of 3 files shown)


typer/models.py
```diff
@@ -425,7 +425,9 @@ def __init__(
         name: str,
         default: Any = inspect.Parameter.empty,
         annotation: Any = inspect.Parameter.empty,
+        kind: inspect._ParameterKind = inspect.Parameter.POSITIONAL_OR_KEYWORD,
```
podhmo, line 428: accessing `inspect._ParameterKind` is ugly a bit, but the type name is `_ParameterKind` 
https://github.com/python/typeshed/blob/a690a14c820238ca7a64cd502343d6b5a649e731/stdlib/inspect.pyi#L125-L134

typer/main.py
```diff
@@ -610,7 +610,10 @@ def get_click_param(
             default_value = parameter_info.default
     elif param.default == Required or param.default == param.empty:
         required = True
-        parameter_info = ArgumentInfo()
+        if param.kind == inspect.Parameter.KEYWORD_ONLY:
```
podhmo, line 613: Where should I write the test?

Afoucaul:
@podhmo do you need any help on this PR? I really feel like this feature would be game breaker.

Afoucaul:
@tiangolo What is missing in this PR so that it can be accepted? How could I help making it complete?

typer/main.py
ryangalamb: Ideally, there would be a tutorial file in `docs_src` demonstrating the behavior. And then a test for that file in `tests/test_tutorial/`. Then reference the file in the markdown documentation with a brief explanation.

Sometimes there are tricky edge cases that don't fit into tutorials. In that case, you can add some tests in the top level of the `tests/` directory.

ryangalamb (changes requested):
As pointed out earlier, tests are the main thing blocking this right now.

FWIW, this is a breaking change (existing CLIs that use keyword-only arguments will change behavior), but I think this is useful enough to justify a breaking change.

Let me know if you have any more questions about how/where to add tests.

podhmo:
Thank you for the comments, but sorry, I've lost the motivation. 

So, feel free to reopen or reuse the code in new pull request.

ryangalamb:
Sounds good, thanks for responding and sharing your code/time!

PR #274, opened 2021-05-02
✏ Fix minor typo in `index.md`

RmStorm:
minor typo fix:
changing `If you need a refreshed` to `If you need a refresher`.


svlandeg (approved):
Thanks!

tiangolo:
Good catch, thanks @RmStorm! :mag: 

And thank you @svlandeg! :cookie:

PR #277, opened 2021-05-07
Windows encoding bugfix

matejfroebe:
Problem: When installing auto-completion in Powershell, the path to
the profile file was decoded with windows-1252 encoding - but such
path didn't exist in my case. The correct encoding was cp850.

Solution:
- try to decode with cp850 also
- check if the decoded path exists (checking for UnicodeDecodeError isn't enough)


typer/_completion_shared.py
```diff
@@ -183,15 +183,19 @@ def install_powershell(*, prog_name: str, complete_var: str, shell: str) -> Path
...
+                if os.path.exists(path_str):
+                    path_obj = Path(path_str.strip())
+                    break
             except UnicodeDecodeError:
-                click.echo("Couldn't decode the path automatically", err=True)
-                raise
+                pass
+        if path_obj is None:
+            click.echo("Couldn't decode the path automatically", err=True)
+            raise click.exceptions.Exit(1)
```
svlandeg, line 197:
```suggestion
            raise 
```

svlandeg (commented):
Hi!

Thanks for your contribution, and I apologize for the delay in reviewing this.

For easier reviewing, I suggest to split the suggested changes up in two PRs:
- one adding the `cp850` encoding - this should be straightforward.
- one for also checking whether the path exists.

After some initial testing, it looks like the second PR would require some changes to the test suite, so I prefer getting the first change in as such, and then build on it. I'll look into this after https://github.com/tiangolo/typer/pull/808 is merged.

typer/_completion_shared.py
```diff
@@ -183,15 +183,19 @@ def install_powershell(*, prog_name: str, complete_var: str, shell: str) -> Path
...
             try:
-                path_str = result.stdout.decode("utf8")
+                path_str = result.stdout.decode(encoding).strip()
+                if os.path.exists(path_str):
+                    path_obj = Path(path_str.strip())
+                    break
             except UnicodeDecodeError:
-                click.echo("Couldn't decode the path automatically", err=True)
-                raise
+                pass
```
lachaib, line 194: Why not use contextlib.suppress there?

tiangolo:
Thanks @matejfroebe! :nerd_face: This was handled in https://github.com/fastapi/typer/pull/808, so I'll close this one now. The fix will be available in the next release in a few hours. 0.13.1 :rocket:

PR #281, opened 2021-05-15
Enable click 8 support

ssbarnea:


svenstaro:
Hey, any progress on this? I'm wondering whether I need to provide an old version of python-click in Arch Linux or whether this is an easy fix but I don't have time to look into the actual code problems right now.

anthraxx:
It should be relatively easy. typer used some internal data structures. In click 8 those have been reworked into API classes. Overall the changes don't look crazy but guess typer needs to have compat code to support both paths now. Someone just needs to jump the hoop

fabaff:
The NixOS packages is no longer able to build due to the missing support for `click>8`. Same issue as with the Arch package.

tiangolo:
Thanks for your interest! :cake: 

This was handled in https://github.com/tiangolo/typer/pull/317, making Typer compatible with **both** Click 7 and Click 8, and is now released in Typer version `0.4.0`. 🎉

Given that, I'll close this one now. :coffee:

PR #284, opened 2021-05-23
🎨 Fix type annotation of `typer.run()`

yassu:
Problem: type annotation of typer.run is

``` python
def run(function: Callable[..., Any]) -> Any
```

However, this function returns `None`.

Solution: type annotation of typer.run changes to

``` python
def run(function: Callable[..., Any]) -> None
```


alexreg:
Alternatively, the last line of the function could be `return app()` instead of `app()`, but this probably isn't necessary.

tiangolo:
Good catch, thanks @yassu! :rocket:

PR #291, opened 2021-06-12
Refactored autocompletion tutorial examples to use list comprehensions.

ajoino:
Refactored autocompletion examples 3 and 4 to use list comprehensions.


svlandeg (approved):
Looks good to me, thanks!

tiangolo:
Thanks for the suggestion! In this case, I think it's more understandable and easier to follow with the explicit lists, so I'll pass on this one. But thanks for the effort! :coffee:

PR #304, opened 2021-07-22
Support existing click.ParamType instance in type hint

JacobHayes:
We had a custom `KeyValueType(click.ParamType)` we used in `click.option(..., type=KeyValueType())` for `key=value` style flags. When converting to `typer`, I was hoping to be able to set these in the param type hint, but ran into `RuntimeError(f"Type not yet supported: {annotation}")`. Adding this extra check fixed my use, and I'm guessing any other custom `ParamType`s?

Is this the right place/way to add this support? If so, happy to dig into how to add a test for this!

cc @rockwellw


tiangolo:
Thanks! Support for this was added in https://github.com/tiangolo/typer/pull/583, the new docs are here: https://typer.tiangolo.com/tutorial/parameter-types/custom-types/, it's available in Typer `0.8.0`, just released. :tada: 

Given that, I'll close this one, but thanks for your effort! :cake: :rocket:

PR #308, opened 2021-08-03
✏ Fix typo, "ASCII codes" to "ANSI escape sequences"

septatrix:
ASCII is the encoding whereas [ANSI escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code) are what the terminals interpret for colors etc.


tiangolo:
Good catch, thank you! :rocket: :cake:

PR #326, opened 2021-08-31
✨ Support Date parameter type (#94)

kozlek:
Implements an extra click type `Date`, based on built-in `click.DateTime` to enable Typer support for Python `datetime.date`.

This approach could unlock others unsupported Python build-in types, even if the best solution would be to implement them directly inside Click.

Another approach - more complex but more flexible - would be to autorise Type's users to define their own extra Click types at Typer's app level using a mapping "annotation -> click type implementation":

```python
from datetime import date, datetime
from typing import Any, Optional, Sequence

import click
from typer import Typer


class Date(click.DateTime):
    name = "date"

    def __init__(self, formats: Optional[Sequence[str]] = None):
        self.formats = formats or ["%Y-%m-%d"]

    def _try_to_convert_date(self, value: Any, format: str) -> Optional[date]:
        try:
            return datetime.strptime(value, format).date()
        except ValueError:
            return None

    def __repr__(self) -> str:
        return "Date"


app = typer.Typer(custom_types={date: Date})

@app.command()
def main(birth: date):
    typer.echo(f"Interesting day to be born: {birth}")
    typer.echo(f"Birth day name: {day_name[birth.weekday()]}")
```

However, this would require more serious discussion and I'll be happy to open a dedicated issue if someone is interested 🙂

Most changes in this PR are tests / documentation related but I would like to say it was a pleasure to discover the powerful intrication between tests and documentation in place in this project 😃
This is well designed and very efficient to write 👏

### DIFF
--- a/docs/tutorial/parameter-types/date.md
+++ b/docs/tutorial/parameter-types/date.md
@@ -1,0 +1,84 @@
+You can specify a *CLI parameter* as a Python <a href="https://docs.python.org/3/library/datetime.html" class="external-link" target="_blank">`date`</a>.
+
+Your function will receive a standard Python `date` object, and again, your editor will give you completion, etc.
+
+```Python hl_lines="2  7 8 9"
+{!../docs_src/parameter_types/date/tutorial001.py!}
+```
+
+Typer will accept any string from the following formats:
+
+* `%Y-%m-%d`
+
+Check it:
+
+<div class="termy">
+
+```console
+$ python main.py --help
+
+Usage: main.py [OPTIONS] BIRTH:[%Y-%m-%d]
+
+Arguments:
+  BIRTH:[%Y-%m-%d][required]
+
+Options:
+  --install-completion  Install completion for the current shell.
+  --show-completion     Show completion for the current shell, to copy it or customize the installation.
+  --help                Show this message and exit.
+
+// Pass a date
+$ python main.py 1956-01-31
+
+Interesting day to be born: 1956-01-31
+Birth week name: Tuesday
+
+// An invalid date
+$ python main.py july-19-1989
+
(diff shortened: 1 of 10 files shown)


InCogNiTo124:
bump. this is what I stumbled upon today and would be nice to have

docs/tutorial/parameter-types/date.md
```diff
@@ -0,0 +1,84 @@
+You can specify a *CLI parameter* as a Python <a href="https://docs.python.org/3/library/datetime.html" class="external-link" target="_blank">`date`</a>.
+
+Your function will receive a standard Python `date` object, and again, your editor will give you completion, etc.
+
+```Python hl_lines="2  7 8 9"
+{!../docs_src/parameter_types/date/tutorial001.py!}
+```
```
svlandeg, line 7: Just a quick note here that if we do want to progress with this PR at some point, the docs should be updated with the new format, and using `Typer()` explicitely in the tutorial examples.

kozlek:
Long time no see ! 👋

I had a CLI app with a `date` parameter (to represent a birth day), which is a bit different from a `datetime` as it doesn't take care of timezone and time.
Having a `date` type is convenient because it matches exactly this use case, and it displays a proper format (`%Y-%m-%d`) when reading the help of the CLI.

Indeed, users may be confused to see that `%Y-%m-%dT%H:%M:%S` is a valid format to represent a birth day when using the built-in `datetime` type.

However, time has passed and Click custom types support has been been added to Typer: https://typer.tiangolo.com/tutorial/parameter-types/custom-types/#type-parser

Regarding this newly (2023) addition, I believe we can close this PR. Users seeking a `date` type are able to define their own Click type and use it using the `click_type` as specified in the documentation.

Thanks for your help !

PR #329, opened 2021-09-05
✨ Support for Pydantic models 

### LINKED ISSUE #111: [FEATURE] Support Pydantic Models as ParamTypes

### Additional context

Here's a simple example:

```python
#!/usr/bin/env python3
import click
import typer


class User(BaseModel):
    id: int
    name = "Jane Doe"


        return User.parse_raw(value)


USER = UserParamType()


    print(num, type(num))
    print(user, type(user))


if __name__ == "__main__":
    typer.run(main)
```
This currently throws:
```
    typer.run(main)
    app()
    return get_command(self)()
```

kdeyev:
Brief support for #111. 
- the plain Pydantic models are handled the same way as Tuples.
- more complicated Pydantic models are handles as FileText.

The implementation definitely can be improved. 
Please let me know what do you think.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,8 @@ classifiers = [
     "License :: OSI Approved :: MIT License"
     ]
 requires = [
-    "click >= 7.1.1, <9.0.0"
+    "click >= 7.1.1, <9.0.0",
+    "pydantic == 1.8.2"
 ]
 description-file = "README.md"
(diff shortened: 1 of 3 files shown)


tiangolo:
Thanks for the interest here! :coffee: 

For now, I'll close this one for cleanup. But we'll continue tracking this internally to work on it again once what's needed underneath is ready. Thanks! :cake:

PR #335, opened 2021-10-08
📝 Clarify testing documentation about checking `stderr`

cgabard:
Hello.
First thank for all your hard work on this project.

Today I wanted to test a small utility function and check that a message was printed into the standard error.

From the documentation, the only mention was 

> You could also check `result.stderr` for "standard error".

But using it directly, in fact, does not work.

After looking the code, it seems that Click's `CliRunner` mix standard output and error on `stdout` by default. You need to ask explicitly to do not mix them in order to be able to read the `result.stderr` property.

Then, I find the current documentation confusing because does not mention that and can lead reader to think it should work out of the box. In fact, I firsts started by double-checking my code before understand that the runner did not have the expected behavior.

So I suggest a small addition to the current note, at least to provide insight for readers.

Please let me know if I can improve it.


tiangolo:
Good point, thank you @cgabard! :cake: 

(and thanks @svlandeg for reviewing and labeling :nerd_face: )

Issue #311, opened 2021-08-12
Support for custom parameter type with `click.ParamType`


yashrathi-git:
Click have [click.ParamType](https://click.palletsprojects.com/en/8.0.x/parameters/#implementing-custom-types), for defining custom parameter types, but it doesn't work with typer (Example snippet below)

I want to use, my own formatting for datetime, ex: today %H:%M, %M:%H(automatically uses date as today). Typer already allows setting, [custom datetime formats](https://typer.tiangolo.com/tutorial/parameter-types/datetime/#custom-date-format) but it doesn't cover my use case.

This is example from [click docs](https://click.palletsprojects.com/en/8.0.x/parameters/#implementing-custom-types) that I tried using with **typer**.
```py
class BasedIntParamType(click.ParamType):
    name = "integer"

    def convert(self, value, param, ctx):
        if isinstance(value, int):
            return value

        try:
            if value[:2].lower() == "0x":
                return int(value[2:], 16)
            elif value[:1] == "0":
                return int(value, 8)
            return int(value, 10)
        except ValueError:
            self.fail(f"{value!r} is not a valid integer", param, ctx)

BASED_INT = BasedIntParamType()

def main(based_int: BASED_INT):
    print(based_int)
    
if __name__=="__main__":
    typer.run(main)
```
It gives this error:
```py
Traceback (most recent call last):
  File "/home/yashrathi/test.py", line 26, in <module>
    typer.run(main)
  File "/home/yashrathi/.local/lib/python3.9/site-packages/typer/main.py", line 859, in run
    app()
  File "/home/yashrathi/.local/lib/python3.9/site-packages/typer/main.py", line 214, in __call__
    return get_command(self)(*args, **kwargs)
  File "/home/yashrathi/.local/lib/python3.9/site-packages/typer/main.py", line 239, in get_command
    click_command = get_command_from_info(typer_instance.registered_commands[0])
  File "/home/yashrathi/.local/lib/python3.9/site-packages/typer/main.py", line 423, in get_command_from_info
    ) = get_params_convertors_ctx_param_name_from_function(command_info.callback)
  File "/home/yashrathi/.local/lib/python3.9/site-packages/typer/main.py", line 404, in get_params_convertors_ctx_param_name_from_function
    click_param, convertor = get_click_param(param)
  File "/home/yashrathi/.local/lib/python3.9/site-packages/typer/main.py", line 656, in get_click_param
    parameter_type = get_click_type(
[...]

PR #337, opened 2021-10-24
✨ Add support for custom `click.ParamType`s

dwreeves:
The following pull request adds support for custom `click.ParamType`s as requested in #311, with the implementation specified in https://github.com/tiangolo/typer/issues/311#issuecomment-932827651.

Example here:

```python
import json
from typing import Any, Dict, Optional, Union

import click
import typer


class JsonParamType(click.ParamType):
    name: str = "json"

    def convert(
        self,
        value: Optional[Union[Dict[Any, Any], str, bytes]],
        param: click.Parameter,
        ctx: click.Context,
    ) -> Optional[Dict[Any, Any]]:
        if isinstance(value, dict) or value is None:
            return value
        try:
            if isinstance(value, (str, bytes)):
                return json.loads(value)
            else:
                self.fail(f"{value!r} was not a str, bytes or None.", param, ctx)
        except OSError as e:
            self.fail([*e.args, None][0], param, ctx)
        except json.JSONDecodeError as e:
            self.fail(f"Bad JSON: {[*e.args, None][0]}", param, ctx)


JSON = JsonParamType()


def main(data: dict = typer.Option(..., param_type=JSON)):
    if "what_i_like" in data:
        typer.echo(f"You like {data['what_i_like']}? Me too!")
    else:
        typer.echo("What do you like?")


if __name__ == "__main__":
    typer.run(main)
```

The above implementation uses the `param_type=?` kwarg to implement a `click.ParamType`. When the ParamType is not specified, the existing behavior is used, i.e. this behavior is backwards compatible.

I added some `assert` statements to provide helpful error messages for users who are using this feature.


dwreeves:
Thanks for the heads up @archiif. I'll look into it within the next few days.

dwreeves:
@archiif TLDR: the issue is that Click resolves argument counts with `click.Parameter.nargs`, whereas the type is resolved with `click.Parameter.type`.

In Typer, both `nargs` _and_ `type` are inferred from the function parameter's type annotation.

So basically, if we ignore the parameter's type annotation, we are ignoring both how Typer resolves the `click.Parameter.type` _and_ how it resolves the `click.Parameter.nargs`, but I only exposed a way to overwrite `type` and not `nargs`.

---

There are a couple ways to resolve this.

### Option # 1: Overhaul Typer internals

My personal opinion, but I think the "best" way, in the longer-term, would be a complete rework of Typer's internals to expose an API that converts type annotations into `click.Parameter`s. Basically, `get_click_param()` should be a class that can be overridden, with a bunch of methods that can all be overwritten modularly. Similar to how Click's internals work.

There are a couple problems with the above approach:

- Obviously, it's a lot of work!
- Because it's such a significant API change, it would require @tiangolo to delegate a lot of responsibility or take some of it on himself. (For what it's worth-- `if-elif-elif-elif`... type conversion is also a longer term problem with SQLModel internals, so coming up with a good system in either repo will help the development of the other). And I don't think right now he is prioritizing this repo. I'm not blaming him for this, open source is thankless work, but that's the reality of the situation.

Because it's so much work and involves so much discretion in API design, I don't think this is appropriate until this project is more actively maintained.

### Option # 2: Add an optional `nargs` kwarg to `typer.Argument` and `typer.Option`.

Exposing `nargs` also helps in situations where e.g. you have a `List[str]` and want `nargs` to be == 2 or something like that. (Since right now Typer can't handle `Tuple[str, str]`).

There is a decision to be made on whether setting `nargs=` should work, be ignored, raise a warning + work, raise a warning + be ignored, or raise an exception.

### Option # 3: Always respect `list[...]` as being variadic.

IMO, this is probably the best solution in the short-term for the reason that it involves the fewest changes to the API.

That said, it's far from perfect and comes with limitations that option \#2 doesn't have. Compared to solutions 1 and 2, it comes with the downside that the user loses some control over the resolution of . Imagine for example that you have some type that looks like this:

```python
class SemicolonDelimitedListParamType(click.ParamType):
    name = "semicolon_delimited_list"
        return value.split(";")
```

^ You probably want this type to output as a `List[str]`, however Typer will output a `List[List[str]]`. And then Typer would be expected to maintain backwards compatibility with this behavior using Typer's defaults.

AlirezaTheH:
@tiangolo Merge this please🙏

AlirezaTheH:
@tiangolo Do you have any plan to merge this? it's really needed.

tiangolo:
Thanks a lot for the work @dwreeves! :rocket: 

Support for this was added in https://github.com/tiangolo/typer/pull/583, the new docs are here: https://typer.tiangolo.com/tutorial/parameter-types/custom-types/, it's available in Typer `0.8.0`, just released. :tada: 

Given that, I'll close this one. But thanks again! And if you find something funky there, please let me know. :nerd_face: 

Issue #88, opened 2020-04-12
[QUESTION] How to use with async?


csheppard:
### First check

* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the Typer documentation, with the integrated search.
* [x] I already searched in Google "How to X in Typer" and didn't find any information.
* [x] I already searched in Google "How to X in Click" and didn't find any information.

### Description

I have existing methods/model functions that use async functions with [encode/databases](https://github.com/encode/databases/) to load data but I'm unable to use these within commands without getting errors such as `RuntimeWarning: coroutine 'something' was never awaited`

How can I make make my `@app.command()` functions async friendly?

csheppard:
Just found https://github.com/pallets/click/issues/85#issuecomment-503464628 which may help me

thedrow:
It'd be nice if we could use Typer with the async-click fork.

mreschke:
Seeing as how FastAPI is an async framework, having an async CLI seems logical.  The main reason being sharing code from the CLI and Web entry points.  You can of course use the asgiref.sync.async_to_sync converter helpers to call existing async methods from the CLI but there are complications here and it makes your cli code clunky.  I replaced typer with smurfix/trio-click (which is asyncclick on Pypi) and it works great, but of course this is just async click, not the cool typer implementation.  Forking typer and replaceing all `import click` with `import asyncclick as click` works like a charm but it means maintenance of the fork yourself.  If @smurfix could keep his async fork maintained and up to date with click upstream, and if typer was based on asyncclick then we would really have something great here.

mreschke:
Thanks @smurfix.   Once you have asyncclick updated, if @tiangolo doesn't have a nice async typer by then, perhaps ill make a good `typer-async` fork of typer and use asyncclick and add to pypi for us all to use.

jessekrubin:
> Thanks @smurfix. Once you have asyncclick updated, if @tiangolo doesn't have a nice async typer by then, perhaps ill make a good `typer-async` fork of typer and use asyncclick and add to pypi for us all to use.

@mreschke I made a pull request to this repo that gets you most of the way to async. https://github.com/tiangolo/typer/pull/128

amfarrell:
@mreschke I've updated @jessekrubin's PR to remove the conflicts with master, in case you find it useful.

mreschke:
[...]

PR #340, opened 2021-11-19
✨ Async support optionally using anyio

skeletorXVI:
I took a shot at supporting async as discussed in #88 while integrated feedback from the issue #88 and PR #332 . Unlike the existing proposal I decided to support both sync and async commands from the same decorator, as I don't see a reason to use two identical decorator functions.

`anyio` is listed as an extra dependency. When anyio is not installed, asyncio is used, When `anyio` and `trio` are installed, the default engine becomes trio.

Other async engines, or customized behavior can be used by providing a custom run function via the `async_runner` parameter to the decorator or Typer instance.

- [x] Support asyncio, anyio and trio
- [x] Anyio is only an optional dependency
- [x] Autocompletion works as expected
- [x] Updated documentation
- [x] Added unit tests


dekoza:
@tiangolo will this reach `main`? :sweat_smile:

svlandeg (commented):
Hi @skeletorXVI and @xqSimone!

I appreciate all the work and effort you've put into this PR, including adding extensive tests and documentation. As async support is still high on our priority list, I've started looking into the best way of supporting it.

I'm going to work a bit on your PR to get it back up-to-date and focus on the async functionality only, to check whether this is the direction we want to move in. I'll be pushing to your branch directly, I hope you don't mind! By doing so, I hope to preserve a lot of the work you've already put into this 🙏

skeletorXVI:
> Hi @skeletorXVI and @xqSimone!
> 
> I appreciate all the work and effort you've put into this PR, including adding extensive tests and documentation. As async support is still high on our priority list, I've started looking into the best way of supporting it.
> 
> I'm going to work a bit on your PR to get it back up-to-date and focus on the async functionality only, to check whether this is the direction we want to move in. I'll be pushing to your branch directly, I hope you don't mind! By doing so, I hope to preserve a lot of the work you've already put into this 🙏

This sounds great, I haven't had time or the need to update as our CLI was working perfectly fine. But as I literally yesterday got a task in, that requires to add a CLI to a new project that uses asyncio and I didn't want to use a very old branch.

As such I would appreciate if this could pick up some steam and be merged into the actual package soon.

If there is need for any assistance just ping me, while I have higher priority tasks to work on right now, I will be able to dedicate some hours to this over the coming weeks.

.github/workflows/test.yml
```diff
@@ -22,8 +22,6 @@ jobs:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
         python-version: [ "3.13" ]
         include:
-          - os: ubuntu-22.04
-            python-version: "3.7"
```
svlandeg: This will be done properly in https://github.com/fastapi/typer/pull/830, but I don't want the CI here to be red just because of Python 3.7.

docs/tutorial/async.md
```diff
@@ -0,0 +1,88 @@
...
+{!../docs_src/asynchronous/tutorial001.py!}
+```
+
+Or using `anyio`:
+
+```Python
+{!../docs_src/asynchronous/tutorial002.py!}
+```
+
+<small>Important to note, `typer.run()` doesn't provide means to customize the async run behavior.</small>
```
svlandeg: This is not saying much - personally I would either remove this statement, or add more details on what is missing / can't be customized.

svlandeg (commented):
Just a quick note here that if we want to progress with this PR at some point, we should update the tutorial examples to use `Typer()` explicitely, as in https://github.com/fastapi/typer/pull/1418.

tiangolo:
Thanks for the interest here! ☕

For now, I'll close this one for cleanup. But we'll continue tracking this internally to work on it again once what's needed underneath is ready. Thanks! 🍰

PR #355, opened 2022-02-12
Add callable support to `show_default`

### LINKED ISSUE #158: [BUG] Type hint for show_default of typer.Option is incorrect
### Describe the bug

For typer.Argument, the type hint for show_default is this: show_default: Union[bool, str] = True,
for typer.Option, the type hint is this: show_default: bool = True,

For an argument, I can override the text that accompanies the default value to be something like "working directory". This 
permits me to do document generation including usage blocks that don't have paths from the build system in them.

For an option, I can do the same but I need to suppress type check. The functionality works regardless of the type hints. 

### To Reproduce

Run mypy against this code:

```
def docker_build(context_path: Path = typer.Argument(Path.cwd(), exists=True,
                                                     show_default=working_directory_label,
                                                     resolve_path=True),
                           dockerfile: Path = typer.Option(Path.cwd() / "Dockerfile", "--file", "-f", exists=True, dir_okay=False,
                                                                              show_default="Dockerfile in working directory", resolve_path =True)):
    pass
```

### Environment

* OS: Centos7
* Typer Version 0.3.1
* Python version: 3.6.8

zanieb:
Adds callable support to `show_default`, allowing a `Callable[[], Union[bool, str]]` to be passed to dynamically determine the default to show in help messages. Closes https://github.com/tiangolo/typer/issues/354.

Previously, `show_default` was incorrectly typed as just `bool` instead of `Union[bool, str]` on `Option` / `OptionType`. This fixes those type annotations. Closes #158.

Additionally, the default string for dynamic options was defined as `default_string = _("(dynamic)")` which appears to be a bug and is confusing syntax if valid. This was simplified to match the line from arguments.

I've also removed the formatted parenthesis from the displayed `show_default` value, as this value should be able to match the style of a real default. This is the only "breaking" change here.

I'll happily extend documentation / tests if you're willing to accept this.

### DIFF
--- a/typer/core.py
+++ b/typer/core.py
@@ -82,7 +82,6 @@ def _typer_param_setup_autocompletion_compat(
 
         self._custom_shell_complete = compat_autocompletion
 
-
 class TyperArgument(click.core.Argument):
     def __init__(
         self,
@@ -106,7 +105,7 @@ class TyperArgument(click.core.Argument):
         ] = None,
         autocompletion: Optional[Callable[..., Any]] = None,
         # TyperArgument
-        show_default: Union[bool, str] = True,
+        show_default: Union[bool, str, Callable[..., Union[bool, str]]] = True,
         show_choices: bool = True,
         show_envvar: bool = True,
         help: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


madkinsz:
Note this not only fixes the type hint for `show_default` but also adds callable support to `show_default` which is not just a type hint change. Just want to clarify since the label is a bit more narrow.

madkinsz:
This conflicts with the refactor of defaults for `rich_utils`. Unfortunately it does not appear straight-forward to reincorporate this improvement. Given that I haven't gotten a response yet, I won't attempt to do so yet. If you're willing to accept this feature enhancement please let me know and I will attempt to resolve the conflicts.

svlandeg:
[maintenance note] I'd suggest reviewing https://github.com/tiangolo/typer/pull/501 first and then updating this PR to review the added feature.

PR #360, opened 2022-02-23
🐛 Fix PowerShell completion with incomplete word

patricksurry:
Fixes https://github.com/tiangolo/typer/issues/359

Standard example in documentation failed because incomplete word was not removed from cwords, so typer thought we were adding an additional parameter which wouldn't be allowed.  Completing on a `List[]` parameter appeared to work because then a subsequent parameter would be allowed.  

Also handles the case where we are completing somewhere other than the end of the line, by ignoring text past the cursor position and then completing normally.


svlandeg (commented):
I can confirm the bug with Powershell: there are no auto-completion suggestions when typing "Ca" or "Se" on `master`. This PR resolves the issue.

Two points I'd like to address/look into before we could merge this PR:

- Is the `_TYPER_CURSOR_POSITION` variable actually necessary? The buggy behaviour is still solved on my system if I remove the `if cursor` lines and only keep the fix of the `args` definition
- Ideally we'd have a unit test that captures this behaviour and fails on `master`. I'm looking into this.

svlandeg:
Update:

- I've removed the _TYPER_CURSOR_POSITION variable from this PR. With the current PR, Powershell completes in the middle of the line even without it, so for now I think we should go with the most minimal fix and potentially reconsider building on top of it if we do run into more issues afterwards
- I've added a unit test that fails on `master` and succeeds on this PR.

With that, I think this is good to merge.

tiangolo:
Great, thank you for the contribution (and the patience) @patricksurry! :cake: 

And thank you @svlandeg for the thorough review, tweaks, and help. :bow: :coffee: 

PR #362, opened 2022-02-26
Multivalued Option that can be used multiple times

stereobutter:
Hi there 🙋🏼‍♂️, 
this is my first PR on `typer` and it adds support for a *cli option* that can be used **multiple times** and has **multiple values** by providing a type hint of the form `list[tuple[...]]`. This is something that I found this morning that works in `click` by setting `multiple=True` together with `nargs` and noticed there is no native support for in typer. 

The documentation and tests for this I copied over and edited accordingly from the `multiple_options` use case. The entry in the documentation _"multiple options with multiple values"_ is a real mouthful but I couldn't think of something better that didn't include some scary language like _"variadic option"_. Any feedback on this (or any other aspect of the PR) is very welcome 😊


tiangolo:
Thanks for the interest here! ☕

Let's start with a Github Discussion with the request for this, the specific use case and the example code you would imagine, and then we can continue from there.

For now, I'll close this one for cleanup. But we'll continue tracking this internally to work on it again once what's needed underneath is ready.

Thanks! 🍰

PR #380, opened 2022-03-30
🐛 Fix import of `get_terminal_size` for click 8.1.0 support and upgrade Black for CI

tiangolo:
🐛 Fix import of `get_terminal_size` for click 8.1.0 support and upgrade Black for CI

Continuation of https://github.com/tiangolo/typer/pull/375 handling CI.


madkinsz (approved):
Funny the lint CI failed due to the same issue :) LGTM. Thanks for taking the time!

henryiii:
Black is affected by a different removal from click 8.1 (one that started with an underscore!), but similar.

tiangolo:
Thanks for the initial work/PR @madkinsz! And thanks for the feedback @henryiii!

This will be available in Typer `0.4.1`, released in the next hour. :rocket: :tada:

PR #384, opened 2022-04-16
👷 Add CI for Python 3.10

tiangolo:
👷 Add CI for Python 3.10


tiangolo:
Ah! Let me try upgrading that.

PR #386, opened 2022-04-19
🎨 Fix header format for "Standard Input" in `docs/tutorial/printing.md`

briancohan:
Based on the TOC and contents of the file, `"Standard Input"` does not appear to belong as a sub-item of `"Standard Output" and "Standard Error"`. Header was changed from h3 `###` to h2 `##`.


svlandeg (approved):
I agree, "Standard Input" and "Standard output" should be at the same level :-)

Thanks for the PR!

tiangolo:
Ah! Nice catch! That wasn't an obvious one. :nerd_face: Thanks @briancohan! :cake: 

And thank you @svlandeg! :cookie: :coffee:

PR #396, opened 2022-05-12
fix typing

leirons:
Argument join_options has incompatible type Sequence[str], but expected List[str]


tiangolo:
Thanks for the effort! :coffee: I suspect this was incorrectly typed in Click, but it seems to be correctly typed now, so it seems to be okay now. Given that, I'll close this one. But thanks! :cake:

PR #398, opened 2022-05-17
Add documentation for enums created via the functional API

Torvaney:
This PR updates the documentation to include discussion of enumerating choices with `enum.Enum`s created with the functional API. This is useful when creating choices dynamically, for example logging.

See further discussion of enums and the functional API in https://github.com/tiangolo/typer/issues/389

### DIFF
--- a/docs/tutorial/parameter-types/enum.md
+++ b/docs/tutorial/parameter-types/enum.md
@@ -66,3 +66,16 @@ Training neural network of type: lstm
 ```
 
 </div>
+
+
+### Functional API
+
+In order to use an `Enum` created using the <a href="https://docs.python.org/3/library/enum.html#functional-api" class="external-link" target="_blank">functional API</a>, you need to create an enum with string values.
+
+You also need to supply the default value as a string (not the enum):
+
+```Python hl_lines="5 9"
+{!../docs_src/parameter_types/enum/tutorial003.py!}
+```
+
+Alternatively, you can create an `Enum` that extends both `str` and `Enum`. In Python 3.11+, there is <a href="https://docs.python.org/3.11/library/enum.html#enum.StrEnum" class="external-link" target="_blank">`enum.StrEnum`</a>. For Python 3.10 or earlier, there is the <a href="https://github.com/irgeek/StrEnum" class="external-link" target="_blank">StrEnum package</a>.
--- a/docs_src/parameter_types/enum/tutorial003.py
+++ b/docs_src/parameter_types/enum/tutorial003.py
@@ -1,0 +1,18 @@
+from enum import Enum
+
+import typer
+
+NeuralNetwork = Enum(
+    "NeuralNetwork",
+    {k: k for k in ["simple", "conv", "lstm"]}
+)
+
+
+def main(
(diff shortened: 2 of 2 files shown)


svlandeg (approved):
Hi! Thanks for this contribution, and apologies for the delay in reviewing.

I agree that this additional section in the documentation could make it more clear for users on how to use "pure" Python `Enum`'s with Typer. 

I updated the PR with the latest from `master` and also included an Annotated version.

It's probably good if Tiangolo has another look at the exact phrasing in the documentation. I'll leave the final decision to merge or not with him.

docs_src/parameter_types/enum/tutorial004_an.py
```diff
@@ -0,0 +1,16 @@
+from enum import Enum
+
+import typer
+from typing_extensions import Annotated
+
+NeuralNetwork = Enum("NeuralNetwork", {k: k for k in ["simple", "conv", "lstm"]})
```
YuriiMotov, line 6: mypy gives me
```
main.py:6: error: Second argument of Enum() must be string, tuple, list or dict literal for mypy to determine Enum members  [misc]
```

docs_src/parameter_types/enum/tutorial004_an.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def main(
+    network: Annotated[NeuralNetwork, typer.Option(case_sensitive=False)] = "simple",
```
YuriiMotov, line 13: mypy gives:
```
main.py:10: error: Incompatible default for argument "network" (default has type "str", argument has type "NeuralNetwork")  [assignment]
```

joaonc: IIRC, this only works if the Enum extends `str`.   You can either use `StrEnum` or have an something like `class MyEnum(str, Enum)`

docs_src/parameter_types/enum/tutorial004_an.py
svlandeg: This particular error is a bit of a `mypy` quirck, if you have it as 
```
NeuralNetwork = Enum("NeuralNetwork", {"simple": "simple", "conv": "conv", "lstm": "lstm"})
```
then it does pass the type check.

docs_src/parameter_types/enum/tutorial004_an.py
svlandeg: Ah, that's unfortunate. We already have examples in the tutorial using `Food(str, Enum)`, no need to repeat that here. And the current example does work, even if `mypy` complains. So I'm a bit on the fence what to do here. I guess our options are:

- merge as-is even if `mypy` complains (we don't test for it on the CI anyway, but good catch @YuriiMotov!)
- merge without the `Annotated` example
- don't merge and assume there are enough examples to use `Enum` and users will figure out what's best for them

YuriiMotov: I think it's better to "merge without the Annotated example".
Alternative would be to add `# type: ignore[assignment]` and explain that for now we have to ignore this error with `Annotated` approach

Issue #341, opened 2021-11-19
[QUESTION] why replace "_" with "-" in command name?


bloodynumen:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional

import typer

app = typer.Typer()


@app.command()
def hello_test(name: Optional[str] = None):
    if name:
        typer.echo(f"Hello {name}")
    else:
        typer.echo("Hello World!")

if __name__ == "__main__":
    app()
```

### Description

expect: `python main.py hello_test --name world`
actually: `python main.py hello-test --name world`

I found the "_" in command name is replaced with "-". 
source code [https://github.com/tiangolo/typer/blob/master/typer/main.py#L390](url)

why do this? 

Thank u

### Operating System

macOS

### Typer Version

0.4.0

### Python Version

3.9.7

daddycocoaman:
This is probably just a design choice based around natural CLI convention. You're usually gonna find dashes in commands in any CLI, not underscores. If you want to keep the underscore, you can just specify the command name manually in the app.command().

mbopfNIH:
This type of undocumented behavior is why I have decided not to use Typer. I like a lot of things about the package, but it's not worth the learning curve of digging through the code when something does work as expected.

targhs:
I guess that helps

manueldeprada:
If fact, this is something that comes from the "click" package which typer uses as a base. If it helps, it comes from here:

https://github.com/pallets/click/blob/dc918b48fb9006be683a684b42cc7496ad649b83/src/click/core.py#L2633

manueldeprada:
> Yeah this is pretty annoying. Is there any way to disable this behaviour? There doesn't seem to be anything in `typer.Option()`. In fact I couldn't find any reference documentation at all...?

I don’t think so, since it comes from click. I switched to Python Fire by google.

Timmmm:
[...]

PR #403, opened 2022-06-10
📝 Add note about how subcommands with function names using underscores are converted to dashes

targhs:
Docs updated for #341


tiangolo:
Makes sense, thanks @targhs! :cake:

PR #408, opened 2022-06-24
👷‍♂️ Consistently use `sys.executable` to run subprocesses, needed by OpenSUSE

theMarix:
In some scenarios, like running pytest outside of an active virtualenv,
it could happen that "coverage" would refer to a different Python
interpreter than the one running the tests. Admittedly that does not
happen in standard development scenarios where you use a virtualenv. But
it easily happens when packaging for Linux distributions that support
multiple versions of Python and it can also happen when running pytest
from a virtualenv without activating it. The latter is something that's
convenient when testing versus multiple Python versions without using
tox.

Explicitly invoking coverage as a module on sys.executable ensures that
the same binary, not process, that's also running the tests is used.
This was in fact already done in some tests but not consistently across
all of them.

### DIFF
--- a/tests/test_compat/test_option_get_help.py
+++ b/tests/test_compat/test_option_get_help.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import sys
 
 import typer.core
 from typer.testing import CliRunner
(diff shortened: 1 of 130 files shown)


theMarix:
Curiously, the build failures triggered by this PR are exactly the ones that #407 fixes. Not sure how best to proceed here. Ideally we can agree to merge #407 and than I'd rebase this onto master and everything will _just work™_.

tiangolo:
Cool, thanks! Let's do it! 🚀

PR #417, opened 2022-07-10
✨ Improve support for CLI translations uisng gettext

mjodmj:
Just adding some missing wrappings of strings around gettext which some may need for the sake of CLI i18n.


ereslibre (approved):
Just chiming in to say that it would be nice to have this one merged :)

Edit: this PR might also consider removing the `TODO` about this very change: https://github.com/tiangolo/typer/blob/046d3e2b4cb1f2feb6d94b856aa75e013d9896c6/typer/core.py#L623-L626

svlandeg (approved):
Thanks for your contribution!

I took the liberty of updating your PR with the latest changes from `master`. This looks good to me.

ereslibre:
Thanks @svlandeg! I do wonder if we should also update some of the strings at https://github.com/tiangolo/typer/blob/master/typer/rich_utils.py#L83-L93

If you are fine with it I can create a PR if you don't want to update this one with those.

svlandeg:
Good idea to add those @ereslibre - I've pushed directly to this PR to keep the work together. Feel free to review!

tiangolo:
Great, thank you @mjodmj! :nerd_face: 

And thanks for the review @svlandeg :bow: 

This will be available in Typer 0.9.4 in the next few hours. :rocket:

ereslibre:
@tiangolo did beat me to it! Sorry for the lag on the review and thank you everyone involved :)

PR #418, opened 2022-07-11
Added dark theme option and auto theme switching

gbdlin:
- Added dark theme option
- Configured automatic theme switching based on `prefers-color-scheme` media query.


svlandeg (commented):
Hi @gbdlin, thanks so much for the contribution!

It looks like in the meantime, dark mode for the docs has been introduced by Tiangolo in PR #908 🎉 As such, I think we can close this one, but thanks again! 🙏

PR #425, opened 2022-07-14
🚸 Improve assertion error message if a group is not a valid subclass

chrisburr:
With the latest typer release my CI (which runs a `typer` based script) started failing with:

```python
  File "/builds/cburr/DIRAC/./integration_tests.py", line 750, in <module>
    app()
AssertionError
```

The cause turned out to be:

https://github.com/tiangolo/typer/blob/fac64ca03e9fda619710702c3364e233e0e644ac/typer/main.py#L505

which was triggered by my script contianing:

```python
class NaturalOrderGroup(click.Group):
    """Group for showing subcommands in the correct order"""

    def list_commands(self, ctx):
        return self.commands.keys()
```

The fix is easy and just requires replacing `click.Group` with `typer.core.TyperGroup` however I think the assertion error should be made this clearer.


tiangolo:
Nice, thank you @chrisburr! :rocket:

PR #435, opened 2022-07-21
add command_group decorator

Soumeh:
Slightly inspired by [Nextcord's subcommand system](https://docs.nextcord.dev/en/stable/interactions.html?highlight=subcommand#subcommands), this decorator could be used to simplify subcommand creation.

My proposed example based on the [reigns and towns example](https://typer.tiangolo.com/tutorial/subcommands/nested-subcommands/#manage-the-land-in-a-cli-app).
```py
import typer

app = typer.Typer()


@app.command_group()
def reigns():
    pass

@reigns.command()
def conquer(name: str):
    print(f"Conquering reign: {name}")

@reigns.command()
def destroy(name: str):
    print(f"Destroying reign: {name}")


@app.command_group()
def towns():
    pass

@towns.command()
def found(name: str):
    print(f"Founding town: {name}")

@towns.command()
def burn(name: str):
    print(f"Burning town: {name}")


if __name__ == "__main__":
    app()
```

Which works exactly like the `lands.py` script from the lined example, except with less code.

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -257,6 +257,67 @@ class Typer:
 
         return decorator
 
+    def command_group(
+        self,
+        name: Optional[str] = Default(None),
+        *,
+        cls: Optional[Type[TyperGroup]] = Default(None),
+        invoke_without_command: bool = Default(False),
+        no_args_is_help: bool = Default(False),
+        subcommand_metavar: Optional[str] = Default(None),
+        chain: bool = Default(False),
+        result_callback: Optional[Callable[..., Any]] = Default(None),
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@Soumeh-zz, thanks for your interest and efforts!

I think that this approach with `command_group` decorator is less flexible and will lead to bad design..

How would this code look like if you decided to place `reigns` and `towns` in different files (as it's done in docs)?
I can imagine people importing main app into those modules to use `@app.command_group()` decorator. I think that would create a mess..

Issue #450, opened 2022-09-01
Unexpected behavior of no_args_is_help with 0 or 1 commands


JacobBush:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer

cli = typer.Typer(
    no_args_is_help=True,
    help="should be displayed when run without args",
)


if __name__ == "__main__":
    cli()
```

### Description

When creating a Typer application with `no_args_is_help=True`, I would expect the help text to be displayed when the CLI is called without arguments.

```
import typer

cli = typer.Typer(
    no_args_is_help=True,
    help="should be displayed when run without args",
)

if __name__ == "__main__":
    cli()
```

```
$ python cli.py
AssertionError: Could not get a command for this Typer instance
```
However, because there are no commands registered, typer raises an AssertionError rather than display the help text in this case.

If we add a command, I would expect the CLI to show the help text when it is called without arguments.
```py
@cli.command("1") 
def _1():
    print(1)
```

```
$ python cli.py
1
```
When invoked from the command line, the program prints `1` and exits, rather than the expected behaviour of printing the help text. I think this behaviour (invoking the sole command when called with no arguments) _could_ make sense as a default if `no_args_is_help` was not set, but in this case it does not feel correct.

To make this more confusing, if you add another command the help text works as expected. The default behaviour changes to no longer run the "1" command.
```py
@cli.command("2") # rest of the code is the same
def _2():
    print(2)
```

```
$ python cli.py
Usage: cli.py [OPTIONS] COMMAND [ARGS]...

  should be displayed when run without args

Options:
  --install-completion [bash|zsh|fish|powershell|pwsh]
                                  Install completion for the specified shell.
  --show-completion [bash|zsh|fish|powershell|pwsh]
                                  Show completion for the specified shell, to
                                  copy it or customize the installation.
  --help                          Show this message and exit.

Commands:
  1
  2
```

### Operating System

macOS

### Typer Version

0.6.1

### Python Version

3.9.12

tiangolo:
Thanks for the report!

[...]

PR #451, opened 2022-09-06
🐛 Fix `no_args_is_help` for 0 and 1 command Typer Apps

JacobBush:
Resolves https://github.com/tiangolo/typer/issues/450


Migi:
For what it's worth, I've reviewed this PR and it looks good to me. But I'm not a maintainer of this project, just someone who encountered this same bug.

svlandeg:
Good catch - it does look like this is a bug we should resolve. We'll look into this.

[maintenance note] As this feature didn't have any documentation examples or tests, I suggest to review/merge https://github.com/tiangolo/typer/pull/751 first.

svlandeg (commented):
Thanks again for this contribution! 

I reviewed the cases outlined in #450 and confirmed that they break on `master` and are fixed with this PR. Further, the unit test you've added here seems to accurately cover these cases. So in terms of behaviour/functionality this PR looks good. 

Thee only thing I'm unsure about is whether we want/need to fix this in the `get_command` function, and whether there might not be unexpected side effects from this fix. I'll leave the final review for Tiangolo.

jonboulle:
Any chance of revisiting this? :eyes:

tiangolo:
Thanks @JacobBush! I would consider things to be running as expected.

In short, the main case with a single sub-command would be:

```Python
import typer

cli = typer.Typer(no_args_is_help=True)


@cli.command("1")
def _1():
    print(1)


@cli.callback()
def callback():
    pass


if __name__ == "__main__":
    cli()
```

It's documented here: https://typer.tiangolo.com/tutorial/commands/#explicit-application

There's a bit more info in the issue/discussion.

If there's something else you would like to consider, let's continue in the discussion. :nerd_face: 

For now, I'll pass on this one, but thanks for the effort! :cake:

PR #458, opened 2022-09-20
feat: Add styling of the command in the suggestion

mortenengen:
This allows the user to add custom styling of the command in the try statement that is printed in case of an error.


YuriiMotov:
@mortenengen, thanks for your interest and efforts!

Currently it's already done by

https://github.com/fastapi/typer/blob/549596c0d77360e13247c0cc819cb8449b702dcf/typer/rich_utils.py#L709-L715

https://github.com/fastapi/typer/blob/549596c0d77360e13247c0cc819cb8449b702dcf/typer/rich_utils.py#L100

mortenengen:
Thanks for the update @YuriiMotov 👍

Issue #427, opened 2022-07-14
click.utils has no attribute _expand_args (click version 7.1.2, typer version 0.6.1)


falcomomo:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer

def main(name: str):
    typer.echo(f'Hello {name}')


if __name__ == '__main__':
    typer.run(main)
```

### Description

* run python typer_script.py
* On the typer.run() call: AttributeError: 'click.utils' has no attribute '_expand_args'

I was on typer version 0.6.1 and click version 7.1.2 (which meets the requirements for typer 0.6.1)

I updated to click version 8.0.4 and it's ok now.

In typer.core:190 you call click.utils._expand_args, but this was only added in click major version 8 from what I can tell. 

### Operating System

Windows

### Typer Version

0.6.1

### Python Version

3.6.13

### Additional Context

None

parvva:
Is there any update on this?

texasfight:
In the meantime, you can just add something along the lines of 
`click>=8.0.0,<9.0.0` 
to your requirements.txt, setup.py, etc. depending on your use-case.

I opened a PR and hopefully @tiangolo can see this at some point even though it seems to only be affecting a few of us.

I wouldn't care that much except for the fact that CLIs are generally going to be installed outside of a virtualenv and other libraries may have previously installed a lower version (like what I ran into).

PR #463, opened 2022-10-06
Bump `click` minimum version to 8.0.0

texasfight:
Click major version 7 does not properly work because of the following error: `AttributeError: module 'click.utils' has no attribute '_expand_args'`

The `_expand_args` method was added in click major version 8.

Fixes issue: #427

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,6 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
-        click-7: [true, false]
       fail-fast: false
 
     steps:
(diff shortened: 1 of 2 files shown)


texasfight:
@tiangolo Currently typer.core:190, which is causing issues, is being ignored by a `pragma: no cover`. Furthermore, the tests never fail since all of the CI/CD testing pipelines are using Ubuntu and not Windows. 

As an alternative to bumping the required version, we could just implement the logic in click.utils._expand_args ourselves since it just uses the `glob` and `os` builtin libraries.

RileyMShea:
It seems for the current version of typer(0.9.0), only `click=8.0.4` is working completely.

Click < 7.x and click >8.1.x both have their own problems with type atm.

- 7.x https://github.com/tiangolo/typer/pull/463
- 8.1.x https://github.com/tiangolo/typer/issues/377

Recommend further narrowing click range down to: `"click >= 8.0.0, <8.1.0"`

LeXofLeviafan:
> It seems for the current version of typer(0.9.0), only `click=8.0.4` is working completely.
> 
> Click < 7.x and click >8.1.x both have their own problems with type atm.
> 
> Recommend further narrowing click range down to: `"click >= 8.0.0, <8.1.0"`

Click 8.1 appears to be supported since Typer 4.1.0 release ([according to release notes](https://github.com/tiangolo/typer/blob/master/docs/release-notes.md#041)). Have you checked if there's actually any issue with Click 8.1?

svlandeg:
Hi! Apologies for the late follow-up. Tiangolo is working on a PR that includes the edits from this one, so I'll go ahead and close this one in favour of https://github.com/tiangolo/typer/pull/760. Thanks for the contribution & discussion though!

PR #464, opened 2022-10-07
✏ Fix link in docs to `datetime.strptime()`

Kobu:
**Issue:**
Documentation claims `formats receives a list of strings with the date formats that would be passed to datetime.strptime()` however the actual link is pointing to `date.strftime()` as can be seen [here](https://typer.tiangolo.com/tutorial/parameter-types/datetime/#custom-date-format)

**Fix**
Replaced the incorrect link by one thats pointing to `datetime.strptime()`


tiangolo:
Good catch, thanks @Kobu! :rocket:

PR #475, opened 2022-11-04
✏️ Add quotes to package name that includes brackets in docs

gjolga:
From poetry's docs: 

>Some shells may treat square braces ([ and ]) as special characters. It is suggested to always quote arguments containing these characters to prevent unexpected shell expansion.
https://python-poetry.org/docs/cli/#add


tiangolo:
Great, thanks @gjolga! 🍰

PR #484, opened 2022-11-05
➕ Bring back pytest-cov because coverage can't detect pytest-xdist

tiangolo:
➕ Bring back pytest-cov because coverage can't detect pytest-xdist


tiangolo:
This still needs some tweaking because coverage is still not detecting everything although everything is obviously still running.

PR #486, opened 2022-11-05
Simplify `comment-docs-preview-in-pr` action by removing `httpx` and use only `pygithub`

Valaraucoo:
**Why?**
* `pygithub` allows you to create issue comments by using `pull_request.create_issue_comment` method
* we can remove `httpx` library from action dependencies


YuriiMotov:
@Valaraucoo, thanks for your interest and efforts!

This action has already been refactored and now doesn't use `httpx`: https://github.com/fastapi/typer/blob/master/scripts/deploy_docs_status.py

Issue #428, opened 2022-07-14
Order of command panels for rich help output


mike0sv:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer

app = typer.Typer(rich_help_panel_order=["first", "second"])


@app.command(rich_help_panel="second")
def create():
    print("Creating user: Hiro Hamada")


@app.command(rich_help_panel="first")
def delete():
    print("Deleting user: Hiro Hamada")


if __name__ == "__main__":
    app()
```

### Description

New typer version introduced `rich_help_panel` option for commands and groups, which groups together different commands for multi-command help message. However there is no way to set the order of different panels (it's semi-random depending on the order of `panel_to_commands` dict inside `rich_format_text` function). It would be nice to have an option to configure this order

### Wanted Solution

see code example above

### Wanted Code

```python
see code example above
```

### Operating System

Other

### Typer Version

0.6.1

### Python Version

3.9.4

johnthagen:
In my own code, I use an `Enum` for this:

```py
@unique
class HelpPanel(str, Enum):
    Creation = "Installation"
    Configuration = "Configuration"
    Startup = "Startup"
```

I wonder if the API used an `Enum` if the order could then be implied by the `Enum` itself?

I use a `Enum` because the current API has a DRY issue where the panel names need to be duplicated and not accidentally misspelled. Though I suppose you could use `str` constants as well to avoid this.

iameskild:
@mike0sv (and others) looking to achieve something similar, this might help for the time being:

```python
from click import Context
import typer
from typer.core import TyperGroup

class OrderCommands(TyperGroup):
  def list_commands(self, ctx: Context):
    """Return list of commands in the order appear."""
    return list(self.commands)    # or modify self.commands as you please

app = typer.Typer(
    cls=OrderCommands,
)


...
```

johnthagen:
@iameskild Thanks for the suggestion.

[...]

PR #489, opened 2022-11-05
Help panel sorted by name, allow to not expand help command tables

asaintsever:
This PR brings following changes and features:

- Rich Help Panels for commands are now sorted by panel name (except for 'Commands' panel always printed first). Currently, command help panels are sorted by command name (ordering being done across all panels) which may results in weird/unexpected display.
- New `STYLE_COMMANDS_TABLE_EXPAND` style to allow to not expand tables of command help panels. The currently forced table expand  may not be desirable when you have several command panels (help text may end up far on the right side and alignment with other panel text too inconsistent). Disabling tables' expand option helps improving the result (in combination with `STYLE_COMMANDS_TABLE_PADDING` style).


ssbarnea:
Is this related to https://github.com/Textualize/rich/discussions/2878 ?

asaintsever:
@ssbarnea yep after looking at your discussion and PR, clearly related. Seems you also stumbled upon same issue with labels ending up not aligned across panels. Your proposal is different than mine though since I only make visible the expand property to fix it.

ssbarnea:
Now we need to find a way to persuade @tiangolo to pick one of the fixes, so we can make use of it.

svlandeg (commented):
Thanks for the contribution!

I had a look at this PR and compared it with `master` and #567. (don't mind the misaligned right-hand side, it's a c/p issue, it's actually nicely aligned on my console)

On `master`, I'd get:
```
┌─ Commands ──────────────────────────────────────────────────────────────────┐
│ create-a-wonderful-new-user                Create a new user. ✨            │
│ delete                                     Delete a user. 🔥                │
└─────────────────────────────────────────────────────────────────────────────┘
┌─ Utils and Configs ─────────────────────────────────────────────────────────┐
│ config    Configure the system. 🔧                                          │
│ sync      Synchronize the system or something fancy like that. ♻            │
└─────────────────────────────────────────────────────────────────────────────┘
┌─ Help and Others ───────────────────────────────────────────────────────────┐
│ help            Get help with the system. ❓                                │
│ report          Report an issue. 🐛                                         │
└─────────────────────────────────────────────────────────────────────────────┘

```

On this PR, with `STYLE_COMMANDS_TABLE_EXPAND = True` I'd get:

```
┌─ Commands ──────────────────────────────────────────────────────────────────┐
│ create-a-wonderful-new-user  Create a new user. ✨                          │
│ delete                       Delete a user. 🔥                              │
└─────────────────────────────────────────────────────────────────────────────┘
┌─ Help and Others ───────────────────────────────────────────────────────────┐
│ help    Get help with the system. ❓                                        │
│ report  Report an issue. 🐛                                                 │
└─────────────────────────────────────────────────────────────────────────────┘
┌─ Utils and Configs ─────────────────────────────────────────────────────────┐
│ config  Configure the system. 🔧                                            │
│ sync    Synchronize the system or something fancy like that. ♻              │
└─────────────────────────────────────────────────────────────────────────────┘
```

This is less white space pushing the description to the right, but still the commands aren't outlined across panels.

With PR #567, I'd get:
```
┌─ Commands ──────────────────────────────────────────────────────────────────┐
│ create-a-wonderful-new-user   Create a new user. ✨                         │
│ delete                        Delete a user. 🔥                             │
└─────────────────────────────────────────────────────────────────────────────┘
┌─ Utils and Configs ─────────────────────────────────────────────────────────┐
│ config                        Configure the system. 🔧                      │
│ sync                          Synchronize the system or something fancy     │
│                               like that. ♻                                  │
└─────────────────────────────────────────────────────────────────────────────┘
┌─ Help and Others ───────────────────────────────────────────────────────────┐
│ help                          Get help with the system. ❓                  │
│ report                        Report an issue. 🐛                           │
└─────────────────────────────────────────────────────────────────────────────┘
```

Which is nicely outlined across panels but does include a bit more whitespace, pushing e.g. the `sync` description to 2 lines instead of 1.

All in all, I think the output of this PR is better than what we'd have on `master`, but the result of PR #567 seems even more preferable. This is mostly a personal preference though.

In terms of functionality, this PR has two additional features to be considered:
- The sorting by name can still be done in a separate PR if we want it
- Making this alignment optional via a global setting can be considered for PR #567 as well.

I'll leave the final decision on this with Tiangolo!

typer/rich_utils.py
```diff
@@ -59,6 +59,7 @@
 STYLE_COMMANDS_TABLE_LEADING = 0
 STYLE_COMMANDS_TABLE_PAD_EDGE = False
 STYLE_COMMANDS_TABLE_PADDING = (0, 1)
+STYLE_COMMANDS_TABLE_EXPAND = True
```
svlandeg, line 62: Having this as a separate option that by default behaves as before, is in itself a clean solution!

tiangolo:
Thanks @asaintsever! :rocket: 

I went for https://github.com/tiangolo/typer/pull/567 to handle the length of the commands.

I want to allow making the order of panels and commands explicit, but not sorted by order. I also want to change that for commands, so that developers can decide which commands are shown first in the help menus. I think commands and their help are in many cases related, or some should be used first, before others, etc. And that doesn't depend on the name but on the intention of the developer, so, I want to order commands in the same order they were added to the app instead of the alphabetically. :nerd_face: 

Given that, I'll pass on this one, but thanks for the effort! :cake:

johnthagen:
@tiangolo Just for clarification:

> I want to allow making the order of panels and commands explicit, but not sorted by order

This isn't currently possible, correct? Just wanting to make sure I hadn't missed this being implemented

- #428

Thanks ❤️

PR #495, opened 2022-11-10
✏️ Fix typo in datetime docs

huxuan:


tiangolo:
Good catch, thanks! 🚀

PR #501, opened 2022-11-15
🐛 Update type annotations for `show_default` parameter and update docs for setting a "Custom default string"

### LINKED ISSUE #158: [BUG] Type hint for show_default of typer.Option is incorrect

plannigan:
Fixes #158.


svlandeg (commented):
Hi, thanks for the PR!

While we're pretty late to follow up on this, I believe this contribution still makes sense, as the feature is available for `typer.Option` - it's just not documented or correctly typed.

I hope you don't mind that I pushed to your branch directly to update it with the latest changes from `master`. I'll also include a new section for the documentation to showcase how this would work in a similar fashion as it does for `typer.Argument`.

plannigan:
Not a problem for the assistance in getting this merged.

tiangolo:
Awesome, thanks @plannigan! :nerd_face:

And thanks a lot @svlandeg for the review and tweaks! :bow: 

This will be available in Typer 0.9.2 in the next few hours. :rocket:

PR #502, opened 2022-11-15
some pylint tidyups

marksmayo:
simplification of some conditional statements
fixing of import orders
cleaning up of unneeded chars

### DIFF
--- a/docs_src/using_click/tutorial001.py
+++ b/docs_src/using_click/tutorial001.py
@@ -7,7 +7,7 @@ import click
 def hello(count, name):
     """Simple program that greets NAME for a total of COUNT times."""
     for x in range(count):
-        click.echo("Hello %s!" % name)
+        click.echo(f"Hello {name}!")
 
 
 if __name__ == "__main__":
--- a/.github/actions/comment-docs-preview-in-pr/app/main.py
+++ b/.github/actions/comment-docs-preview-in-pr/app/main.py
@@ -62,7 +62,7 @@ if __name__ == "__main__":
             "body": f"📝 Docs preview for commit {use_pr.head.sha} at: {settings.input_deploy_url}"
         },
     )
-    if not (200 <= response.status_code <= 300):
+    if not 200 <= response.status_code <= 300:
         logging.error(f"Error posting comment: {response.text}")
         sys.exit(1)
     logging.info("Finished")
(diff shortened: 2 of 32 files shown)


docs_src/using_click/tutorial001.py
```diff
@@ -7,7 +7,7 @@
 def hello(count, name):
     """Simple program that greets NAME for a total of COUNT times."""
     for x in range(count):
-        click.echo("Hello %s!" % name)
+        click.echo(f"Hello {name}!")
```
svlandeg, line 10: While F-strings are typically nicer and more readable, there are a few cases where you might prefer not using them. When we're logging things for instance, you might prefer avoiding the `str()` functionality for cases where the statement wouldn't actually get logged (e.g. if the logging level is higher). That said - I'm not actually sure how `click.echo()` operates under the hood so I'm not sure whether we want this rewrite in this case or not.

svlandeg (commented):
Thanks for the PR! Definitely appreciate your contribution.

I left a few spurious comments. I think it will be up to `tiangolo` to decide whether we want some of these refactorings, or whether we'd keep to the original repo source style.

docs_src/using_click/tutorial001.py
tiangolo: Ah! This one is just that it was an exact copy of Click's tutorial. But they now updated it to use f-strings, so I would take this change. :nerd_face:

tiangolo:
Thanks @marksmayo! :nerd_face: 

And thanks @svlandeg for the thorough review! :bow: 

For most of the code style changes I would just use what is configured in pre-commit (I'm not a fan of PyLInt, actually).

Soon, after removing support for Python 3.6, I will migrate all to Ruff.

This PR covers too many different things, so it would be better to split it to be able to take some things while not taking others.

Specifically, I would take the changes in the Click examples from `"some str {blah}".format(blah=blah)` to f-strings. That was just an exact copy of Click's tutorials, but now they updated them, we can update them here as well.

For now, I'll close this one, but feel free to add a new PR changing those Click f-strings! :rocket:

docs_src/using_click/tutorial001.py
svlandeg: cf https://github.com/tiangolo/typer/pull/891

PR #509, opened 2022-11-21
Add rich print using _get_rich_console

feluelle:


feluelle:
@tiangolo I would appreciate a review on this one.

cc @willmcgugan as the author of this module (and rich ❤️)

We might want to change how we load the configuration such as `TERMINAL_WIDTH` here. WDYT?

tiangolo:
Thanks! For now I prefer to pass on this one, but thanks for the interest! :coffee:

Issue #523, opened 2023-01-03
New setting to increase the width of the rich (exception) output


brot:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer

app = typer.Typer()


@app.command()
def main(name: str = "morty"):
    deep_dict_or_json = {
        "this_is_a_long_key": {
            "this_is_the_next_long_key": {
                "this_is_the_next_long_key": {
                    "this_is_the_next_long_key": {
                        "this_is_the_next_long_key": {
                            "this_is_the_next_long_key": {
                                "this_is_the_next_long_key": {
                                    "this_is_the_next_long_key": {
                                        "this_is_the_next_long_key": {
                                            "this_is_the_next_long_key": {
                                                "and_once_again_a_very_long_key": {
                                                    "but_this_is_not_the_end": {"end": True}
                                                }
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
    print(name + 3)


if __name__ == "__main__":
    app()
```

### Description

If I want to include the locals in the nice rich output and such a local includes a very big and deep dictionary (e.g: Kubernetes pod description) the output is very long, although I have a lot of horizontal space in my terminal.

### Wanted Solution

So I want to be able to increase the width of the yellow and red rich exception border to make the output more useful and readable

### Wanted Code

```python
app = typer.Typer(pretty_exceptions_with=120)
```

### Operating System

macOS

### Typer Version

0.7.0

### Python Version

3.11.1

indiVar0508:
hi, would it be okay if I made a PR for it?

YuriiMotov:
Currently it fits this block in terminal width by default:

```
python main.py
╭──────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────────────╮
│ /home/yurii/code/typer/motov/helloworld/main.py:33 in main                                                                                                                                                 │
│                                                                                                                                                                                                            │
[...]

PR #528, opened 2023-01-07
Add #523

indiVar0508:
add parameter as suggested feature in issue #523


indiVar0508:
Hi @tiangolo , 
Did you get a chance to review this PR

Napo2k:
I definitely would like for this to be merged in! @tiangolo

nikita-b:
Yes, feature looks really good. @tiangolo Could you please, review?

docs/tutorial/exceptions.md
```diff
@@ -202,6 +202,17 @@ $ python main.py
 
 </div>
 
+
+## Set Output Width
+
+If you want to control the width of the yellow and red Rich exception borders, you can set the parameter `pretty_exceptions_width` to a specific integer (it's 100 by default):
+
+```Python hl_lines="3"
+{!../docs_src/exceptions/tutorial005.py!}
```
svlandeg, line 211: Alternatively, should we rename tutorial004 to 005 and call the new one from this PR 004?

docs/tutorial/exceptions.md
```diff
@@ -202,6 +202,17 @@ $ python main.py
...
+```
+
+This prevents artificial line breaks in cases where there is sufficient horizontal space on the console.
+
```
svlandeg, line 215: Tiangolo: I'm not sure whether we want to show output here, it gets a bit unwieldy.

tests/test_tutorial/test_exceptions/test_tutorial005.py
```diff
@@ -0,0 +1,34 @@
...
+from pathlib import Path
+
+from typer.testing import CliRunner
+
+from docs_src.exceptions import tutorial005 as mod
+
+runner = CliRunner()
+
+
+def test_traceback_rich_pretty_width():
```
svlandeg, line 13: This test simply checks that the program runs like expected, but it doesn't (yet) actually test the Rich borders length as I wasn't quite sure how to do that 🤔

svlandeg (commented):
Thanks for this contribution @indiVar0508!

I tested this feature with the code from https://github.com/tiangolo/typer/issues/523 and I added this example to the docs as `docs_src/exceptions/tutorial005.py`. When running this on my system on `master`, the output gets split up by artificial newlines: 

image

However, setting `pretty_exceptions_width=120` with this PR shows it more nicely:

image

So I think having this additional parameter makes sense. I'll leave the decision with Tiangolo though. I also added a unit test and a section in the docs, but I'm not 100% sure this is how we want it so I left some review comments where I was unsure.

svlandeg:
Merging in the latest from `master` has made the CI on this PR fail - I'm looking into it.

thejevans:
If this is going to be added, it would be great if all options available to [`rich.traceback.install`](https://rich.readthedocs.io/en/stable/reference/traceback.html#rich.traceback.install) could be used, maybe as a dict parameter instead of a parameter for each one.

svlandeg:
There's a recently merged PR that has conflicting behaviour for the setting of `width` in `Traceback.from_exception` - I'll look into it.

svlandeg:
Ok, so there has been a related [PR](https://github.com/fastapi/typer/pull/788) that was recently merged and that unifies the width of the Rich console for help and errors.

On the new `master`, this is then the output of the scripts as mentioned above:

image

i.e. the error message takes up the full width of the console, which is already better than the original situation described here.

Nonetheless, I still think it can be interesting to have this as a user-adjustable setting. So, I've merged the functionality from both PRs by setting the default of `pretty_exceptions_width` to `rich_utils.MAX_WIDTH` if available, and `100` otherwise.

typer/main.py
```diff
@@ -81,14 +81,15 @@ def except_hook(
...
 
         rich_tb = Traceback.from_exception(
             type(exc),
             exc,
             exc.__traceback__,
             show_locals=exception_config.pretty_exceptions_show_locals,
             suppress=supress_internal_dir_names,
-            width=rich_utils.MAX_WIDTH,
+            width=exception_config.pretty_exceptions_width or MAX_WIDTH,
         )
```
svlandeg: Due to the recently merged https://github.com/fastapi/typer/pull/1128, I've had to change the code slightly to not import and use `MAX_WIDTH` earlier on, but instead use `None` as default on L156 and then set `MAX_WIDTH` here whenever `pretty_exceptions_width` is `None`.

YuriiMotov:
Is the intention of this PR to handle situations when it's impossible to determine the width of the terminal screen?

Currently you can specify the width of the terminal screen using `TERMINAL_WIDTH` ENV variable.

https://github.com/fastapi/typer/blob/0bc7cced82ff95681ecc095d0ef461840a24862c/typer/rich_utils.py#L74-L75

<details><summary>See in the details</summary>

```
TERMINAL_WIDTH=70 python main.py
╭──────────────── Traceback (most recent call last) ─────────────────╮
│ /home/yurii/code/typer/motov/helloworld/main.py:33 in main         │
│                                                                    │
│   30 │   │   │   }                                                 │
│   31 │   │   }                                                     │
│   32 │   }                                                         │
│ ❱ 33 │   print(name + 3)                                           │
│   34                                                               │
│   35                                                               │
│   36 if __name__ == "__main__":                                    │
│                                                                    │
│ ╭──────────────────────────── locals ────────────────────────────╮ │
│ │ deep_dict_or_json = {                                          │ │
│ │                     │   'this_is_a_long_key': {                │ │
│ │                     │   │   'this_is_the_next_long_key': {     │ │
│ │                     │   │   │   'this_is_the_next_long_key': { │ │
│ │                     │   │   │   │                              │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │                          │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │   │                      │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │   │   │                  │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │   │   │   │              │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │   │   │   │   │          │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │   │   │   │   │   │      │ │
│ │                     'this_is_the_next_long_key': {             │ │
│ │                     │   │   │   │   │   │   │   │   │   │   │  │ │
│ │                     'and_once_again_a_very_long_key': {        │ │
│ │                     │   │   │   │   │   │   │   │   │   │   │  │ │
│ │                     │   'but_this_is_not_the_end': {           │ │
│ │                     │   │   │   │   │   │   │   │   │   │   │  │ │
│ │                     │   │   'end': True                        │ │
│ │                     │   │   │   │   │   │   │   │   │   │   │  │ │
│ │                     │   }                                      │ │
│ │                     │   │   │   │   │   │   │   │   │   │   │  │ │
│ │                     }                                          │ │
│ │                     │   │   │   │   │   │   │   │   │   │   }  │ │
│ │                     │   │   │   │   │   │   │   │   │   }      │ │
│ │                     │   │   │   │   │   │   │   │   }          │ │
│ │                     │   │   │   │   │   │   │   }              │ │
│ │                     │   │   │   │   │   │   }                  │ │
│ │                     │   │   │   │   │   }                      │ │
│ │                     │   │   │   │   }                          │ │
│ │                     │   │   │   }                              │ │
│ │                     │   │   }                                  │ │
│ │                     │   }                                      │ │
│ │                     }                                          │ │
│ │              name = 'morty'                                    │ │
│ ╰────────────────────────────────────────────────────────────────╯ │
╰────────────────────────────────────────────────────────────────────╯
TypeError: can only concatenate str (not "int") to str
```

</details>

I think we don't need to change anything. Probably just document this ENV variable

tiangolo:
Thanks for the interest! I think this wouldn't be necessary right now, I would suspect it would be a misconfiguration in the terminal displaying the content.

If you still have problems, please create a new Discussion question with all the information of your environment so that we can try and replicate it to understand the problem better.

For now, I'll pass on this one. Thanks! :coffee:

PR #545, opened 2023-02-01
light/dark mode switch added

Alirezaaraby:
toggle dark/light mode added to docs

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -10,15 +10,15 @@ theme:
       primary: black
       accent: teal
       toggle:
-        icon: material/brightness-7
+        icon: material/lightbulb-outline
         name: Switch to dark mode
 
     - media: "(prefers-color-scheme: dark)"
       scheme: slate
       primary: black
       toggle:
-        icon: material/brightness-4
-        name: Switch to system light mode
+        icon: material/lightbulb
+        name: Switch to light mode
   icon:
     repo: fontawesome/brands/github-alt
   logo: img/icon-white.svg
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -10,14 +10,14 @@ theme:
       primary: black
       accent: teal
       toggle:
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -6,8 +6,19 @@ theme:
   name: material
   custom_dir: docs/overrides
   palette:
-    primary: black
-    accent: teal
+    - media: "(prefers-color-scheme: light)"
+      primary: black
+      accent: teal
+      toggle:
+        icon: material/brightness-7
```
atmartinez, line 13: I would consider matching these icons to the FastApi [documentation](https://github.com/tiangolo/fastapi/blob/6879082b3668edd213d035b6e9a90a4bccf32e01/docs/en/mkdocs.yml#L13) to maintain consistency between the projects.

Alirezaaraby: good idea

atmartinez (approved):
This looks good to me for this specific doc style change. I would consider swapping the accent colors to amber to match FastApi docs, but the scope of this PR per the title is to just add the toggle switch.

Alirezaaraby:
> This looks good to me for this specific doc style change. I would consider swapping the accent colors to amber to match FastApi docs, but the scope of this PR per the title is to just add the toggle switch.

Firstly, typer doc has it's own style so i think there's no need to change the color to amber. And secondly, I wrote light/dark **mode** switch, and the mode is a key word here which shows this pr will add dark mode switch to typer docs :)

svlandeg:
[maintenance note:] Related PR: https://github.com/tiangolo/typer/pull/418

svlandeg (commented):
Hi @Alirezaaraby, thanks so much for the contribution!

It looks like in the meantime, dark mode for the docs has been introduced by Tiangolo in PR #908 🎉 As such, I think we can close this one, but thanks again! 🙏

PR #524, opened 2023-01-03
⬆ Upgrade Rich, support 13.x

musicinmybrain:
Based on the changelog, the only relevant breaking change is that support for Python 3.6 is removed:

https://github.com/Textualize/rich/blob/v13.0.0/CHANGELOG.md#1300---2022-12-30

This is OK even though we support Python 3.6 in typer, because we do still allow older versions of rich that are compatible with Python 3.6.

PR #551, opened 2023-02-07
Bump Rich constraint

mhindery:
There are more recent releases of the Rich dependancy. This caused problems trying to use the package in combination with other libraries that have a dependancy on Rich being more recent. I've updated the constraint from version `<13` to `<14`. Currently the latest Rich version is `13.3.1`.


mzebrak:
It seems that an identical PR already exists: #524

tiangolo:
Thanks @mhindery! :cake: 

And thanks for the note @mzebrak. :coffee: 

Yep, this was handled in https://github.com/tiangolo/typer/pull/524. Given that, I'll close this one. Thanks again for the effort! :nerd_face:

PR #548, opened 2023-02-04
🐛 Fix support for `UnionType` (e.g. `str | None`) with Python 3.11

jonaslb:
Potential solution for #533, which already had good details and a proposal for a solution, based on using `get_args` and `get_origin` from the `typing` module instead of the `__args__` and `__origin__` attributes. So I went and did that.

We also needed to actually handle the `UnionType` in addition to the `Union` from before. 

Now there was a bit of a headache in terms of writing this in a backwards compatible way. What I've done is place the compatibility code into `_compat_utils.py`. But let me know if you'd prefer to keep it in `main.py`.

I also added a single test, just an adapted copy of another one with the `x | None` syntax. It is conditioned on Python >= 3.10, so it shouldn't fail on older Pythons.

PR #556, opened 2023-02-17
Bump mypy to 1.0.1 & remove Python 3.6 support

jonaslb:
This bumps mypy and fixes necessary minor things that turned up as a conseqence of it.

Related to #548 where I couldn't really find a solution that worked with the old mypy, but when I fixed it to work with a newer mypy I didn't realize it was pinned (which makes sense). But anyway this mypy update should make the other PR pass tests...


jonaslb:
Ah, looks like mypy has dropped Python 3.6, so it won't work with the tests there.... Maybe I will look into making typer drop 3.6 then.

jonaslb:
That's it, although now there are two topics in the PR, they are in separate commits and overall the changes are still minor.

kudlatyamroth:
> Ah, looks like mypy has dropped Python 3.6, so it won't work with the tests there.... Maybe I will look into making typer drop 3.6 then.

Click drop support for 3.6 in 8.1 version ( https://click.palletsprojects.com/en/8.1.x/changes/#version-8-1-0 )
so Typer also not support it :)

svlandeg:
>  Maybe I will look into making typer drop 3.6 then.

Hi! Note that Python 3.6 support was recently dropped in 0.11.0.

jonaslb:
Yeah looks like this has been long superseded by other changes, so closing.

PR #565, opened 2023-03-12
Deprecate Python 3.6

gliptak:
https://devguide.python.org/versions/


gliptak:
@tiangolo please review

PR #567, opened 2023-03-13
Ensure commands column is the same on all panels

ssbarnea:
Previously the width of table columns for commands in each panel used
auto-width, making hard to read the commands.


Skyler-Altol:
This is awesome, @tiangolo can we get this merged please? A bunch of people at my work have not liked the fact that the panels didn't align, and that help text for commands was sometimes wayyyy off to the far right instead of being close to the command name as you'd expect.

Tremeschin:
Adding my two cents in;

I've been waiting ages for this to be accepted 😭, I discovered this PR near the time it was created after my research advisor pointed the "misalignment" on a meeting last year showcasing an internal management tool; Even thought I've some level of OCD myself, I've never really noticed this, but now I can't unsee it 😂

There's a workaround to install the PR as a git dependency on pip/poetry with..

- `pip install git+https://github.com/tiangolo/typer.git@refs/pull/567/head`

..but, magically, some PCs on the lab didn't have git, and future-proofing documentation for non tech-savy students on multiple OSs wasn't ideal, and stuff could easily break on their hands I reckon.

The CLIs are much more elegant and professional with the alignment, please @tiangolo, consider merging this PR ❤️

svlandeg:
[maintenance note] related PR: https://github.com/tiangolo/typer/pull/489

NikosAlexandris:
> * pip install git+https://github.com/tiangolo/typer.git@refs/pull/567/head

For some reason this downgrades to typer 0.7.0 here.

svlandeg:
> No idea why pip downgrades typer when install this PR.

That's probably because this PR was created branching off an older version of `master`. I'm bringing it up-to-date now.

typer/rich_utils.py
```diff
@@ -628,13 +639,20 @@ def rich_format_help(
                 )
                 panel_to_commands[panel_name].append(command)
 
+        # Identify the longest command name in all panels
+        max_cmd_len = 0
+        for _, commands in panel_to_commands.items():
+            for command in commands:
+                max_cmd_len = max(max_cmd_len, len(command.name or ""))
```
svlandeg, line 646: Depending on Tiangolo's preference, we could also rewrite this with a list comprehension, which could be slightly more efficient in doing the `max` calculation:
```suggestion
        max_cmd_len = max([len(c.name) for _, commands in panel_to_commands.items() for c in commands if c.name], default=0)
```

typer/rich_utils.py
```diff
@@ -487,7 +488,17 @@ def _print_commands_panel(
...
+        style="bold cyan",
+        no_wrap=True,
+        width=cmd_len,
+    )
+
+    # A big ratio makes the description column be greedy and take all the space
+    # available instead of allowing the command column to grow and misalign with
+    # other panels.
+    commands_table.add_column("Description", justify="left", no_wrap=False, ratio=10)
+    commands_table.add_column("xx", style="bold yellow", justify="left", no_wrap=False)
```
svlandeg, line 501: Why is this `"xx"` column needed? On my end, things seem to still work correctly without it.

svlandeg (approved):
I tested this with a slightly altered version of `docs_src/commands/help/tutorial006.py` (making the command names differ more in length) and I do think this change improves readability overall of the console output, as also demonstrated by the screenshots in the PR description.

typer/rich_utils.py
tiangolo: Good idea! I just added that. :rocket:

typer/rich_utils.py
tiangolo: Agreed, it seems this last line is not necessary, I just removed it.

tiangolo:
Awesome, thank you @ssbarnea! :rocket: 

Thanks for the discussion here everyone. :coffee: 

Thanks for the help and review as always @svlandeg! :bow: 

This will be available in the next few hours in Typer 0.12.2 :tada:

PR #571, opened 2023-03-21
🐛 Fix preserving case in enum values

avaldebe:
fixes #570: Enum values that differ in case get conflated


mcanueste (approved):

avaldebe:
> Is there any reason why this has yet to be merged? Can I help?

Is there something I should do?
I updated my branch once, should I do it again?

svlandeg (commented):
Hi, thanks for the fix!

I could replicate the bug on `master` and verify that this PR fixes the issue. I looked into the code edits in detail and those look fine to me as well.

I do think that the original example from #570 was slightly more clear, so I'll go ahead and update the test case you've included if that's OK.

svlandeg (approved):
This looks good to me, though I'd love a quick verification by @tiangolo on the following points:

- The original code called `.lower()` clearly on purpose - are there any use-cases we're not supporting anymore after this rewrite? I don't personally think so as users can set the `case_sensitive` flag on `typer.Option` (as documented). Still, should we consider this PR as potentially breaking user's previous code?
- <s>Which version of the test would you like to keep? There's `tests/test_enum_case.py` but I also included a version with `docs_src/parameter_types/enum/tutorial003.py`. If we want the latter, we can go ahead and delete `tests/test_enum_case.py`.</s>

svlandeg:
After internal discussion with @tiangolo, I updated the PR further to use the same example from the docs, and to add a note to the documentation. The current documentation has a specific section on how to make the enums case insensitive, which seems to confirm that the current behaviour on `master` is indeed a bug, and that the enum should be case sensitive by default as in this PR.

tiangolo:
Awesome, thanks @avaldebe! :rocket: 

And thanks @krzysztofwos for the report. :nerd_face: 

Also, thanks @svlandeg for the help! :bow: 

This will be available in Typer 0.9.1 in the next few hours. :rocket:

PR #581, opened 2023-04-21
pyproject.toml: use modern PEP 621 flit config

gotmax23:


gotmax23:
@tiangolo, I've rebased this. Do you mind taking a look?

tiangolo:
Thanks for the interest! The build system was replaced with PDM, including updating the metadata. Given that, I'll close this one. Thanks for the effort! :cake:

PR #443, opened 2022-08-15
Support custom parameter types

paulo-raca:
While most CLI parameters are strings, ints, etc, we sometimes need custom
types.

Currently typer has no support for it (See #77), and while there are a few hacks, the
likely solution is to add a 'str' argument and parse it inside the main function.

This PR adds support for custom types in 3 different ways:
- Manually specifying a `click_type`
- Manually specifying a `parse` function
- Using a Callable type annotation (It is very common for types to have a string constructor, like `int("1")`)

PR #583, opened 2023-04-25
✨ Add support for custom types and parsers (updates/builds upon PR #443)

### LINKED ISSUE #77: [FEATURE] - Extend with custom data types
I want to extend the CLI app with custom data types not supported by Typer/Click. Right now, the Typer app raises an error on encountering an unsupported type. 

We could add a method to the Typer object to register custom data types along with a function to convert terminal input string to an object of the desired type. 

A possible API could be:

```py
app = typer.Typer()
app.register_custom_type(_type, _deserializer)
```

Here is the monkey-patched solution I use right now without providing a `registration` API. If this seems like a useful feature, I'll try refine this and submit a PR. 

```py3
import datetime
from typing import Any, Callable, Optional

import typer
import click


_get_click_type = typer.main.get_click_type
REGISTERED_TYPES = {
    datetime.date: lambda string: datetime.datetime.strptime(string, '%Y-%m-%d').date()
}

class TyperCustomParam(click.ParamType):
    name = 'CustomParameter'
    def __init__(self, typ: type, deserializer: Optional[Callable] = None): 
        self.name = typ.__name__
        self._deserializer = typ if deserializer is None else deserializer

        try:
            return self._deserializer(value)
        except Exception as E: 
            self.fail(
                f"couldn't serialize {value} to an instance of type {self.name}, error: {E}"
            )

def supersede_get_click_type(
    *, annotation: Any, parameter_info: typer.main.ParameterInfo
) -> click.ParamType:
    if annotation in REGISTERED_TYPES:
        return TyperCustomParam(annotation, REGISTERED_TYPES[annotation])
    else:
        return _get_click_type(annotation=annotation, parameter_info=parameter_info)

typer.main.get_click_type = supersede_get_click_type
```

### Alternatives
Alternatives would be to either modify the original routine or create a wrapper function to accept a Click supported datatype and convert later. Letting users create custom types directly is preferred since that lets us create CLIs without touching existing code.

Thank you for looking into this. Similar to FastAPI, typer is a really nice library.

jpurviance:
Built upon #443 and adds documentation. Related to #77

Add support for custom types:
- a `parse` function can be provided to create the type.
- a `click.ParamType` parse can be provided to create the type.


tiangolo:
Thank you @jpurviance and @paulo-raca! :rocket: :cake: :man_dancing: 

I tweaked the docs and tests a bit. :nerd_face: 

I'll release this in the next few hours in Typer `0.8.0`. :tada:

PR #586, opened 2023-04-27
✏ Fix spelling (shinny -> shiny)

runofthemill:
Just a small spelling fix - cheers! :)


tiangolo:
Ah! Good catch! Thanks! :mag: :cake:

PR #587, opened 2023-05-01
⬆️ Upgrade pip in CI, try and see if that handles the conflicts

tiangolo:


tiangolo:
It seems that doesn't do it. Probably needs upgrading Flit, but I can only do that when dropping support for 3.6, so I'll have to do that first.

PR #647, opened 2023-08-02
✨ Add `typer.set_rich_*()` to allow Rich formatting to be disabled

JacobKochems:
Implements a feature similar to the one suggested by @anthonyraf in issue https://github.com/tiangolo/typer/discussions/622 to optionally disable Rich output.

Two simple test cases and a section of documentation has also been added.

It affects all instances of `Typer` globally.
The code from the test cases illustrates how to use it:
```
import typer

typer.set_rich_help(False)

app = typer.Typer()


@app.command()
def main(arg: str):  # pragma: no cover
    pass


if __name__ == "__main__":
    app()
```
```
import typer

typer.set_rich_traceback(False)

app = typer.Typer()


@app.command()
def raise_():
    raise ValueError  # raise some error to test traceback output


if __name__ == "__main__":
    app()
```

Resolves #578, resolves #622
Bonus: It is also a workaround for #646.


suvayu (changes requested):
> typer.enable_rich(False)

I guess you mean `typer.enable_rich_help(False)` since that's what is in the code.

I ran the tests locally, and checked against my own example script.  Everything works as advertised.  The implementation is also straightforward.

That said, I think the functions should be called `disable_rich_*` since enable is already the default behaviour, so in the `enable_rich_*` form it will always be a double negative.  Which is a bit weird.

JacobKochems:
Thanks for the review. I'll rename it as you suggested. I was thinking the same way about my previous naming choice anyway shortly after I published the PR but wanted to wait for a response before touching the code again. I don't know why I even named it that way to begin with. ;-)

nickmuoh:
Is there any plan on having this merged and released? It's a really valuable contribution as not scenarios where rich and typer are installed mean users want to use rich with typer.

JacobKochems:
I renamed the two functions and the thereby affected variables to be more semantically precise and more specific to their purpose. For the exposed functions I settled on the form: `set_rich_*` which is semantically more appropriate than the previous `enable_rich_*` form but keeps the boolean parameter to be able to conditionally re-enable it.

docs/tutorial/printing.md
```diff
@@ -140,6 +140,16 @@ In general, **Typer** tends to be the entry point to your program, taking the fi
 
 The best results for your command line application would be achieved combining both **Typer** and **Rich**.
 
+### If you don't want to use Rich formated output
+
+In case you'd rather like to use the simple text ouput, but need **Rich** to be installed in your environment for any other reason, you can disable the use of **Rich** by **Typer**.
+There are two functions in the **Typer** module to control the output format globally, affecting all `typer.Typer()` instances:
```
suvayu, line 146:
```suggestion
### Configuring Rich formatted output

By default, **Typer** uses the **Rich** library to format text output.  However it's an optional dependency, and is used only if it is installed in your environment.  In case you would rather have finer-grained control over when **Typer** uses rich formatting, you can do so with the two package level functions shown below to control the output format globally.  This affects all `typer.Typer()` instances.
```
Since the new version of the global functions are setters instead of a toggle, the documentation should reflect that

suvayu (changes requested):
I ran the tests, and tried on my example script, everything works. I suggested wording change in the documentation to match the new function names.  Feel free to adapt that.  I think it could be improved a bit by merging with the previous heading "Typer and Rich" and making it more coherent.  As it is now, it kind of sticks out.

docs/tutorial/printing.md
JacobKochems: I commited your suggestions.
> I think it could be improved a bit by merging with the previous heading "Typer and Rich" and making it more coherent. As it is now, it kind of sticks out.

The previous heading speaks about the roles of `Typer` and `Rich` respectively. There might be a way to neatly integrate it there and make it more coherent but on the other hand it might be more discoverable the way it is. I think that is a decision we can leave up to the maintainer.

suvayu: okay fair point, I'll approve this

suvayu (approved):
Okay, looks good now.  Maybe at some point @tiangolo would like to merge the docs with the "Typer and Rich" section to make it more coherent.

coreegor:
Hello everyone,
Is there any plan on having this merged ? 
@JacobKochems can you set an assignee ?

coreegor:
@JacobKochems So you can't set @tiangolo as the assignee ?

JacobKochems:
> @JacobKochems So you can't set @tiangolo as the assignee ?

No, I can't set **anyone** as assignee. Not myself and certainly not tiangolo. Nor would I want to. We are all guests here and it isn't my place to assign any task to anybody in this repository.

JacobKochems:
@coreegor Since you already @ed him I would assume that he already got an email notification but either didn't have time to respond yet or he time boxes his work on a given project at certain intervals.

If you need this feature rather sooner than later, I see two options:

1. Use my fork for the time being
2. Write a nice email directly to tiangolo

I'd prefer option 1. We all have to navigate a world where seemingly each and everything is competing for our attention. There is probably a reason why the last commit is from July 30.

JacobKochems:
@coreegor I just merged the latest changes from 'tiangolo:master'. Could you also do a review of this PR according to [this](https://typer.tiangolo.com/help-typer/#review-pull-requests)? You can find this functionality under `Files changed` on the right hand side, named: `Review changes`.

JacobKochems:
To all following this PR: The original motivation for this PR is the issue #646 for which this is a possible workaround. However, I'd rather would like to see that original issue addressed as well. If some of you could review/replicate it, it would be much appreciated.

anthonyraf (approved):

coreegor (approved):

melsabagh:
@JacobKochems I can replicate #646 with the latest Rich (13.7.0). To disable Rich though, I am doing the following: 

```python
import typer.core

typer.core.rich = None

...
```

anthonyraf:
> @JacobKochems I can replicate #646 with the latest Rich (13.7.0). To disable Rich though, I am doing the following:
> 
> ```python
> import typer.core
> 
> typer.core.rich = None
> 
> ...
> ```

This is working well, thanks

JacobKochems:
> @JacobKochems I can replicate #646 with the latest Rich (13.7.0). To disable Rich though, I am doing the following:
> 
> ```python
> import typer.core
> 
> typer.core.rich = None
> 
> ...
> ```

Thank you for this very simple workaround. Recently, yet another effort related to issues #578 and #622 has been made: #726. This in combination with the kwarg [pretty_exceptions_enable=False](https://stackoverflow.com/a/76375308) of `typer.Typer()` seems to accomplish the same with fewer changes.
Tough, I'm not sure about the intended semantics regarding `rich_markup_mode` and if the proposed solution in #726 (or mine for that matter, I can't test it at the moment) addresses the printing of the rounded panel boxes and rich markup in general independently from each other. This seems to be a desired [feature](https://github.com/tiangolo/typer/discussions/578#discussioncomment-6139966).

strangemonad:
I'd also like to add a vote for and option that allows globally disabling rich. Just because rich is installed and available doesn't always mean it's appropriate to render rich output from the cli. I'd like to not have to revert to disabling features piecemeal e.g. `prettr_exceptions_enabled=False + ...`

svlandeg:
Hi @JacobKochems,

The `format-imports.sh` script has indeed been deleted in PR #797. `format.sh` works as before, but `ruff` is a `test` dependency of `typer` and should be installed.

I noticed an issue with the `test-cov-html.sh` earlier this week as well, but I can't reproduce it anymore and now it just works fine on my system when I run
```
bash scripts/test-cov-html.sh
```

If this issue persists on your system, I suggest you open a discussion thread for that specifically so we can follow up there.

With respect to this PR - I want to review it in more detail, but I've looked at the related PR #726 first.

JacobKochems:
Ok, I did need to install `pytest-cov` and `pytest-xdist`. I didn't remember this to be necessary when I set up the repo the first time around but now `bash scripts/test-cov-html.sh` runs. It produces a lot of test fails but so does the master branch of tiangolo.

svlandeg:
@JacobKochems: I put up a related PR here: https://github.com/tiangolo/typer/pull/859 - I think the same mechanism can probably be reused to implement the global switch you want. If you (or others) have time to review #859, I'd appreciate it!

JacobKochems:
As mentioned in my post on https://github.com/tiangolo/typer/pull/859#issuecomment-2163754301 I will close this PR in favor of #859 and attempt to reimplement this PR on top it.

PR #648, opened 2023-08-05
Add keyword arguments to `typer.main.run()`

StarterCraft:
The proposed changes to the Typer library's run method allow for passing keyword arguments to its implicit Typer instance.

This enables more customization options for users when running their functions with the main method. The updated code includes all keyword arguments from Typer constructor.

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -1050,7 +1050,56 @@ def get_param_completion(
     return wrapper
 
 
-def run(function: Callable[..., Any]) -> None:
-    app = Typer(add_completion=False)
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@StarterCraft, thanks for your interest in project!

I personally think that this way it will be hard to maintain the consistency of this interface (`typer.run()`) with `typer.Typer()`.
Do you think it's possible to create a script that would check this consistency automatically in CI?

Anyway, let's wait for Sebastian to approve the idea.

PR #651, opened 2023-08-08
Unpin flit-core build dependency

tjni:
I'd like to be able to build this project in [nixpkgs](https://github.com/NixOS/nixpkgs) using the latest version of flit-core. Please let me know what you think. Thank you.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=2,<3"]
+requires = ["flit_core >=2"]
 build-backend = "flit_core.buildapi"
 
 [tool.flit.metadata]


g0hl1n:
Thanks! Works for me using YoctoProject/OpenEmbedded to build python3-typer.

rubienr:
Are there any updates if/when it will be merged?

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=2,<3"]
+requires = ["flit_core >=2"]
```
svlandeg, line 2:
```suggestion
requires = ["flit_core >=4"]
```
As libraries can have breaking changes on major versions, it's typically good practice to have an upper pin. 

That said - it looks like v3.x of `flit_core` works fine, so we could consider extending the range to `<4`.

tiangolo:
The build system was replaced with PDM, so this would currently not apply. Given that, I'll close this one, but thanks for the effort! :cake:

Issue #410, opened 2022-07-03
Empty tuple returned for Option instead of expected `None` when no values passed


Andrew-Sheridan:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer
from typing import Optional, List

app = typer.Typer(add_completion=False)


@app.command()
def main(
    foos: Optional[tuple[str]] = typer.Option(None, "--foo"),
    bar: Optional[str] = typer.Option(None, "--bar"),
):
    print(f"{type(foos)=}, {foos=}")
    print(f"{type(bar)=}, {bar=}")


if __name__ == "__main__":
    app()
```

### Description

If an option is declared as `foo: Optional[List[str]] = typer.Option(None)` it should provide `None` if nothing is passed. Currently it is providing `()`, an empty tuple.

In comparison `bar: Optional[str] = typer.Option(None)` does provide `None` when missing, as expected.

```console
❯ python issue.py --help
Usage: issue.py [OPTIONS]

Options:
  --foo TEXT
  --bar TEXT
  --help      Show this message and exit.
```

Nothing passed causes the returned value for `foo` to be an empty tuple, it should be `None`:
```console
❯ python issue.py
type(foos)=<class 'tuple'>, foos=()
type(bar)=<class 'NoneType'>, bar=None
```

Passing values it returns a tuple of the values
```console
❯ python issue.py --foo Hello --foo World --bar Goodbye
type(foos)=<class 'tuple'>, foos=('Hello', 'World')
type(bar)=<class 'str'>, bar='Goodbye'
```

### Operating System

macOS

### Typer Version

0.4.1

### Python Version

3.10.0

### Additional Context

https://github.com/tiangolo/typer/issues/127#issuecomment-721710767

jonasjancarik:
The same thing happens with lists (an empty list is returned instead of None) when you use `Optional[List[str]] = typer.Option(None)` (as suggested in https://typer.tiangolo.com/tutorial/multiple-values/multiple-options/)

svlandeg:
[...]

Issue #533, opened 2023-01-12
Using `some_type | None` syntax for type annotations causes error in python 3.11


PhilReinhold:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer
from datetime import datetime
print(typer.__version__)

app = typer.Typer()

@app.command()
def f(x: datetime | None = None):
    print(x)

if __name__ == "__main__":
    app()
```

### Description

In python 3.10 this script runs as expected, but on 3.11 I get the following error

```
Traceback (most recent call last):
  File "/Users/pcrein/qdash/test.py", line 16, in <module>
    app()
  File "/Users/pcrein/.pyenv/versions/qdash-3.11.0/lib/python3.11/site-packages/typer/main.py", line 328, in __call__
    raise e
  File "/Users/pcrein/.pyenv/versions/qdash-3.11.0/lib/python3.11/site-packages/typer/main.py", line 311, in __call__
    return get_command(self)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^
  File "/Users/pcrein/.pyenv/versions/qdash-3.11.0/lib/python3.11/site-packages/typer/main.py", line 364, in get_command
    click_command = get_command_from_info(
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pcrein/.pyenv/versions/qdash-3.11.0/lib/python3.11/site-packages/typer/main.py", line 577, in get_command_from_info
    ) = get_params_convertors_ctx_param_name_from_function(command_info.callback)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pcrein/.pyenv/versions/qdash-3.11.0/lib/python3.11/site-packages/typer/main.py", line 553, in get_params_convertors_ctx_param_name_from_function
    click_param, convertor = get_click_param(param)
                             ^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #348, opened 2022-01-04
PEP 604 Support


AngellusMortis:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from pathlib import Path
import typer


def main(output: Path | None):
    typer.guess_name(f"Hello {name}")


if __name__ == "__main__":
    typer.run(main)
```

### Description

Using an argument with the type of `pathlib.Path | None` (as per [PEP 604](https://www.python.org/dev/peps/pep-0604/)) raise the error:

```
Type not yet supported: pathlib.Path | None
```

### Wanted Solution

`| None` types should work instead of having to need to use `Optional[]`.

### Wanted Code

```python
from pathlib import Path
import typer


def main(output: Path | None):
    typer.echo(output)


if __name__ == "__main__":
    typer.run(main)
```

### Operating System

Linux

### Typer Version

0.4.0

### Python Version

3.10.1

navhits:
I'm thinking if you can do this `output: Path = None`? When you have `Path | None`, the function expects either `Path` or `None` to be passed. Plus I believe Typer is tested for up to Python 3.9. This means most of the features from Python 3.10 might not work as expected. Let's just say certain recursive dependencies used by Typer also might have similar troubles.

SRv6d:
+1
Even using Python 3.10 without the `__future__`  imports PEP 604 isn't supported.
@tiangolo Are there any plans to support this in the near future?

navhits:
Could this be more of a Pydantic issue?

SRv6d:
@navhits I'm not exactly sure what you mean. Pydantic supports PEP 604 type hints.

navhits:
If that's the case, then try building from [python3.10 branch](https://github.com/tiangolo/typer/tree/python3.10). In my understanding 604 is a 3.10 focussed enhancement.

navhits:
Ok. There's a [0.4.1](https://github.com/tiangolo/typer/releases/tag/0.4.1) release already. Try this version @SRv6d

SRv6d:
[...]

PR #664, opened 2023-09-13
🐛 Fix default value of `None` for CLI Parameters when the type is `list | None` and the default value is `None`

theowisear:
If an optional list type option has default_value set to None, we expect its value to also be set to None if no option has been given to the command.

Related to issue #410


svlandeg (commented):
Thanks!

This PR still requires a test to check for the correct behaviour - that also makes it easier to review as there's a minimal example to test on.

svlandeg (approved):
I've added an example in the docs and a unit test, and confirmed that this PR indeed changes the type of the argument to `None` as you would expect.

The fix itself looks good to me, but I'll leave this one up for review by @tiangolo to get his opinion as well.

tiangolo:
Great, thank you @theowisear! :cookie:

And thanks for the help here @svlandeg! :bow: :cake: 

This will be available in Typer 0.10.0, in the next few hours/minutes. :rocket:

johnthagen:
@tiangolo From the edited title of this PR (using `|` syntax) and the subsequent release notes for 0.10.0, I thought (and others too might as well) that this PR fixed

- #348
- #533

But on closer inspection it seems this is a different issue? Are there any plans to review

- #676

or otherwise address `|` support for Typer? Thanks. ❤️

typer/main.py
```diff
@@ -631,9 +631,11 @@ def convertor(value: Any) -> Any:
 
 
 def generate_list_convertor(
-    convertor: Optional[Callable[[Any], Any]]
-) -> Callable[[Sequence[Any]], List[Any]]:
-    def internal_convertor(value: Sequence[Any]) -> List[Any]:
+    convertor: Optional[Callable[[Any], Any]], default_value: Optional[Any]
+) -> Callable[[Sequence[Any]], Optional[List[Any]]]:
+    def internal_convertor(value: Sequence[Any]) -> Optional[List[Any]]:
+        if default_value is None and len(value) == 0:
```
solesensei, line 637: bug. 

If `default_value is None`, then the second condition, which attempts to execute `len(None)`, will always result in an error."

PR #683, opened 2023-10-26
📝 Update docs for building a package, file structure example

davidbgk:
I was a bit confused at first sight about where the `main.py` file was in that tree.

### DIFF
--- a/docs/tutorial/package.md
+++ b/docs/tutorial/package.md
@@ -471,7 +471,8 @@ The file would live right beside `__init__.py`:
 ├── README.rst
 ├── rick_portal_gun
 │   ├── __init__.py
-│   └── __main__.py
+│   ├── __main__.py
+│   └── main.py
 └── tests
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks for your contribution!

I'm not sure whether it's more or less confusing by including the `main.py` file in the tree, potentially taking away the focus from `__main__.py` which is what that part of the documentation is about.

In principle this edit is correct though, I think it's really just a matter of personal preference what is more clear.

davidbgk:
> I'm not sure whether it's more or less confusing by including the `main.py` file in the tree, potentially taking away the focus from `__main__.py` which is what that part of the documentation is about.

In my case, as a newcomer, it was more confusing but I understand it requires more feedback!

tiangolo:
Makes sense, thank you @davidbgk! :nerd_face: 

And thanks @svlandeg for the help! :bow:

PR #721, opened 2023-12-19
🐛 Fix evaluating stringified annotations in Python 3.10 (also `from __future__ import annotations`)

heckad:
Fix case when annotated type inside string or `from __future__ import annotations` used.


svlandeg (commented):
This looks correct to me, and the PR is related to earlier discussions https://github.com/tiangolo/typer/discussions/684 and https://github.com/tiangolo/typer/discussions/598.

I could replicate the issue with Python 3.10 and the fix from this PR indeed seems to resolve it.

tiangolo:
Awesome, thanks @heckad! :rocket: 

And thank you @svlandeg for the help! :bow: 

This will be available in Typer 0.9.3, released in the next few hours. :rocket:

PR #726, opened 2023-12-22
🐛 Make sure `rich_markup_mode=None` disables Rich formatting

liambeguin:
As stated in the documentation, `rich_markup_mode=None` is the default mode.
Make sure that this is true even when rich is installed, fix current tests assuming it's on by default, and add new test cases to validate `rich_markup_mode` options.


svlandeg (commented):
Using the example code snippet from #853, I can confirm that on `master` and in an environment with Rich installed, `rich_markup_mode=None` does not seem to have the desired effect as described in the [documentation](https://typer.tiangolo.com/tutorial/commands/help/#rich-markdown-and-markup):

> By default, rich_markup_mode is None, which disables any rich text formatting.

This PR fixes that, and effectively disables Rich for the given Typer `app`.

I'm putting this PR in draft until the issues with `coverage` are resolved - as some of the tests have been changed, there are code paths no longer covered by the test suite, and we need to fix that before we can merge this.

typer/core.py
```diff
@@ -656,6 +657,7 @@ def main(
         complete_var: Optional[str] = None,
         standalone_mode: bool = True,
         windows_expand_args: bool = True,
+        rich_markup_mode: MarkupMode = None,
```
svlandeg, line 660: I don't think this argument should be added to the `main` function - instead we should use `self.rich_markup_mode`. I'll experiment with this and push to your branch directly.

tests/test_tutorial/test_arguments/test_help/test_tutorial006.py
```diff
@@ -32,5 +33,6 @@ def test_script():
         [sys.executable, "-m", "coverage", "run", mod.__file__, "--help"],
         capture_output=True,
         encoding="utf-8",
+        env={**os.environ, "PYTHONIOENCODING": "utf-8"},
```
svlandeg, line 36: Note: as in #824, not entirely clear to me why we need this, but the Windows tests fail if we don't have it.

typer/core.py
svlandeg: Fixing this also resolved some of the coverage issues.

svlandeg:
Hi @liambeguin or anyone else - if you have time, I'd appreciate a review of https://github.com/tiangolo/typer/pull/859, which I've opened as an alternative to this PR (building on the same idea but without changing the defacto default).

liambeguin:
> Hi @liambeguin or anyone else - if you have time, I'd appreciate a review of #859, which I've opened as an alternative to this PR (building on the same idea but without changing the defacto default).

Hi @svlandeg, thanks for taking the time to look at this. I was a bit confused by the need for a second PR after all the traffic on this one, but if I understand correctly #859 does the same thing but keeps the old behaviour as default and fixes the documentation? I don't mind either way, as long as I can disable it.

Also, I think it would be best to not duplicate PR like this and edit the first one once a consensus is reached.

liambeguin:
Oh I see! well I appreciate you taking the time to fix the coverage tests :) 

It might make more sense to do it the way you did then.

tiangolo:
Thanks @liambeguin! :coffee: 

This was handled in https://github.com/fastapi/typer/pull/859

The fix will be available in Typer `0.12.5`, released in the next few hours. :tada: 

Thanks a lot @svlandeg for all the help as always! :bow:

Issue #334, opened 2021-10-06
typer.Argument with autocompletion


mdantonio:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
Not working:
    myparam: str = typer.Argument(
        "",
        shell_complete=my_autocomplete,
    )

Working:
    myparam: str = typer.Option(
        "",
        shell_complete=my_autocomplete,
    ),
```

### Description

I'm trying to enable shell complete on an Argument but it does not work. On the opposite with Options everything is working as expected

I think that the same problem was already in the previous versions of typer (so that it is NOT introduced by typer 0.4.0 & click 8... but i'm not totally sure about that)

Based on tests on click, shell_complete on arguments should work: https://github.com/pallets/click/blob/48cb86d85fc1abfcf1478ee660e42aaa5382fd64/examples/completion/completion.py#L25

in typer I can't find tests with arguments, I only found this with option https://github.com/tiangolo/typer/blob/a1520dcda685220a9a796288f5eaaebd00d68845/tests/assets/compat_click7_8.py#L20

### Operating System

Linux

### Operating System Details

Ubuntu 20.04

### Typer Version

0.4.0

### Python Version

Python 3.9.5

### Additional Context

click                  8.0.1

patricksurry:
this (mostly) works for me using `typer.Argument('default', autocompletion=my_autocomplete)` tho am running into a weird bug with pwsh/win11 vs bash/osx

mdantonio:
@patricksurry autocompletion works, but it is deprecated in click

```
DeprecationWarning: 'autocompletion' is renamed to 'shell_complete'. The old name is deprecated and will be removed in Click 8.1. See the docs about 'Parameter' for information about new behavior.
```


I'm still unable to make Arguments work with shell_complete.

On the opposite, shell_complete works very well with Options.

This is the situation, at least for me:

|          | autocomplete | shell_completion |
|----------|--------------|------------------|
| Option   | :heavy_check_mark:             | :heavy_check_mark:                 |
[...]

PR #737, opened 2024-01-25
🐛 Fix `shell_complete` not working for Arguments

bckohan:
As pointed out in #334 the shell_complete parameter on Arguments is not called. This is a very simple bug where [shell_complete was simply omitted as a parameter passed to TyperArgument](https://github.com/tiangolo/typer/blob/3a7264cd56181690805f220cb44a0301c0fdf9f3/typer/main.py#L918-L942).

This PR adds the missing parameter resulting in a one line change in the library and it adds one additional completion test that tests that shell_complete works as expected when given to an Argument.


typer/main.py
```diff
@@ -936,6 +936,7 @@ def get_click_param(
                 expose_value=parameter_info.expose_value,
                 is_eager=parameter_info.is_eager,
                 envvar=parameter_info.envvar,
+                shell_complete=parameter_info.shell_complete,
```
dbosk, line 939: I'm thinking that this one might require the use of `get_param_completion` same as the `autocompletion` on the line below (940). Looking at the code of `get_param_completion` it has some features that are probably applicable here too: mapping the parameters so that not all completion parameters are required (e.g. one can use just the `incomplete: str` instead of needing also the context.

tests/assets/compat_arg_complete_click7_8.py
```diff
@@ -0,0 +1,22 @@
+import click
+import typer
+
+app = typer.Typer()
+
+
+def shell_complete(ctx: click.Context, param: click.Parameter, incomplete: str):
```
dbosk, line 7: It would also be nice to test without `ctx` and `param`. In many cases I just use something like `shell_complete(incomplete: str)`. I think this requires the use of `get_param_completion`, see my other comment.

dbosk (commented):
I think we need `get_param_completion` to be able to skip some of the parameters for the `shell_complete` function, such as `shell_complete(incomplete: str)`.

typer/main.py
bckohan: Hi Daniel, good catch! I had the same thought, bit if you look just above at line 905 you'll see shell_complete is not wrapped in get_param_completion for Options. Also, the implementation of get_param_completion only works for the autocomplete interface where the second argument is the list of string arguments appearing on the command line before the parameter in question where as in the new shell_complete interface the middle parameter is an instance of click.Parameter.

I do not disagree that it would be nice to wrap shell_complete with get_param_completion like behavior because often the context and Parameter are unnecessary - but this would be a deeper change that would also touch an existing interface and it was my hope to keep this PR as simple as possible to hopefully get it in a 0.9.1 bug fix release. If there's no expectation of that, I'd be happy to submit a chunkier PR for a 0.10 release...

dbosk: Makes sense as you say. I haven't looked at the rest of the code.

svlandeg (approved):
Nice catch! The one line fix seems correct to me, and I can confirm that it fixes the bug that was reported earlier in #334.

Thanks for adding a test! I just moved it out of `compat` and into `test_others.py` where there were a few other `completion` tests as well. Not sure if this is where Tiangolo wants it or someplace else 😇

tiangolo:
Great, thank you @bckohan! :rocket: 

And thanks a lot for the review and help @svlandeg! :bow: 

This will be available in Typer `0.12.4` released in the next few hours. :tada:

PR #676, opened 2023-10-03
🐛 Fix support for UnionType with Python 3.11

standag:
Resolves #533

PR #739, opened 2024-01-27
Support for optional as UnionType (Python >= 3.10)

mkomet:
This resolves the issue https://github.com/tiangolo/typer/issues/533

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -46,6 +46,12 @@ try:
 except ImportError:  # pragma: nocover
     rich = None  # type: ignore
 
+if sys.version_info >= (3, 10):
+    from types import UnionType
+else:
+    # Python < 3.10 doesn't have UnionType, so we define it manually as non inheritable type
+    UnionType = type("", (), {})
+
 _original_except_hook = sys.excepthook
 _typer_developer_exception_attr_name = "__typer_developer_exception__"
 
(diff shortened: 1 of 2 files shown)


johnthagen:
Also closes 

- #348

Another PR that also address this issue

- #676

typer/main.py
```diff
@@ -816,8 +822,9 @@ def get_click_param(
     is_tuple = False
     parameter_type: Any = None
     is_flag = None
+    is_union_type = lenient_issubclass(type(main_type), UnionType)
```
duongdominhchau, line 825: Hi, I'm new to this kind of stuff, could you please clarify this for me please? From the variable name, it is a `type` already, why do we need to convert it to `type()` here? Is it just your preference or there are some edge cases being handled implicitly here?

svlandeg (commented):
Hi, thanks for this PR!

There's a few other PRs that address the same issue, of which I think #548 would be the prime candidate to merge. That PR uses existing functionality such as `utils.needs_py310` and `._typing.is_union`.  If we merge #548, I think we can go ahead and close this one.

I'll leave the final decision with Tiangolo though.

tiangolo:
Thank you for the work @mkomet! :coffee: 

And thanks @svlandeg for the help :bow: 

This should have been fixed by https://github.com/fastapi/typer/pull/548

It is now available in Typer `0.12.4` :rocket: 

Given that, I'll close this one.

PR #745, opened 2024-03-01
Add StrEnum example for python 3.11+

Matthieu-LAURENT39:
Python 3.11 added StrEnum, that can replace using `MyEnum(str, Enum)`.

This adds that usage to the Enum parameter type docs, while also keeping the old usage for users on older versions of Python.

### DIFF
--- a/docs/tutorial/parameter-types/enum.md
+++ b/docs/tutorial/parameter-types/enum.md
@@ -1,8 +1,18 @@
 To define a *CLI parameter* that can take a value from a predefined set of values you can use a standard Python <a href="https://docs.python.org/3/library/enum.html" class="external-link" target="_blank">`enum.Enum`</a>:
 
-```Python hl_lines="1  6 7 8 9  12 13"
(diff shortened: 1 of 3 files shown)


svlandeg:
Thanks for the contribution! I'll update this with `master` and add a test to double check the behaviour 🙏

svlandeg:
Quick update: we'll pause this PR for a bit, as we want to put a mechanism in place first to have source examples for different Python versions, as well as corresponding tests that require specific Python versions too. We'll get back to this PR once we've taken care of that!

svlandeg (commented):
I think that this new format of supporting multiple versions of the same script in the docs with the `{* *}` syntax might not yet be supporting Python 3.11. Will look into this and get back to this PR afterwards.

Update: I think this needs https://github.com/tiangolo/markdown-include-variants/pull/28 to be merged & released first.

svlandeg:
Great, this now works:

image

Leaving this to Tiangolo for a final review!

svlandeg (approved):

PR #750, opened 2024-03-18
🐛 Add missing `default_factory` in `Argument` overloads

m9810223:


svlandeg (approved):
Thanks for the contribution, nice catch!

These parameters are indeed missing in both overload declarations of `Argument`, and you've inserted them in the right spot. Thanks again!

tiangolo:
Good catch, thanks @m9810223! :rocket: 

And thank you @svlandeg for the review! :bow:

PR #755, opened 2024-03-22
📝 Update docs example for a Typer/Click group to make new subcommands explicit

svlandeg:
In the [documentation](https://typer.tiangolo.com/tutorial/using-click/#including-a-click-app-in-a-typer-app), an example is given of a Typer app that includes a Click app. The example reads

> // Notice we have both subcommands, top and hello

But for the user to actually see this, they should call `main.py` with `--help` specifically. I've updated the docs accordingly, and added a unit test to ensure that both commands are indeed shown in the help function.

### DIFF
--- a/docs/tutorial/using-click.md
+++ b/docs/tutorial/using-click.md
@@ -94,12 +94,19 @@ Check it:
 <div class="termy">
 
 ```console
-$ python main.py
+$ python main.py --help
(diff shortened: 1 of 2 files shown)


tiangolo:
Great, thank you @svlandeg! :cake:

PR #757, opened 2024-03-23
🐛 Add support for an argument of type `Optional[Tuple]` and default value `None`

Asthestarsfalll:
Support optional tuples like this:
```python
from typing import Optional, Tuple
import typer

def main(arg: Optional[Tuple[int]] = None):
    pass

typer.run(main)
```


svlandeg (approved):
I relabeled this as a bug, as it feels like this is functionality that you'd expect to work, but currently doesn't on `master`. The fix seems correct to me, and I've slightly rewritten the unit test to be even more in line with the preceding `test_optional()`.  

Thanks for your contribution!

tiangolo:
Great, thanks for your contribution @Asthestarsfalll! :rocket: 

And thanks for the help @svlandeg! :bow: :cake:

PR #770, opened 2024-03-27
feat: Add `pretty_exceptions_suppress` option to extend suppressed frames in pretty exceptions

tlambert03:
This pull request extends the pretty_exceptions feature by adding `pretty_exceptions_suppress`, which lets the user pass an iterable of modules or string paths to suppress in addition to Typer and clic

### DIFF
--- a/docs/tutorial/exceptions.md
+++ b/docs/tutorial/exceptions.md
@@ -202,6 +202,18 @@ $ python main.py
 
 </div>
 
+## Suppress Frames from Pretty Exceptions
+
+By default, typer will **omit** all the parts of the traceback that come from the internal parts in Typer and Click, but you can exclude frames from additional
+frameworks using `pretty_exceptions_suppress`, which should be a list of modules or str paths.
+
+```Python
+import typer
+import sqlalchemy
+
+app = typer.Typer(pretty_exceptions_suppress=[sqlalchemy])
+```
+
 ## Disable Pretty Exceptions
 
 You can also entirely disable pretty exceptions with the parameter `pretty_exceptions_enable=False`:
--- a/typer/main.py
+++ b/typer/main.py
@@ -7,8 +7,8 @@ from enum import Enum
 from functools import update_wrapper
 from pathlib import Path
 from traceback import FrameSummary, StackSummary
-from types import TracebackType
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Type, Union
+from types import ModuleType, TracebackType
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tuple, Type, Union
 from uuid import UUID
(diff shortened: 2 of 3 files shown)


Mark90:
I'm not a maintainer, but the change looks pretty straight-forward to me.
Would be great if this option could be added.
Maybe this can be rebased and reviewed, or does it need to be linked to a feature issue first?

tlambert03:
fixed conflicts.  Would be great to get a review from any maintainer.  @tiangolo, @svlandeg?  This has sat here for quite a while now.  Do you have any thoughts?

svlandeg:
I'll have a look! 🙏

docs/tutorial/exceptions.md
```diff
@@ -196,6 +196,18 @@ $ python main.py
 
 </div>
 
+## Suppress Frames from Pretty Exceptions
```
svlandeg, line 199: There's a few things we need to do to improve the docs here:

- Move this section up to BEFORE the section "Disable Short Output"
- Anywhere on this page, whenever it talks about "Click and Typer" being excluded, it should instead reference the "suppressed libraries" (or such, depending on exact wording used in the new section)
- The example given here should be a full example, as in all other sections on this page. That means it should be a fully runnable code example that has `sqlmodel` lines in its error stack trace, and the output should be displayed so the user can verify it
- The example should be tested properly in the test suite at `tests/test_tutorial/test_exceptions`, at which point it should check e.g. that `sqlalchemy` is part of the output (or not) with the setting off (or on). 

BTW - is "frames" the correct term here? I would have opted for "Libraries", but I'm not sure either way.

docs/tutorial/exceptions.md
```diff
@@ -196,6 +196,18 @@ $ python main.py
...
+```Python
+import typer
+import sqlalchemy
+
+app = typer.Typer(pretty_exceptions_suppress=[sqlalchemy])
+```
+
 ## Disable Pretty Exceptions
 
 You can also entirely disable pretty exceptions with the parameter `pretty_exceptions_enable=False`:
```
svlandeg, line 213: Question: what should happen when both `pretty_exceptions_short=False` and `pretty_exceptions_suppress=[sqlalchemy]` are set?

docs/tutorial/exceptions.md
```diff
@@ -196,6 +196,18 @@ $ python main.py
 
 </div>
 
+## Suppress Frames from Pretty Exceptions
+
+By default, typer will **omit** all the parts of the traceback that come from the internal parts in Typer and Click, but you can exclude frames from additional
+frameworks using `pretty_exceptions_suppress`, which should be a list of modules or str paths.
```
svlandeg, line 202: How exactly would "str paths" be defined?

PR #776, opened 2024-03-29
WIP: test docs CI

svlandeg:
PR to investigate potential issue with the docs CI. Noticed on other PRs, but I wanted to isolate the behaviour.

### DIFF
--- a/docs/tutorial/printing.md
+++ b/docs/tutorial/printing.md
@@ -104,7 +104,7 @@ But for advanced use cases, you could create a `Console` yourself.
 {!../docs_src/printing/tutorial003.py!}
 ```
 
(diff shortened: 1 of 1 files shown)


svlandeg:
Confirmed that CI works just fine!

tiangolo:
The problem of docs not building was because I had an incorrect config, fixed here: https://github.com/tiangolo/typer/pull/778

Then there was a problem that building docs was raising an error, that's because of an issue in MkDocs Material, handled in a workaround here: https://github.com/tiangolo/typer/pull/779 (disabling the social plugin for now).

I think that should resolve it, so I'll close this one. :nerd_face:

PR #780, opened 2024-03-30
✨ Add `typer-slim` package without extras, make `typer` include `typer-slim[default]` and integrate Typer CLI (`typer` command) into Typer

tiangolo:
✨ Add `typer-slim` package without extras, make `typer` include `typer-slim[default]` and integrate Typer CLI (`typer` command) into Typer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,9 @@ requires = ["pdm-backend"]
 build-backend = "pdm.backend"
 
 [project]
-name = "typer"
+# pip install typer-slim: is only the Typer library, without extras
+# pip install typer: is typer-slim[standard], with extras and typer-cli
+name = "typer-slim"
 dynamic = ["version"]
 description = "Typer, build great CLIs. Easy to code. Based on Python type hints."
 authors = [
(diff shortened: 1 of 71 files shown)


pyproject.toml
```diff
@@ -3,7 +3,9 @@ requires = ["pdm-backend"]
 build-backend = "pdm.backend"
 
 [project]
-name = "typer"
+# pip install typer-slim: is only the Typer library, without extras
+# pip install typer: is typer-slim[standard], with extras and typer-cli
+name = "typer-slim"
```
FCamborda, line 8: Perhaps you want to update the documentation to let users know that `typer` actually includes `typer-slim` and other packages?
This PR offuscates that and this change breaks users of [FawltyDeps](https://github.com/tweag/FawltyDeps). We had declared `typer` as a dependency (-> failed), but it's now the project `typer-slim` the one that has a `typer` package.

svlandeg: Hi @FCamborda! Could you open an discussion thread with this question? Because I worry that it might otherwise get lost on this merged PR 🙏

FCamborda: @svlandeg Sure! Let me know if this needs improvement: https://github.com/tiangolo/typer/discussions/786

PR #796, opened 2024-04-09
🐛 Fix Rich formatting with no commands

svlandeg:
When taking the `max` of the list of commands, define a `default` value to guard against calling `max` on an empty list, which would raise an error. (this is also how I had it in the original [suggestion](https://github.com/tiangolo/typer/pull/567#discussion_r1553956210))

Fixes https://github.com/PrefectHQ/prefect/issues/12617
Fixes https://github.com/tiangolo/typer/discussions/795

### DIFF
--- a/tests/test_rich_utils.py
+++ b/tests/test_rich_utils.py
@@ -39,7 +39,7 @@ def test_rich_utils_click_rewrapp():
 
 
 def test_rich_help_no_commands():
-    """Ensure that the help still works for a Typer instance wit no commands, but with a callback."""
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
I'm not really sure how to design a unit test for this 🤔

mattkram:
Thanks for the implementation @svlandeg! Beat me to the chase!

tiangolo:
Nice, thank you! @svlandeg :cake:

Sorry I didn't realize we needed the default you had originally suggested. :grimacing: 

This will be available in the next minutes in Typer 0.12.3. :rocket:

PR #800, opened 2024-04-17
✨ Implement list parsing from string with separators

### LINKED ISSUE #554: DOC: Documentation of passing multiple values in "option" vs "argument" isn't sufficiently explicit
### Privileged issue

### Issue Content

I was looking into passing multiple (N>=1, all N >=1 allowed) white space separated values via an option.

This does not seem possible at the moment, but the documentation wasn't very clear. It's _only_ possible to pass an exact number of space separated values via an option, but not a list: https://typer.tiangolo.com/tutorial/multiple-values/options-with-multiple-values/

Multiple values are only possible via argument, yet the documentation for the two types reads almost identical (at least the first few paragraphs): https://typer.tiangolo.com/tutorial/multiple-values/arguments-with-multiple-values/

It would be good to be precise in the first paragraph about the fact that options don't allow you to pass N>=1 values.

Maybe one could accept multiple values by specifying a separator that's not white space, like comma-separated - but that's a feature request in a bit different direction.

libklein:
Implements feature #554.

This merge request introduces the functionality to parse List[T] options from strings using a specified separator. An example of how this can be utilized is demonstrated [below](https://github.com/tiangolo/typer/blob/a6a797dc7958ac2cbfbae4c70bab2ee6d34fc353/docs_src/multiple_values/multiple_options/tutorial003_an.py):

```python
def main(number: List[float] = typer.Option([], multiple_separator=",")):
    print(f"The sum is {sum(number)}")

#  Passing --number 1,2,3 outputs 6  
#  Works with --number 1,2 --number 3 as well. Output would be 6 again.
```

The implementation works by overwriting the `process_value` function of the click parser, doing string splitting before forwarding the result to click. This ensures full compatibility with the existing features of both Click and Typer. The feature does not allow whitespace only separators to avoid ambiguities. 

Feedback is much appreciated!

Open questions:
* Currently, error handling is implemented in the constructor of `typer.Option`, which delays error detection until the command execution. Should we consider moving this error handling to the `OptionInfo` constructor to allow for earlier failure detection?
ToDo:
- [x] Add to doc_src
- [x] Write tests
- [x] Update documentation


svlandeg:
Hi, thanks for your PR! I'll put this in draft as long as the test suite is failing.

typer/core.py
```diff
@@ -419,6 +421,7 @@ def __init__(
         show_envvar: bool = False,
         # Rich settings
         rich_help_panel: Union[str, None] = None,
+        multiple_separator: Optional[str] = None,
```
ivantodorovich, line 424: I think just `separator` would get straight to the point here, no need to preprend with "multiple" IMO

libklein: Hmm, I agree that multiple_separator is a poor choice. It's too verbose and does not communicate it's intended usage well. How do you feel about item_separator or list_separator? I think separator may be a bit too generic. It would not be immediately clear to me that the separator keyword argument is legal only for list types.

ivantodorovich: IMO it's not a problem if it's too generic.

Now it's only legal for list types, true, but that doesn't mean we can't give it another meaning if we find another use case for a "separator" on other types

Take for example the existing file parameters: https://github.com/tiangolo/typer/blob/3cdd682244358cf9108291ce8695a93878c46f92/typer/params.py#L54-L59

None of them have a prefix, and yet it's fine.

Same for the existing Number parameters: https://github.com/tiangolo/typer/blob/3cdd682244358cf9108291ce8695a93878c46f92/typer/params.py#L48-L51

And others like Choise or Datetime

libklein: I see your point. I've updated the PR accordingly.

ivantodorovich (approved):
Thanks 💚

docs/tutorial/multiple-values/multiple-options.md
```diff
@@ -87,3 +87,51 @@ The sum is 9.5
 ```
 
 </div>
+
+## Passing multiple values in a single argument
+
+Many users expect to be able to pass multiple arguments with a single .
```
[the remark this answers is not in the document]
libklein: Good catch. Updated and elaborated a bit more.

svlandeg (approved):
Hi, thanks for the PR! It's really well written, with the appropriate changes and additions to the documentation, example code and unit tests. Really appreciated 🙏 

I pushed a few smaller edits to update the documentation as well as pointing out that it does in fact work without quotes, as long as there is no whitespace inbetween, e.g. 

```
python main.py --number 2,3,4.5
```

The PR looks good to merge to me, but I defer to Tiangolo for the final decision.

libklein:
Is there anything I can contribute?

svlandeg:
At this point the PR is in a good shape to be reviewed by Tiangolo - we've got this on our internal board for him to review when has time 😉  Thanks for your patience!

DvdNss:
Hey there,

I coded a quick workaround for the "Typer doesn't support whitespace-separated multi-value options" (sorry won't have time for a PR) after moving from argparse to typer for a big project.

My users are used to running 'app some_command --filters filter1 filter2 filter3 --envs env1 env2 env3' and I thought using 'app --filters filter1 --filters filter2 --filters filter3 --envs env1 --envs env2 --envs env3' would be too much of a gap/pain for them and I, so I simply wrote a quick function that preprocesses the sysargv from whitespace-separated multi-value options to the typer format before they're given to typer.

For anyone who needs it as well:
`def preprocess_args():
    """
    Typer doesn't support whitespace-separated multi-value options.

    We preprocess the sysargv so that:
    - python3 app.py some_command --filters filter1 filter2 filter3 --environments env1 env2 env3

    becomes:
    - python3 app.py some_command --filters filter1 --filters filter2 --filters filter3 --environments env1 --environments env2 --environments env3

    //!\\ DOWNSIDE: options should always be after arguments in the CLI command //!\\
    """

    logger.debug(f"Initial CLI command is: {sys.argv}")

    # get main cmd
    final_cmd = []
    for idx, arg in enumerate(sys.argv):
        if any(arg.startswith(_) for _ in ['-', '--']):
            break
        else:
            final_cmd.append(arg)
    logger.debug(f"Main command is: {final_cmd}")

    # get options and their values
    for idx, arg in enumerate(sys.argv):
        if any(arg.startswith(_) for _ in ['-', '--']):
            opt_values = [] 
            for value in sys.argv[idx+1:]:
                if any(value.startswith(_) for _ in ['-', '--']):
                    break
                else:
                    opt_values.append(value)
            
            [final_cmd.extend([arg, opt_value]) for opt_value in opt_values]

    # replace by reformatted
    logger.debug(f"Final command is: {final_cmd}")
    sys.argv = final_cmd
    

if __name__ == "__main__":
    preprocess_args()
    app()`

Not that the sysargv preprocessing above won't be taken into account in your test files if you use the typer.testing CliRunner, so in your tests you'll have to use the typer format instead.

Happy long weekend to all of you

misse:
I'd love to use this!

misse:
@libklein would this work for Enum type choices too?

dylankiss:
@tiangolo Little loving nudge ☝️ 😅

libklein:
@misse Yes, it is compatible with all of typer's (and click's) features.

libklein:
Hi @svlandeg,

more than a year has passed since creating the PR. Are there any updates concerning the roadmap?

svlandeg:
Hi @libklein,

I have updated the PR to adhere to the new documentation format, and it is still awaiting further review by Tiangolo. It has been put on our internal review queue. I'd like everyone to refrain from adding additional pings, as that just creates a longer PR thread and generates a lot of noise in our notifications on our end. Instead, if you'd like this PR to be merged, please add an emoticon on the very first post ☝️, which gives us a good idea on how "popular" the idea/PR is.

I do realise it can be a bit frustrating to have this PR sit here for so long. There's just a lot of working in maintaining several open-source repo's (Typer, FastAPI, SQLModel, ...) and we do our best to keep up. Thanks for your understanding! 🙏 

> Hi @svlandeg,
> 
> more than a year has passed since creating the PR. Are there any updates concerning the roadmap?

tiangolo:
Thanks for the interest and work on this! ☕

I want to implement a few big refactors and features that might solve many of these use cases.

I also need to do that before expanding the surface of what is supported with additional external PRs, because that also means I would have to keep maintaining and handling backwards compatibility for those new changes while doing those big refactors.

I have not wanted to close PRs that target the same or similar problems before I have an alternative, but then I have left many PRs open for too long, which is not useful or clear either. Sorry for that.

And then, after I make those big refactors, the problem might be solved... and if not, a bunch of conflicts will show up, so it will be better and easier to start from scratch (from the new code state).

So, I'll close this one for now to keep things clean.

For the specific use case, the best would be to document it in a new GitHub Discussion (or 👍 any existing one) with the specific problem to solve (without details of the implementation) to track the feature request.

Thank you! 🍰

PR #801, opened 2024-04-19
🐛  Fix sourcing of completion path for Git Bash

svlandeg:
## My system
When running 

```
typer --install-completion
```

on my Windows system with Git Bash, Typer fetches `Path.home()` which is `C:\Users\Sofie`. This is correct in itself, but Git Bash can't work directly with paths that contain backslashes. After installation,  `.bashrc` contains
```
source C:\Users\Sofie\.bash_completions\typer.sh
```

and after rebooting the terminal, it will complain:

```
bash: C:UsersSofie.bash_completionstyper.sh: No such file or directory
```

To get this to work with Git Bash, `.bashrc` should instead contain
```
source 'C:\Users\Sofie\.bash_completions\typer.sh'
```
and then it works - no error upon starting the console, and the completion feature works.

## TODO

This PR needs to be carefully checked on a "normal" Bash system to double check that the single quote doesn't harm any other environments. I wouldn't think so, but I'm not 100% sure.


typer/_completion_shared.py
```diff
@@ -100,13 +100,13 @@ def install_bash(*, prog_name: str, complete_var: str, shell: str) -> Path:
     # It seems bash-completion is the official completion system for bash:
     # Ref: https://www.gnu.org/software/bash/manual/html_node/A-Programmable-Completion-Example.html
     # But installing in the locations from the docs doesn't seem to have effect
-    completion_path = Path.home() / f".bash_completions/{prog_name}.sh"
+    completion_path = Path.home() / ".bash_completions" / f"{prog_name}.sh"
```
svlandeg, line 103: Note that this particular fix is not required to solve the issue on my end. I just think it's safer to always use the `Pathlib` separator `/` instead of the literal `"/"`.

ivantodorovich: It really makes no difference because `Pathlib` perfectly understands `/` in paths 😊 it's designed to be cross-platform

typer/_completion_shared.py
```diff
@@ -100,13 +100,13 @@ def install_bash(*, prog_name: str, complete_var: str, shell: str) -> Path:
...
     rc_path = Path.home() / ".bashrc"
     rc_path.parent.mkdir(parents=True, exist_ok=True)
     rc_content = ""
     if rc_path.is_file():
         rc_content = rc_path.read_text()
-    completion_init_lines = [f"source {completion_path}"]
+    completion_init_lines = [f"source '{completion_path}'"]
```
ivantodorovich, line 109: It's safe to quote in posix, too.
In fact it's the right thing to do, even on posix, because the home directory might contain spaces

However, I recommend quoting using `shlex.quote`, which will also convert other corner cases

typer/_completion_shared.py
svlandeg: Yea I think you're right - it makes no difference because `Pathlib` will deal with the literal string correctly, too. Honestly I'd still prefer the edit for esthetic reasons (part of the same path is using `/`, the other part is using `"/"`) but I don't feel strongly. Will leave it up to Tiangolo - and either way as I said this is not actually related to the bug fix.

ivantodorovich: Hey, I also think it's more esthetic this way 😉 just pointing out it doesn't really make it any safer

typer/_completion_shared.py
svlandeg: `shlex.quote` isn't actually supposed to be used on Windows, it's marked in the documentation as being unsafe and unreliable: https://docs.python.org/3/library/shlex.html#shlex.quote.

svlandeg: I'd very much prefer to avoid using `shlex.quote` altogether. I'd propose to merge this PR as is, and see whether any further issues get reported that would warrant a more complex solution.

tiangolo:
Awesome, thank you @svlandeg! :bow: 

I just tested it here on bash under Docker and it's all working.

PR #803, opened 2024-04-24
✨ Support for pydantic options

### LINKED ISSUE #111: [FEATURE] Support Pydantic Models as ParamTypes

pypae:
## 📝 Summary
This PR adds support for pydantic options as requested in the following issue: #111
In particular it implements the behavior I described in https://github.com/tiangolo/typer/issues/111#issuecomment-643147865.

The implementation changes the signature of the callback to add all fields of the (possibly nested) `pydantic` models of the original callback as `typer.Option`s. As a result the implementation is largely independent of `typer` and could also be used as a standalone package which could be activated with an additional decorator on the command functions:

<details>
<summary>Usage as a decorator (draft)</summary>

```python
import pydantic
import typer
from typer_pydantic import unwrap_pydantic_models


class User(pydantic.BaseModel):
    id: int
    name: str = "Jane Doe"

@unwrap_pydantic_models
def main(num: int, user: User):
    print(num, type(num))
    print(user, type(user))


if __name__ == "__main__":
    typer.run(main)

```

</details>

## 🔀 Related PRs
- https://github.com/tiangolo/typer/pull/723 (complementary)
- https://github.com/tiangolo/typer/pull/329 (alternative)
- https://github.com/tiangolo/typer/pull/630 (alternative)

## ☑️ To Do
- [x] Make linter happy
- [ ] Describe advantages over other approaches
- [ ] Add support for `typer.Option` and `typer.Argument` inside `pydantic` models
- [x] Add tests for nested `pydantic` models
- [ ] Make sure validation errors are reported properly
- [x] Make pydantic dependency optional
- [ ] ~~Allow sequences of pydantic models.~~ I suggest implementing indexed list options first, see the [comment below](https://github.com/tiangolo/typer/pull/803#issuecomment-2079341538).

### DIFF
--- a/docs_src/parameter_types/pydantic/tutorial001.py
+++ b/docs_src/parameter_types/pydantic/tutorial001.py
@@ -1,5 +1,6 @@
-import pydantic
 import typer
+
+import pydantic
 
 
 class User(pydantic.BaseModel):
--- a/docs_src/parameter_types/pydantic/tutorial001.py
+++ b/docs_src/parameter_types/pydantic/tutorial001.py
@@ -1,0 +1,16 @@
+import pydantic
+import typer
+
+
+class User(pydantic.BaseModel):
+    id: int
+    name: str = "Jane Doe"
+
+
+def main(num: int, user: User):
+    print(num, type(num))
+    print(user, type(user))
+
+
+if __name__ == "__main__":
+    typer.run(main)
--- a/docs_src/parameter_types/pydantic/tutorial002.py
+++ b/docs_src/parameter_types/pydantic/tutorial002.py
@@ -1,0 +1,24 @@
+from typing import List, Optional
+
+import typer
+
+import pydantic
+
+
+class Pet(pydantic.BaseModel):
+    name: str
+    species: str
+
+
+class Person(pydantic.BaseModel):
+    name: str
+    age: Optional[float] = None
+    pet: Pet
+
+
+def main(person: Person):
+    print(person, type(person))
+
+
+if __name__ == "__main__":
+    typer.run(main)
--- a/docs_src/parameter_types/pydantic/tutorial002.py
+++ b/docs_src/parameter_types/pydantic/tutorial002.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import Optional
 
 import typer
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ homepage = "https://github.com/tiangolo/typer"
 standard = [
     "shellingham >=1.3.0",
     "rich >=10.11.0",
+    "pydantic >= 2.0.0",
 ]
 
 [tool.pdm]
--- a/tests/test_tutorial/test_parameter_types/test_pydantic/test_tutorial001.py
+++ b/tests/test_tutorial/test_parameter_types/test_pydantic/test_tutorial001.py
@@ -1,6 +1,7 @@
 import subprocess
 import sys
 
+import pytest
 import typer
 from typer.testing import CliRunner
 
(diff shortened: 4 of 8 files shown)


pypae:
### Problem
If we want to allow sequences of pydantic models, things get unreadable quite fast. 

Let's look at an example of a nested pydantic model with a sequence of another model:
`pets.py`
```python 
from typing import Optional, List

import typer

import pydantic


class Pet(pydantic.BaseModel):
    name: str
    species: str


class Person(pydantic.BaseModel):
    name: str
    age: Optional[float] = None
    pets: List[Pet]


def main(person: Person):
    print(person, type(person))


if __name__ == "__main__":
    typer.run(main)
```

The script could be called like this:
```python 
$ python pets.py --person.name Jeff --person.pets.name Lassie --person.pets.species dog
```

If we want to add multiple pets, we can just supply `--person.pets.name` and `person.pets.species` multiple times.
```python 
$ python pets.py --person.name Jeff --person.pets.name Lassie --person.pets.species dog --person.pets.name Nala --person.pets.species cat
```
We don't explicitly state which pet names and species belong together and have to rely on the correct order of parameters. In my opinion this is potentially confusing for the CLI user and may lead to bugs.
 
### Potential Solution
To make the mapping more explicit, we could allow to enable `typer.Option` lists to be indexed.
Like for nested pydantic models, I suggest sticking to the syntax `traefik` uses for lists. I.e. [`entrypoints.<name>.http.tls.domains[n].main`](https://doc.traefik.io/traefik/reference/static-configuration/cli/)

Indexed lists could be implemented independently of this PR and should work for all lists. 
I suggest adding an `indexed` flag on `typer.Option` like shown in the example below.

`indexed_list.py`
```python 
from typing import List

import typer

def main(indexed_list: List[int] = typer.Option(..., indexed=True)):
    print(indexed_list)

if __name__ == "__main__":
    typer.run(main)
```

This would then produce the following help text: 
```bash
$ python pets.py --help
                                                                                
 Usage: pets.py [OPTIONS]                                    
                                                                    
╭─ Options ────────────────────────────────────────────────────────╮
│ *  --indexed-list[n]        INTEGER   [default: None] [required] │
│    --help                             Show this message and      │
│                                       exit.                      │
╰──────────────────────────────────────────────────────────────────╯
```

And could be used like this:

```bash
$ python pets.py --indexed-list[1] 0 --indexed-list[0] 1 --indexed-list[2] 2
[1, 0, 2]
```
*Note how the order of the input parameters doesn't matter anymore because the indices are given explicitly.*

### Notes on Implementation

Implementing this might not be trivial, but I think it could be possible by forwarding unknown options as [described in the `click` docs](https://click.palletsprojects.com/en/8.1.x/advanced/#forwarding-unknown-options).

---
Do you have any opinions on this?

pypae:
I started working on a standalone package to support this functionality, so you can use `pydantic` models with typer right now: https://github.com/pypae/pydantic-typer.

It's still WIP, and for now only covers the same basic behavior as this PR.

svlandeg (commented):
I haven't reviewed this in detail yet, but just quickly a few administrative points that we'd need to address if we do want to go forward with this PR:

- The tutorial files should be updated to use `Typer()` instead of `typer.run`, the new format since https://github.com/fastapi/typer/pull/1418
- There should be a new markdown file `docs/tutorial/parameter-types/pydantic.md` explaining this and referencing the tutorial files. This file should be included in `mkdocs.yml`.

svlandeg (commented):
Closing this PR as there is ongoing work in the pipeline to address this more fundamentally, e.g. https://github.com/fastapi/typer/pull/1831

PR #669, opened 2023-09-20
✨ Add support for Literal type

pchanial:
Fixes #76

PR #805, opened 2024-04-24
🔥 Remove unused functionality from `_typing.py` file

ivantodorovich:
It was only used to import the `get_args` and `get_origin` methods, but we can import them from `typing_extensions` directly.


svlandeg (commented):
Hi, thanks for the contribution! It would be great to clean this up.

As you're deleting quite a bit of code though, can you elaborate a bit more on why it's all become obsolete?

typer/_typing.py
```diff
@@ -1,630 +0,0 @@
...
-    "ReprArgs",
-    "AnyClassMethod",
-    "CallableGenerator",
-    "WithArgsTypes",
-    "get_args",
-    "get_origin",
-    "get_sub_types",
-    "typing_base",
-    "get_all_type_hints",
-    "is_union",
```
svlandeg, line 343: Note that the unmerged PR https://github.com/tiangolo/typer/pull/548 uses `is_union` (as well as  `get_args` and `get_origin`) so either way I'd recommend merging that PR before we handle this one.

ivantodorovich:
> Hi, thanks for the contribution! It would be great to clean this up.
> As you're deleting quite a bit of code though, can you elaborate a bit more on why it's all become obsolete?

This file was extracted from pydantic, before pydantic removed it with its support of python3.6
The header said:
```Copied from pydantic 1.9.2 (the latest version to support python 3.6.)```

As mentioned above, the only 2 methods used in this big file were `get_args` and `get_origin`, which can easily be imported from the maintained `typing_extensions` package.

My guess is that this file was useful for python3.6, but now it's not, as libraries like `typing_extensions` or even the standard library have grown and include other alternatives

**Review by svlandeg — COMMENTED** [review]
Currently, `main.py` uses the following functionality:
```
from ._typing import get_args, get_origin, is_literal_type, is_union, literal_values
```

Further, the `_typing.py` file has evolved from supporting 3.6-specific functionality to supporting other functions that may differ in Python versions 3.8 or 3.10 etc. As such, simply removing this feels too risky to me.

Instead, I would suggest to merge my PR https://github.com/fastapi/typer/pull/850 which only cleans up Python 3.6 related code, and go from there. I'll leave the final decision with Tiangolo though.

Either way - thanks again for the contribution! 🙏

ivantodorovich:
Hello @svlandeg !

I understand your concerns.

At the time of my PR only two methods were imported and used from the `_typing.py` module: `get_args` and `get_origin`, and simply using their `typing_extensions` counterparts worked perfectly fine.

However, I see new developments have changed that situation since then, and so it would require more analysis, and the solution might not be just as simple anyway.

I do believe we should aim to drop this file and favor more battle tested alternatives like the `typing_extensions` module, which we are already using. Besides the maintenance burden, I believe it encourages contributors to keep using and improving the methods defined here instead of analyzing those of `typing_extensions`, making the situation harder to solve.

svlandeg:
@ivantodorovich: my counter proposal, reflected by the current state of this PR:

- Import `get_args` and `get_origin` from `typing_extensions` as you suggested

But, instead of entirely deleting the file:
- Remove most of the dead (Pydantic-specific) code from `_typing.py`
- Keep a few functions that may be relevant to Typer, such as `is_union` that is currently being used on `master`, and things like `is_literal_type` and `literal_values` that would be used by https://github.com/fastapi/typer/pull/669

This will make the file much more manageable in the future, and it'll be easier to refactor functionality further when/if we need it.

ivantodorovich:
Thanks @svlandeg !

That looks great ❤️

tiangolo:
Great, thank you @ivantodorovich! :coffee: 

And thanks @svlandeg for all the work and help here. :bow: :cake:

PR #808, opened 2024-04-25
✨ Support cp850 encoding for auto-completion in PowerShell

svlandeg:
This PR continues from https://github.com/tiangolo/typer/pull/277 by @matejfroebe and focuses on adding support for the `cp850` encoding needed for Powershell on Windows. 

To keep the changes minimal per PR and better understand CI failures, it does not yet introduce the `path.exists` check - I suggest to do this in follow-up work as it would require adding (and potentially changing) some unit tests.


svlandeg (commented):
[maintenance note]: It would be nice to merge in #824 first, and hopefully get a failing test that would have shown this issue earlier!

tiangolo (approved):
Looks good, thank you! :bow: :rocket:

PR #809, opened 2024-04-25
✅ Add `__init__.py` files to fix test suite

svlandeg:
When running the test suite locally (both from the IDE as from Git Bash on Windows), I get a failure on `tests/test_tutorial/test_subcommands/` both for `test_tutorial001.py` as well as `test_tutorial003.py` because the `import items` statement fails. The reason why it fails is because a `__init__.py` file is missing in the original `docs_src` module. Even though the CI here doesn't seem to be impacted, I think we might as well add the proper `__init__.py` files.


svlandeg:
> Even though the CI here doesn't seem to be impacted

In fact, perhaps we should consider extending the test suite to run on a grid, including `windows-latest` and `macos-latest`?

tiangolo:
Awesome! :rocket: 

Yes, it would be great to run tests on Windows and MacOS, would you like to add that? :pray:

PR #810, opened 2024-04-25
✅ Generalize test suite to run on Windows

svlandeg:
Ensure that the test suite runs properly on Windows by:
- adding quotes when paths can have spaces or backslashes
- using `Pathlib` to automatically parse the correct path separator, rather than relying on the literal string `"/"`

Related PR: https://github.com/tiangolo/typer/pull/809


tiangolo:
Awesome, thank you! :rocket:

PR #812, opened 2024-04-27
Contributing and tools

ivantodorovich:
📝 Update the Contributing documentation

* Remove the outdated `flit` section
* Mention pre-commit and how to use it
* Describe how to set-up the development dependencies
* Describe how to run tests

---

⬆ Bump ruff pre-commit hook

This also does a minor formatting diff on some files

---

🤖 Add mypy pre-commit hook

With this, it's not needed to install `mypy` directly, as pre-commit will do it in a separate virtual env

---

🌱 Run lint in a separate ci job 

Before this commit, the linting was done in the test matrix (once per python version). This is not necessary, as the linting is the same for all versions.

Now, the linting is fully delegated to pre-commit, and executed as a separate job.

By doing this we delegate `ruff` completely to `pre-commit`, and so the linting script and the `ruff` requirement for testing is removed.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -4,7 +4,82 @@ First, you might want to see the basic ways to [help Typer and get help](help-ty
 
 If you already cloned the repository and you know that you need to deep dive in the code, here are some guidelines to set up your environment.
 
-### Virtual environment with `venv`
+### Pre-commit hooks
+
+We use <a href="https://pre-commit.com/" class="external-link" target="_blank">pre-commit</a> to run checks and formatting before submitting code for review.
+
+#### Install pre-commit
+
+We recommend <a href="https://pipx.pypa.io/" class="external-link" target="_blank">pipx</a> to install pre-commit in an isolated environment.
+
+=== "pipx"
+
+    <div class="termy">
+
+    ```console
+    $ pipx install pre-commit
+    ```
+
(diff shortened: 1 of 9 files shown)


svlandeg:
Hi @ivantodorovich, thanks for these contributions! To get this reviewed and merged more easily, I suggest to break this PR up into atomic bits, each representing one functionality change or improvement.

docs/contributing.md
```diff
@@ -87,54 +162,18 @@ If it shows the `pip` binary at `env/bin/pip` then it worked. 🎉
...
-After activating the environment as described above, install `flit`:
+With the virtual environment activated, you can install Typer in "editable" mode:
 
 <div class="termy">
 
 ```console
-$ pip install flit
-
----> 100%
+$ pip install --editable .
```
svlandeg, line 172: This does require at least [pip 21.3](https://pip.pypa.io/en/stable/news/#v21-3), as it'll otherwise fail with a note about not finding `setup.py` or `setup.cfg`. So it might be good to add that as a note here as well.

svlandeg:
As most of these edits have gone out-of-date with the current `master` branch, I'll go ahead and close this one. Always happy to reconsider new PRs of course, and ideally let's stick to one atomic change per PR 🙏

PR #814, opened 2024-04-30
🐛 Fix usage of `Annotated` with future annotations in Python 3.7+

ivantodorovich:
Fixes https://github.com/tiangolo/typer/discussions/802

`RuntimeError: Type not yet supported: typing_extensions.Annotated[...]`

The solution is to use the `get_type_hints` method from `typing_extensions` instead of the built-in in the `typing` module, as it includes backports that support properly evaluating the forward reference.


kinuax:
PR Review

- [X] I read the [Review Pull Requests docs](https://typer.tiangolo.com/help-typer/#review-pull-requests).

I checked the changes locally on Linux and all tests pass on 3.8..3.12. I reverted the bugfix, run all tests again on 3.8..3.12 and `test_future_annotations.py` fails on 3.8 as expected:

    RuntimeError: Type not yet supported: typing_extensions.Annotated[bool, <typer.models.OptionInfo object>]

kinuax (approved):

svlandeg (approved):
Thanks for the PR - great catch and fix!

I confirmed locally that the behaviour is broken with Python 3.8 and fixed with this PR.

@ivantodorovich: as a "pro tip", next time you create a PR to resolve a bug, you could consider committing the failing test separately as the first commit. This will run the CI, which will be red, clearly showcasing the problem. Then you'd push your fix, the CI would be rerun and would be green, demonstrating that you've fixed things 😉 

One final comment for Tiangolo: not sure whether you want this new test in the `tests` root or perhaps some place else like `test_compat` - I'll leave it like this for now for you to decide.

ivantodorovich:
> @ivantodorovich: as a "pro tip", next time you create a PR to resolve a bug, you could consider committing the failing test separately as the first commit. This will run the CI, which will be red, clearly showcasing the problem. Then you'd push your fix, the CI would be rerun and would be green, demonstrating that you've fixed things 😉

Good tip! ❤️

tiangolo:
Awesome, thank you @ivantodorovich! :rocket: 

And thanks for the review with all the details @svlandeg! :bow: 

This will be available in Typer `0.12.4` released in the next few hours. :tada:

PR #816, opened 2024-04-30
Fix use of options_metavar

gar1t:
Typer was ignoring this setting when provided to commands. Refer to added test for details.

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -181,7 +181,7 @@ class Typer:
         help: Optional[str] = Default(None),
         epilog: Optional[str] = Default(None),
         short_help: Optional[str] = Default(None),
-        options_metavar: str = Default("[OPTIONS]"),
+        options_metavar: Optional[str] = None,
         add_help_option: bool = Default(True),
         hidden: bool = Default(False),
         deprecated: bool = Default(False),
@@ -202,7 +202,7 @@ class Typer:
                 help=help,
                 epilog=epilog,
                 short_help=short_help,
-                options_metavar=options_metavar,
+                options_metavar=options_metavar or self.info.options_metavar,
                 add_help_option=add_help_option,
                 hidden=hidden,
                 deprecated=deprecated,
@@ -221,7 +221,7 @@ class Typer:
         help: Optional[str] = None,
         epilog: Optional[str] = None,
         short_help: Optional[str] = None,
-        options_metavar: str = "[OPTIONS]",
+        options_metavar: Optional[str] = None,
         add_help_option: bool = True,
         no_args_is_help: bool = False,
         hidden: bool = False,
@@ -242,7 +242,9 @@ class Typer:
                     help=help,
                     epilog=epilog,
                     short_help=short_help,
-                    options_metavar=options_metavar,
+                    options_metavar=(
+                        options_metavar or self._info_val_str("options_metavar")
+                    ),
                     add_help_option=add_help_option,
                     no_args_is_help=no_args_is_help,
                     hidden=hidden,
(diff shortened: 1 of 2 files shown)


typer/main.py
```diff
@@ -181,7 +181,7 @@ def callback(
         help: Optional[str] = Default(None),
         epilog: Optional[str] = Default(None),
         short_help: Optional[str] = Default(None),
-        options_metavar: str = Default("[OPTIONS]"),
+        options_metavar: Optional[str] = None,
```
svlandeg, line 184: Should this not use the `Default` class?
```suggestion
        options_metavar: Optional[str] = Default(None),
```

typer/main.py
```diff
@@ -272,7 +274,7 @@ def add_typer(
...
```
svlandeg, line 277: Should this not use the `Default` class?
```suggestion
        options_metavar: Optional[str] = Default(None),
```

svlandeg (commented):
Thanks! 

I have yet to check the test case, but a few quick first comments:

svlandeg (commented):
I double checked the unit test and found that the second option (`c2`) is already working correctly on `master`, while the first one (`c1`) fails. Nevertheless, it's good to have both options checked here.

One thing I wonder about though. If you change `"OPTIONS"` to `"OPTS"` or something different entirely like `"STUFF"`, the box with options in the help menu would still be named `"Options"`. 

```
$ python script.py --help

 Usage: script.py [STUFF]

┌─ Options ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ --install-completion          Install completion for the current shell.                                                                                                                                                                                                                         │
│ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                                                                                                                                                                  │
│ --help                        Show this message and exit.                                                                                                                                                                                                                                       │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

Is this something that should be addressed as well?

typer/main.py
```diff
@@ -202,7 +202,7 @@ def decorator(f: CommandFunctionType) -> CommandFunctionType:
                 help=help,
                 epilog=epilog,
                 short_help=short_help,
-                options_metavar=options_metavar,
+                options_metavar=options_metavar or self.info.options_metavar,
```
svlandeg, line 205: Shouldn't this use `_info_val_str`?

typer/main.py
```diff
@@ -294,7 +296,7 @@ def add_typer(
...
```
svlandeg, line 299: Shouldn't this use `_info_val_str`?

gar1t:
> One thing I wonder about though. If you change `"OPTIONS"` to `"OPTS"` or something different entirely like `"STUFF"`, the box with options in the help menu would still be named `"Options"`.
> 
> ```
> $ python script.py --help
> 
>  Usage: script.py [STUFF]
> 
> ┌─ Options ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
> │ --install-completion          Install completion for the current shell.                                                                                                                                                                                                                         │
> │ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                                                                                                                                                                  │
> │ --help                        Show this message and exit.                                                                                                                                                                                                                                       │
> └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
> ```
> 
> Is this something that should be addressed as well?

I think that's a good idea!

But the goal of this PR is to fix a bug that currently prevents `options_metavar` from propagating to sub-commands. I believe the current PR accomplishes that and should be merged and closed.

I would classify your point here as an enhancement to the API to support configuration of the panel label/title. This could be captured as a new feature request, while merging this PR.

svlandeg (approved):

PR #824, opened 2024-05-03
✅ Add CI configs to run tests on Windows and MacOS

svlandeg:
Extend test suite:
- Testing on a matrix of different Python and OS versions
- Not running all possible combinations as that seems overkill
- Run the latest Python version (3.11 here) on all three OS's
- Not yet including Python 3.12 - I can update https://github.com/tiangolo/typer/pull/807 separately after merging this.

In total, this runs the CI two more times than before (7 instead of 5). 

## Fixes 

To get the CI to go green for all systems, following edits/fixes were required:

### Ensure we're always reading in UTF-8

`tests/test_tutorial/test_parameter_types/test_file/test_tutorial004(_an)` failed on Windows:
```
AssertionError: assert 'la cig�e�a trae al ni�o' in 'some settings\nla cigüeña trae al niño'
```

To fix it, I set the `encoding` for `read_text` to be `utf8`.

### Skip completion tests for Windows and MacOS

This isn't the cleanest of solutions, but it seems to me that these completion tests were very much written to work on Linux/bash only. In the future, it'd be nice to have MacOS & Windows compatible tests for this - I can make an issue for it to follow up later?

### Use OS in coverage report name
By including the OS in the  `coverage` file, we prevent overwriting / confusion:

image

### Windows encoding issue with `coverage run`

Added an `env` argument to `subprocess.run` calls that were failing for the `test_commands/test_help` `test_script` tests. I'm still not clear why we were getting `UnicodeDecodeError`'s for these. I double checked all files in the `typer` codebase and couldn't find any with a wrong decoding. Something must be happening when calling `coverage run`. Either way, by taking inspiration from [ cpython#105312](https://github.com/python/cpython/issues/105312#issuecomment-1919348455) and [chaiNNer#2815](https://github.com/chaiNNer-org/chaiNNer/pull/2815) it seems to be fixed.

### MacOS issue with Python dir caching

The Python installation is being cached with `actions/cache@v3`, which refers to `env.pythonLocation`. However on macOS, this directory appears to be empty. The "Post Run" log would mention 
> Cache Size: ~0MB (478 B)

and then the next run would appear to "successfully" restore from this cache, but the commands wouldn't actually be available:

image

Cf also a related discussion https://github.com/actions/setup-python/issues/813. I decided to implement the same workaround as https://github.com/microsoft/torchgeo/pull/2024, basically disabling the caching for MacOS.

### Windows issue with Python dir caching

One remaining issue is to investigate why the Windows caching isn't currently working:

image

This is not necessary to get this PR green, so I suggest to keep this as follow-up work.

### Fix combining coverage files from different OS's

When combining the different `.coverage.XXX` files from the different runs in the matrix, `coverage` needs to [remap](https://coverage.readthedocs.io/en/latest/cmd.html#re-mapping-paths) the locations. It's easier to just use the [`relative_files`](https://coverage.readthedocs.io/en/latest/config.html#config-run-relative-files) option as suggested in the Coverage docs.

### DIFF
--- a/tests/test_completion/test_completion.py
+++ b/tests/test_completion/test_completion.py
@@ -3,14 +3,11 @@ import subprocess
 import sys
 from pathlib import Path
 
-import pytest
-
 from docs_src.commands.index import tutorial001 as mod
+from tests.utils import needs_linux
 
 
-@pytest.mark.skipif(
-    not sys.platform.startswith("linux"), reason="Test designed for Linux/bash"
-)
+@needs_linux
 def test_show_completion():
     result = subprocess.run(
         [
@@ -25,9 +22,7 @@ def test_show_completion():
     assert "_TUTORIAL001.PY_COMPLETE=complete_bash" in result.stdout
 
 
-@pytest.mark.skipif(
-    not sys.platform.startswith("linux"), reason="Test designed for Linux/bash"
-)
+@needs_linux
 def test_install_completion():
     bash_completion_path: Path = Path.home() / ".bashrc"
     text = ""
--- a/tests/test_completion/test_completion.py
+++ b/tests/test_completion/test_completion.py
@@ -4,10 +4,13 @@ import sys
 from pathlib import Path
 
 import pytest
+
 from docs_src.commands.index import tutorial001 as mod
 
 
-@pytest.mark.skipif(not sys.platform.startswith("linux"), reason="Test designed for Linux/bash")
+@pytest.mark.skipif(
+    not sys.platform.startswith("linux"), reason="Test designed for Linux/bash"
+)
 def test_show_completion():
     result = subprocess.run(
         [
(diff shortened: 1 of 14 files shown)


svlandeg:
Ok, the Windows & MacOS-specific issues seem to be addressed, but now the coverage report has broken 😭 

> No source for code: '/Users/runner/work/typer/typer/docs_src/arguments/default/tutorial001.py'.

I continue investigating...

tests/test_completion/test_completion.py
```diff
@@ -3,9 +3,14 @@
...
 from pathlib import Path
 
+import pytest
+
 from docs_src.commands.index import tutorial001 as mod
 
 
+@pytest.mark.skipif(
+    not sys.platform.startswith("linux"), reason="Test designed for Linux/bash"
+)
```
tiangolo, line 13: Maybe we could move this and the one below to `tests/utils.py`, there's a `@needs_py310` there, we could have a `@needs_linux` with this that we can re-use in both places.

tiangolo (commented):
Dang, you're good :bow: :sunglasses: 

Impeccable job with all this! :sparkles: 

And all the data and references in the description... :chef_kiss: (we need that chef kiss emoji in GitHub).

---

Just a tiny nitpick, mainly to be annoying. :sweat_smile:

tests/test_completion/test_completion.py
tiangolo: Yep! That's perfect! :rocket:

tiangolo (approved):

tiangolo:
Amazing job, thank you! :bow: :rocket:

PR #828, opened 2024-05-09
✨ Add support for `rich_help_panel` of commands in generated docs

kinuax:
The generated Markdown docs ignore the division given by `rich_help_panel` while listing the commands. This is a proposal to support it.


kinuax:
I added the tests to verify the cases with default (`None`, "", "Commands") and custom panel titles. All format, linting, and tests pass locally on 3.11. CI only breaks on 3.7 with a syntax error due to the walrus operator. Having Python 3.7's end of life almost a year ago I wonder if this a good moment to drop its support, I can take care of that before considering these changes.

kinuax:
As clarified at https://github.com/tiangolo/typer/pull/830, `click` still supports 3.7 so I have refactored to avoid assignment expressions and pass 3.7 tests. Tests are updated too to verify mixed commands in same app and pass coverage. Ready to be reviewed.

svlandeg:
Tricky merge conflict with https://github.com/fastapi/typer/pull/944, which has changed the order of commands being displayed. Should be all good and green again.

kinuax:
@svlandeg, many thanks for checking this out and fixing the merge conflicts! :pray:

svlandeg (commented):
I think this is a nice feature! I tested it with 
```
typer docs_src/commands/help/tutorial006.py utils docs
```
to see the difference between `master` and this PR, and you nicely get the subdivision in the markdown with this PR: 
image

whereas on `master` it's just:

image

There's a few issues with the code though. I'll comment inline, and will put the PR in draft as I push some changes.

typer/cli.py
```diff
@@ -1,13 +1,16 @@
...
+from itertools import chain
 from pathlib import Path
-from typing import Any, List, Optional
+from typing import Any, Dict, List, Optional
 
 import click
 import typer
 import typer.core
 from click import Command, Group, Option
+from typer.rich_utils import _RICH_HELP_PANEL_NAME, COMMANDS_PANEL_TITLE
```
svlandeg: There's two issues here:
- this will simply crash when `rich` is not installed
- even if `rich` is installed, we want to [lazy-load](https://github.com/fastapi/typer/pull/1128) it

I will have a look into fixing this.

typer/rich_utils.py
```diff
@@ -592,14 +592,10 @@ def rich_format_help(
         if getattr(param, "hidden", False):
             continue
         if isinstance(param, click.Argument):
-            panel_name = (
-                getattr(param, _RICH_HELP_PANEL_NAME, None) or ARGUMENTS_PANEL_TITLE
-            )
+            panel_name = get_panel_name(param, ARGUMENTS_PANEL_TITLE)
```
svlandeg, line 578: These three edits at L595, L598 and L644 can, in principle, be removed from the PR. I added them because the newly created internal function `get_panel_name` simplifies the code IMO.

svlandeg (approved):
I think this is ready for review by Tiangolo.

One final remark I have here is that this code is not tested in our test suite without having `rich` installed. This is why the initial implementation didn't result in a test failure, while it really should have. https://github.com/fastapi/typer/pull/1297 should improve that situation (that said, I think these PRs can be reviewed/merged independently).

typer/rich_utils.py
kinuax: Agree, thanks!

svlandeg (commented):
Hmmm, something seems to have been broken after merging `master`, maybe there is a merge conflict that wasn't handled correctly, or there is an interdependency with another PR.  `test_doc_no_rich` doesn't show the "Greet" and "Farewell" headers anymore, though `test_doc_title_output` does 🤔 I'll continue to investigate.

UPDATE: wait, so actually isn't this correct? The option is called `rich_help_panel`, so it shouldn't affect non-Rich output, right? I'm going to update the PR to reflect this.

tests/assets/cli/multiapp-docs-norich.md
```diff
@@ -0,0 +1,102 @@
+# `multiapp`
```
svlandeg, line 1: This is basically the same file as what `multiapp-docs.md` was before, without the "Greet" and "Farewell" headers.

tests/test_cli/test_doc.py
```diff
@@ -138,7 +140,7 @@ def test_doc_no_typer():
             "run",
             "-m",
             "typer",
-            "tests/assets/cli/empty_script.py",
+            "tests.assets.cli.empty_script",
```
svlandeg, line 143: drive-by edit that is not necessary for this PR 🤷

kinuax:
@svlandeg thanks for fixing the tests! :pray:

PR #830, opened 2024-05-15
🔥 Drop support for Python 3.7

kinuax:
CPython and click already dropped support for Python 3.7. I would propose to proceed accordingly here in order to benefit from:
- reducing the CI tests
- utilizing improved types without branching
- making available the syntax provided by 3.8:
    - assignment expressions
    - positional-only arguments
    - f-strings with =

3.7 occurrences are updated. `Literal` related branching is cleaned. `_typing` is updated and simplified. All local tests pass on 3.8-3.12 and Linux.

Context: https://github.com/tiangolo/typer/pull/828


kinuax:
CI tests pass.

.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         python-version: [ "3.11" ]
         include:
           - os: ubuntu-latest
-            python-version: "3.7"
+            python-version: "3.8"
```
svlandeg, line 23: In fact, L22 and L23 can just be deleted. The CI currently works so that each older Python version is tested on one OS, and the latest Python is tested on all three OS's. Python 3.8 is already being tested on L24 and L25, so these lines could just be removed.

svlandeg (commented):
Thanks for the very thorough PR!

I'll have to ask @tiangolo what the term is he wants to drop Python 3.7 by.

Note - I'm also maintaining spaCy and there we just [bumped](https://github.com/explosion/spaCy/pull/13471) the Typer pin to allow any v0.x. If Typer drops Python 3.7 which spaCy v3 still supports, we'll have to again restrict the upper bound for Typer there. And I can imagine spaCy not being the only repo that would run into this. 

I think we'll want to drop 3.7 at some point, it's just a question of when :-)

.github/workflows/test.yml
kinuax: You are right. Thanks for reviewing the changes and clarifying the test procedure with older Python versions. :)

kinuax:
@svlandeg, thanks for reviewing the PR and for sharing your thoughts. :+1: 

I understand the concern about avoiding breaking dependent libraries.

Besides, I rechecked the `click` repo and I would like to inform (and also correct previous statements) that, although old, the specific change that drops 3.7 is not released yet (it was introduced in `main` branch just after the latest release 8.1.7).
This means that, in my opinion, the consideration of dropping support in `Typer` would happen, at least, after a new  `click` release takes place.

typer/_typing.py
```diff
@@ -448,25 +367,16 @@ def is_callable_type(type_: Type[Any]) -> bool:
     return type_ is Callable or get_origin(type_) is Callable
 
 
-if sys.version_info >= (3, 7):
```
killjoy1221, line 451: I think you inverted this condition. The else block applies if 3.6 was being used.

tiangolo:
Thanks @kinuax! I want to keep support for Python 3.7 for a while. It's currently not being a problem (yet) so I prefer to delay this a bit.

I would also want to have some features and fixes in before dropping support, so that people stuck in 3.7 can benefit from them too.

typer/_typing.py
kinuax: Right, good catch, thanks!

kinuax:
@tiangolo, all right, I understand, thanks for clarifying.

kinuax:
Merge conflicts are solved and tests are passing. Ready to be reviewed.

kinuax:
Latest merge conflicts are solved.

kinuax:
Recent `Click` release 8.2.0 drops support for Python 3.7.

svlandeg (approved):
All our other repos (`asyncer`, `fastapi`, `sqlmodel`) have now also dropped support for Python 3.7, as has Click. So I think we should go ahead and merge this to avoid issues on other PRs.

@kinuax: thanks for your continued work on this! Much appreciated 🙏

tiangolo (approved):
The time has come, let's do this! 🚀 🔥

Issue #819, opened 2024-05-02
Rich markup in generated Markdown docs


tiangolo:
### Discussed in https://github.com/tiangolo/typer/discussions/818

<div type='discussions-op-text'>

<sup>Originally posted by **tiangolo** May  2, 2024</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorials in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer
from typing_extensions import Annotated

app = typer.Typer(rich_markup_mode="rich")


@app.command()
def hello(
    name: Annotated[
        str, typer.Option(help="The [bold]cool[/bold] name of the [green]user[/green]")
    ],
):
    print(f"Hello {name}")


if __name__ == "__main__":
    app()
```

### Description

* Create this script in `main.py`
* Generate a README with `typer main.py utils docs --output demo.md`

The output is:

````Markdown
# `hello`

**Usage**:

```console
$ hello [OPTIONS]
```

**Options**:

* `--name TEXT`: The [bold]cool[/bold] name of the [green]user[/green]  [required]
* `--install-completion`: Install completion for the current shell.
* `--show-completion`: Show completion for the current shell, to copy it or customize the installation.
* `--help`: Show this message and exit.
````

It includes the sections with `[green[user[/green]` verbatim, those are rendered by Rich in the terminal, but not in the generated Markdown, they look like this:

<details>
<summary>Rendered Markdown</summary>

# `hello`

**Usage**:

```console
$ hello [OPTIONS]
```

**Options**:

* `--name TEXT`: The [bold]cool[/bold] name of the [green]user[/green]  [required]
* `--install-completion`: Install completion for the current shell.
* `--show-completion`: Show completion for the current shell, to copy it or customize the installation.
* `--help`: Show this message and exit.

</details>

It might be possible to generate the HTML that would be rendered by Markdown for those sections with Rich's [`export_html()`](https://rich.readthedocs.io/en/stable/reference/console.html#rich.console.Console.export_html) or maybe `export_text()`. Not sure but might work.

### Operating System

Linux, Windows, macOS

### Operating System Details

[...]

PR #847, opened 2024-05-28
✨ Render Rich markup as HTML in Markdown docs

svlandeg:
Fixes https://github.com/tiangolo/typer/issues/819: ensure that Rich-formatted help texts get rendered properly in Markdown by using the Rich built-in `export_html` functionality.

I've expanded the example from #819 to ensure that the following are covered: Command help, Option help, Argument help, default value, `[required]` flag.

## TODO

Double check whether existing functionality and printing behaviour has remained unchanged.


typer/core.py
```diff
@@ -360,8 +360,13 @@ def get_help_record(self, ctx: click.Context) -> Optional[Tuple[str, str]]:
         if self.required:
             extra.append(_("required"))
         if extra:
-            extra_str = ";".join(extra)
-            help = f"{help}  [{extra_str}]" if help else f"[{extra_str}]"
+            extra_str = "; ".join(extra)
+            extra_str = f"[{extra_str}]"
+            if rich is not None:
+                # This is needed for when we want to export to HTML
+                extra_str = rich.markup.escape(extra_str).strip()
```
svlandeg, line 367: This is the only bit that worries me a bit (here and L559) as this changes core functionality anytime `get_help_record` is called on an argument or an option. Existing tests initially failed in `test_doc.py` but I was able to remedy those by using `highlight=False`, `overflow="ignore"` and `crop=False` in the `rich_to_html` helper function.

Nevertheless @tiangolo, it would be good to get your eyes on this 🙏

tiangolo: Thank you! Yep, just checked and it's all good! :rocket:

tiangolo (approved):
Great, thank you! :rocket: :bow:

tiangolo:
Just for completeness, extra info:

I'm not sure what changed, maybe it was something in a recent release in Rich, because tests were passing here before merging, but then they were no longer passing :sweat_smile: 

I changed it here: https://github.com/fastapi/typer/pull/1055, it was about how to capture and render the Rich console.

seowalex:
This seems to have resulted in extra backspaces in some cases: https://github.com/fastapi/typer/discussions/1058.

PR #850, opened 2024-06-03
🔥 Remove Python 3.6 specific code paths

svlandeg:
Support for Python 3.6 was [dropped](https://github.com/tiangolo/typer/pull/758) recently. This PR removes some code in `_typing.py` that has become redundant since then.


tiangolo:
Awesome, very nice cleanup! Thank you @svlandeg :bow: :cake:

PR #859, opened 2024-06-12
🐛 Ensure `rich_markup_mode=None` disables Rich formatting

svlandeg:
This PR is an alternative to #726.

First, it fixes the bug described in https://github.com/tiangolo/typer/discussions/853, namely that
```
cli = typer.Typer(rich_markup_mode=None)
```

does not actually disable Rich formatting on the current `master`, when Rich is installed. This is because many of the methods in `core.py` only check whether `rich` is available before calling the functions in `rich_utils.py`. Instead, the value of `rich_markup_mode` should be checked as well, as in in PR #726 and as in this PR.

However, when we do that, we effectively make `None` the default value, even when Rich is installed. This aligns with the current [documentation](https://typer.tiangolo.com/tutorial/commands/help/#rich-markdown-and-markup):

> By default, rich_markup_mode is None, which disables any rich text formatting.

However, one could argue that this behaviour is a bit unexpected, and that Rich formatting should be the default when Rich is installed. This is what the test suite currently assumes on `master`, and attempting to change the test suite (as in #726) shows that there would be many breaking cases for users as well.

So, as an alternative to #726 I suggest to fix the original bug, while keeping the default behaviour as it was: Rich formatting if Rich is installed, no formatting if it isn't installed. By introducing a new var `DEFAULT_MARKUP_MODE` this behaviour is formalized. This will also pave the way to potentially expending this functionality and allowing the user to set this default globally, as proposed in #647. For now, I would focus first on getting this bug resolved and deciding what the defaults should look like.

If we do decide to merge this PR, we can close #726 (or vice versa).


tests/test_rich_markup_mode.py
```diff
@@ -0,0 +1,40 @@
+import typer
+import typer.completion
+from typer.testing import CliRunner
+
+runner = CliRunner()
+rounded = ["╭", "─", "┬", "╮", "│", "├", "┼", "┤", "╰", "┴", "╯"]
+
+
+def test_rich_markup_mode_none():
```
svlandeg, line 9: This test currently fails on `master`, which is clearly a bug IMO.

JacobKochems:
> So, as an alternative to #726 I suggest to fix the original bug, while keeping the default behaviour as it was: Rich formatting if Rich is installed, no formatting if it isn't installed. By introducing a new var `DEFAULT_MARKUP_MODE` this behaviour is formalized. This will also pave the way to potentially expending this functionality and allowing the user to set this default globally, as proposed in #647. For now, I would focus first on getting this bug resolved and deciding what the defaults should look like.

I second the idea of keeping the default behaviour as a function of Rich's installation status. I think the easiest way towards the user defined global default is to close #647 and merge this bugfix. After which I would take a shot at reimplementing #647 on top of this fix.

JacobKochems (approved):
I cloned this branch, reviewed the test
```
tests/test_rich_markup_mode.py
```
and ran it locally on my machine. Both test cases pass.
The documentation change looks good to me, too.

moskupols:
This would really help in my project. I like having a canonization test for --help of my utilities. Currently such a test has to depend on whether rich is installed, and with rich installed the output is full of unicode characters. I'd prefer enforcing the plain help for this test.

tiangolo:
Awesome, thank you @svlandeg! Thanks for all the clear explanations, clean code, and help with everything. :bow: :rocket: 

tests/test_rich_markup_mode.py
kraj: yes it fails for 0.12.5 release here as well
```
def test_rich_markup_mode_rich():
        app = typer.Typer(rich_markup_mode="rich")

        @app.command()
        def main(arg: str):
            """Main function"""
            print(f"Hello {arg}")

        assert app.rich_markup_mode == "rich"

        result = runner.invoke(app, ["World"])
        assert "Hello World" in result.stdout

        result = runner.invoke(app, ["--help"])
>       assert any(c in result.stdout for c in rounded)
E       assert False
E        +  where False = any(<generator object test_rich_markup_mode_rich.<locals>.<genexpr> at 0x7f931ae59d80>)

tests/test_rich_markup_mode.py:40: AssertionError
```

tests/test_rich_markup_mode.py
```diff
@@ -0,0 +1,40 @@
...
+    assert app.rich_markup_mode is None
+
+    result = runner.invoke(app, ["World"])
+    assert "Hello World" in result.stdout
+
+    result = runner.invoke(app, ["--help"])
+    assert all(c not in result.stdout for c in rounded)
+
+
+def test_rich_markup_mode_rich():
```
kraj, line 26: This test fails with 0.12.5 release
```
def test_rich_markup_mode_rich():
        app = typer.Typer(rich_markup_mode="rich")

        @app.command()
        def main(arg: str):
            """Main function"""
            print(f"Hello {arg}")

        assert app.rich_markup_mode == "rich"

        result = runner.invoke(app, ["World"])
        assert "Hello World" in result.stdout

        result = runner.invoke(app, ["--help"])
>       assert any(c in result.stdout for c in rounded)
E       assert False
E        +  where False = any(<generator object test_rich_markup_mode_rich.<locals>.<genexpr> at 0x7f931ae59d80>)

tests/test_rich_markup_mode.py:40: AssertionError
```

svlandeg: Hi @kraj, thanks for the comment. That's odd - I can't reproduce the test failure locally and the CI goes green with this change as well.

Could you please open a new [discussion thread](https://github.com/fastapi/typer/discussions) to look into this more, and provide us there with more information about your system setup and the failure you're seeing? 🙏

kraj: I build a minimal image which just have a kernel+busybox+typer and its dependencies. So perhaps thats the difference, some implicit dependency coming in indirectly on full distributions like ubuntu, fedora etc in CI

PR #863, opened 2024-06-16
🩹 Add fix for typer --help.

rokbot:
Add mising run command on:
$ typer --help


tiangolo:
Thanks for the interest @rokbot, but indeed, this is as intended. The `run` command won't be available unless there's an actual app to run.

Also, completion of the script depends on the dynamic `run` command being created. If the interface was:

```console
$ typer run ./some_script.py
```

it wouldn't be possible to provide shell completion for the commands and CLI parameters defined in `./some_script.py`.

Given that, I'll pass on this, but thanks! :coffee:

PR #866, opened 2024-06-17
✅ Fix tests for completion flags

rokbot:
--show-completions
zsh, fish, pwsh users could run tests with their default shell

--install-completions
(WIP: missing fish, pwsh)
this test is skipped due to side-effects

### DIFF
--- a/tests/test_completion/test_completion.py
+++ b/tests/test_completion/test_completion.py
@@ -5,7 +5,7 @@ from pathlib import Path
 
 from docs_src.commands.index import tutorial001 as mod
 
-from ..utils import needs_linux, needs_cleaning
+from ..utils import needs_cleaning, needs_linux
 
 
 @needs_linux
--- a/tests/test_completion/test_completion.py
+++ b/tests/test_completion/test_completion.py
@@ -5,44 +5,73 @@ from pathlib import Path
 
 from docs_src.commands.index import tutorial001 as mod
 
-from ..utils import needs_linux
+from ..utils import needs_linux, needs_cleaning
 
 
 @needs_linux
-def test_show_completion():
+def test_show_completion_flag():
     result = subprocess.run(
         [
             "bash",
             "-c",
-            f"'{sys.executable}' -m coverage run '{mod.__file__}' --show-completion",
+            "typer --show-completion",
         ],
         capture_output=True,
+        text=True,
         encoding="utf-8",
-        env={**os.environ, "SHELL": "/bin/bash", "_TYPER_COMPLETE_TESTING": "True"},
     )
-    assert "_TUTORIAL001.PY_COMPLETE=complete_bash" in result.stdout
+    curr_shell_path = os.environ["SHELL"]
+    # check for default user shell
+    shell = "bash"
+    if "zsh" in curr_shell_path:
+        shell = "zsh"
+    if "fish" in curr_shell_path:
+        shell = "fish"
+    if "pwsh" in curr_shell_path:
+        shell = "pwsh"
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Hi!

Thanks for your contribution! Could you explain in a bit more detail the changes you've made to the tests, and why?

I also noticed that the tests are currently failing, so I put this PR in draft node. Do you have time to try and get these fixed so that the test suite is green again? This will make the review process go more smoothly.

svlandeg:
> the SHELL environment variable make this test to fail when the current shell isn't bash.

You're right - the `@needs_linux` tag is not perfectly suited because a different shell can be used on Linux, and vice versa this test succeeds in Bash on Windows. So, what I suggest we do here instead is to write a new tag, something like `@needs_bash`, and decorate these two unit tests with that instead. Then we don't have to touch anything else in the unit test framework - in general I'm hesitant to remove/reduce tests, so I'd prefer a minimal change instead.

svlandeg:
> in this particularly case, these unit tests only were executed "on linux", now will be "needs bash" to me this is just a slow deprecation cycle that leads to "just delete this tests" at the end

Just to provide a bit more background: the `needs_linux` or `needs_bash` were not introduced as some sort of "slow deprecation". Instead, they were introduced to keep them specifically when we [extended](https://github.com/tiangolo/typer/pull/824) our CI to include Windows and MacOS testing. At the time, I noted that follow-up work could consider extending these tests to be less shell/OS-specific. It looks like you attempted to do this, but the test suite was still failing, probably because you didn't adjust the `subprocess.run` call that executes `"bash"`.

Please don't get me wrong - I really appreciate the effort you've made into updating these tests and explaining why. It's just that from a maintainer's point of view, it's much easier to review atomic changes instead of 3 edits done for 3 different reasons - especially when the final PR is not passing the CI.

So, to ensure we follow up on this work properly and to capture some of the arguments you've made here, I started a new PR to first address the `needs_bash` change which is straightforward.

>  I'm more worried about the side effects that the test suite leaves after executing it, to me the "tester" system must not be affected by the test suite, like the --install-completion tests, that make changes to your system and not clean after finish when you execute it.

I definitely agree that we should look into the "side effects" caused by the `test_install_completion()` test - good point. I'll look into it and will likely open another PR later on.

Thanks again! 🙏

svlandeg:
@rokbot: I looked into this more. Are you sure the current implementation of `test_install_completion` causes side effects? It has a line 
```
bash_completion_path.write_text(text)
```
which should restore the original bash profile, so in fact it shouldn't have any side effects as far as I can see.

svlandeg:
In the meantime, https://github.com/fastapi/typer/pull/888 has been merged, which adds the `needs_bash` functionality. I suggest that we close this PR and optionally follow up with atomic, separate PRs to address one issue at a time (and to ensure that the test suite succeeds for each change). Thanks again for the good discussion!

tiangolo:
Thanks for the interest @rokbot! :coffee: 

And thanks for the help @svlandeg! :bow: Sofie, maybe we could have the "install completion" tests run only under some opt-in env var or something like that, that we could enable by default on CI, or we could enable manually when testing locally, but wouldn't affect others that come and try to run the tests. What do you think? :thinking:  Anyway, it's also not urgent, but we can think if we can find a way to improve that process.

---

I'll close this PR now and we can continue later in subsequent PRs. :cake:

svlandeg:
> Sofie, maybe we could have the "install completion" tests run only under some opt-in env var or something like that, that we could enable by default on CI

Yea, that's a good idea! I'll look into it when I have some time in the next few weeks.

PR #870, opened 2024-06-24
Support ZDOTDIR in zsh completion

chrishepner:
zsh supports placing its startup files in a directory other than $HOME, [when the user specifies $ZDOTDIR in ~/.zshenv](https://zsh.sourceforge.io/Intro/intro_3.html).

typer currently always writes to `~/.zshrc` and `~/.zfunc`. This means that for users that have $ZDOTDIR set to something other than $HOME, `typer --install-completion` has no functional effect on the user's shell.

Update `install_zsh` to install completion to `$ZDOTDIR/.zshrc` and `$ZDOTDIR/.zfunc/` when set.

### DIFF
--- a/tests/test_completion/test_completion_install.py
+++ b/tests/test_completion/test_completion_install.py
@@ -1,9 +1,11 @@
 import os
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 from unittest import mock
 
+import pytest
 import shellingham
 import typer
 from typer.testing import CliRunner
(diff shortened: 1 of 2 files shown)


chrishepner:
Hmm, it looks like the absolute `fpath` breaks the test on Windows. Is zsh completion meaningful on Windows? I'm not sure if this is a meaningful regression or something I should suppress.

svlandeg (commented):
Hi, thanks for the PR!

I believe there are ways to use `zsh` on Windows with `Git Bash`, though I have no experience with it myself. Either way, the test was working before so it feels like we should be able to get this simple check working on Windows as well.

I double checked on a Windows system to see what was going on, and it seems like the `subprocess.run` call isn't actually executed properly. The content of `new_text` is still `echo "custom .zshrc"` when the assert fails.  So I don't think the issue is with `fpath`.

`result.stderr` contains the following:
> Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python

Which seems to point at an issue with the runtime environment, probably due to the way how you've used `mock.patch.dict` to make changes to `os.environ`. Can you think of a different way to run this check?

PR #872, opened 2024-06-27
🐛 Ensure that the value of `is_flag` is passed correctly to Click

iurisilvio:
Discussion in https://github.com/tiangolo/typer/discussions/873.

Click supports [optional value](https://click.palletsprojects.com/en/8.1.x/options/#optional-value).

To make it work, you have to set `is_flag=False, flag_value="any value"`, then:

* `--foo` return defined `flag_value`
* `--foo something` return `something`

It was broken because `is_flag` was passed as `None` when `parameter_info.is_flag` was explicitly `False`.


svlandeg (approved):
Thanks for this contribution! The discussion thread you opened in #873 clearly highlights the issue and the fact that the current Typer behaviour is inconsistent with how Click does it. Typer doesn't have any documentation around these specific parameters so it's fair to assume it would/should behave in the same way as Click does.

I could confirm the erratic behaviour and failing unit test on `master`, and with this PR the issues are resolved. The unit test perfectly captures the problems. The fix is minimal and accurate.

As such, I think this PR looks good to merge. Thanks again! 🙏

svlandeg:
Maybe one point to consider when Tiangolo reviews this, is the lack of Typer documentation for these settings. Perhaps we can add a separate page somewhere, explaining how these arguments work and relate to eachother. We could potentially do this in follow-up work though.

tiangolo:
Thanks @iurisilvio!

I realized it didn't really make sense to me to have this option when it doesn't fit well with regular type annotations, I think it's better to have a simpler syntax for a boolean flag, and have it just be `True` or `False`. So chatting with @svlandeg, the conclusion was to drop this parameter completely to have one main way to achieve these things.

This was done in https://github.com/fastapi/typer/pull/987

Given that, I'll now close this one, but thanks for the effort! :coffee:

alanwilter:
Heavens, please show an example of how to use your approach and not the `click` one!

I need to handle --my_arg whose default is 'hello':

```sh
app
# my_arg is None or False
app --my_arg
# my_arg = hello
app --my_arg bye
# my_arg = bye
```

PR #888, opened 2024-07-24
✅ Add `needs_bash` test fixture

svlandeg:
When we [generalized](https://github.com/tiangolo/typer/pull/824) the test suite to also run on Windows and Mac, we introduced a fixture `needs_linux` to only run two specific tests on Bash/Linux. However, these specific tests may fail on a Linux system with a different shell. So, this PR adds the fixture to `needs_bash` as well. Only keeping `needs_bash` is not sufficient as MacOS tests will fail on the CI.

Thanks to @rokbot for pointing this out in https://github.com/tiangolo/typer/pull/866.

In follow-up work, ideally we should extend these tests to make them more applicable across OS and shell versions, but I want to get this quick fix in first to avoid having the test suite fail on non-bash Linux systems.


tiangolo:
Great, thank you @svlandeg! :rocket:

PR #891, opened 2024-07-26
♻️ Use F-strings in Click examples in docs

svlandeg:
As discussed in https://github.com/tiangolo/typer/pull/502#issuecomment-2016506065:

> This PR covers too many different things, so it would be better to split it to be able to take some things while not taking others.

> Specifically, I would take the changes in the Click examples from "some str {blah}".format(blah=blah) to f-strings. That was just an exact copy of Click's tutorials, but now they updated them, we can update them here as well.

Thanks to @marksmayo for originally catching these cases 🙏


svlandeg:
Note that if we end up merging #938, we can simply close this one. Or we can merge this one first, and update the other one afterwards 🤷.

tiangolo:
Nice, thank you! :rocket: :cake:

PR #944, opened 2024-08-21
✨ Show help items in order of definition

### LINKED ISSUE #933: Show help items in order of definition, not alphabetically
### Privileged issue

### Issue Content

Show help items in order of definition, not alphabetically

For example:

```Python
import typer

app = typer.Typer()


@app.command()
def first(word: str | None = None):
    """
    First declared command should show up first.
    """
    print(f"first {word}")


@app.command()
def after(word: str | None = None):
    """
    Last declared command should show up last, not alphabetically.
    """
    print(f"after {word}")
```

This is the current output:

```console
$ typer demo.py run --help
                                                                            
 Usage: typer [PATH_OR_MODULE] run [OPTIONS] COMMAND [ARGS]...              
                                                                            
 Run the provided Typer app.                                                
                                                                            
╭─ Options ────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                              │
╰──────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────╮
│ after   Last declared command should show up last, not alphabetically.   │
│ first   First declared command should show up first.                     │
╰──────────────────────────────────────────────────────────────────────────╯
```

This is the ideal output:

```console
$ typer demo.py run --help
                                                                            
 Usage: typer [PATH_OR_MODULE] run [OPTIONS] COMMAND [ARGS]...              
                                                                            
 Run the provided Typer app.                                                
                                                                            
╭─ Options ────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                              │
╰──────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────╮
│ first   First declared command should show up first.                     │
│ after   Last declared command should show up last, not alphabetically.   │
╰──────────────────────────────────────────────────────────────────────────╯
```

svlandeg:
Fixes #933 

Click's `Group` class specifically [orders](https://github.com/pallets/click/blob/main/src/click/core.py#L1684) the items alphabetically:
```
    def list_commands(self, ctx: Context) -> list[str]:
        """Returns a list of subcommand names in the order they should appear."""
        return sorted(self.commands)
```

For `TyperGroup`, we override this behaviour and remove the `sorted` function. This works, because during creation of the group, the commands are kept in a list as they get declared, so `self.commands` has the original order.

### DIFF
--- a/docs/tutorial/commands/index.md
+++ b/docs/tutorial/commands/index.md
@@ -257,6 +257,39 @@ Commands:
 
 </div>
 
+
+## Sorting of the commands
+
+Note that by design, **Typer** shows the commands in the order they've been declared.
+
+So, if we take our original example, with `create` and `delete` commands, and reverse the order in the Python file:
+
+```Python hl_lines="7  12"
+{!../docs_src/commands/index/tutorial004.py!}
+```
+
+Then we will see the `delete` command first in the help output:
+
+<div class="termy">
+
+```console
+// Check the help
+$ python main.py --help
+
+Usage: main.py [OPTIONS] COMMAND [ARGS]...
+
+Options:
+  --install-completion  Install completion for the current shell.
+  --show-completion     Show completion for the current shell, to copy it or customize the installation.
+  --help                Show this message and exit.
+
+Commands:
+  delete
+  create
+```
+
+</div>
+
 ## Click Group
 
 If you come from Click, a `typer.Typer` app with subcommands is more or less the equivalent of a <a href="https://click.palletsprojects.com/en/7.x/quickstart/#nesting-commands" class="external-link" target="_blank">Click Group</a>.
--- a/typer/core.py
+++ b/typer/core.py
@@ -739,8 +739,8 @@ class TyperGroup(click.core.Group):
             markup_mode=self.rich_markup_mode,
         )
 
-    def list_commands(self, ctx: click.Context) -> list[str]:
+    def list_commands(self, ctx: click.Context) -> List[str]:
         """Returns a list of subcommand names.
-         Note that in Click's Group class, these are sorted.
-         In Typer, we wish to maintain the original order of creation (cf Issue #933)"""
-        return self.commands
+        Note that in Click's Group class, these are sorted.
+        In Typer, we wish to maintain the original order of creation (cf Issue #933)"""
+        return [n for n, c in self.commands.items()]
--- a/typer/core.py
+++ b/typer/core.py
@@ -738,3 +738,9 @@ class TyperGroup(click.core.Group):
             ctx=ctx,
             markup_mode=self.rich_markup_mode,
         )
+
+    def list_commands(self, ctx: click.Context) -> list[str]:
+        """Returns a list of subcommand names.
+         Note that in Click's Group class, these are sorted.
+         In Typer, we wish to maintain the original order of creation (cf Issue #933)"""
+        return self.commands
--- a/docs_src/commands/index/tutorial004.py
+++ b/docs_src/commands/index/tutorial004.py
@@ -1,0 +1,17 @@
+import typer
+
+app = typer.Typer()
+
+
+@app.command()
+def delete():
+    print("Deleting user: Hiro Hamada")
+
+
+@app.command()
+def create():
+    print("Creating user: Hiro Hamada")
+
+
+if __name__ == "__main__":
+    app()
--- a/tests/assets/cli/multiapp-docs-title.md
+++ b/tests/assets/cli/multiapp-docs-title.md
@@ -18,8 +18,22 @@ The end
 
 **Commands**:
 
-* `sub`
 * `top`: Top command
+* `sub`
+
+## `multiapp top`
+
+Top command
+
+**Usage**:
+
+```console
+$ multiapp top [OPTIONS]
+```
+
+**Options**:
+
+* `--help`: Show this message and exit.
 
 ## `multiapp sub`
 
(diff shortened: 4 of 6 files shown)


typer/core.py
```diff
@@ -738,3 +738,9 @@ def format_help(self, ctx: click.Context, formatter: click.HelpFormatter) -> Non
             ctx=ctx,
             markup_mode=self.rich_markup_mode,
         )
+
+    def list_commands(self, ctx: click.Context) -> List[str]:
+        """Returns a list of subcommand names.
+        Note that in Click's Group class, these are sorted.
+        In Typer, we wish to maintain the original order of creation (cf Issue #933)"""
+        return [n for n, c in self.commands.items()]
```
svlandeg, line 746: Note that we can't just 
```
return sorted(self.commands)
```
here as `self.commands` is actually a `MutableMapping[str, Command]` and so this wouldn't pass type checks.

svlandeg:
`test_doc()` is failing because of two reasons:

- With this PR, "hello", "hi" and "bye" should stay in that order -> that part is correct and we should change the test accordingly
- With this PR, "top" is being documented before "sub". I'm unsure about this particular case, because the definition in `multi_app.py` is like this:

```
sub_app = typer.Typer()

@sub_app.command()
def hello(...):
    ...

app = typer.Typer(help="Demo App", epilog="The end")
app.add_typer(sub_app, name="sub")


@app.command()
def top():
...
```
As such, I'm unsure why "top" is now being shown in the help first.

I'll investigate further and will keep this in draft while doing so.

svlandeg:
Ok, so we need to consider how subgroups are treated here versus commands.

Assume the following simplified code:
```
import typer


app = typer.Typer(help="Demo App")
app.add_typer(typer.Typer(), name="sub")


@app.command()
def top():
    """
    Top command
    """
    typer.echo("top")
```
Then run
```
typer example.py run --help
```
and you'll get
```
┌─ Commands ──────────────────────────────────────────────────────────────────┐
│ top   Top command                                                           │
│ sub                                                                         │
└─────────────────────────────────────────────────────────────────────────────┘
```
So even though `sub` is added to the `app` first, we still see `top` on top. On `master`, `sub` would be first just because it's alphabetically first.

Now, why is this? Because of the [code](https://github.com/fastapi/typer/blob/master/typer/main.py#L475) in `get_group_from_info()` in `main.py`, which first processes all registered commands, and adds them to the list of `commands`, and only in a next loop processes the registered groups and adds those to `commands` as well:
```
    commands: Dict[str, click.Command] = {}
    for command_info in group_info.typer_instance.registered_commands:
        command = ...
        if command.name:
            commands[command.name] = command
    for sub_group_info in group_info.typer_instance.registered_groups:
        sub_group = ...
        if sub_group.name:
            commands[sub_group.name] = sub_group
```

So @tiangolo: are we happy with registered commands taking precedence of registered groups, with this PR? If so, we can have this PR as is and the changes to `tests/assets/cli/multiapp-docs-title.md` (specifically the top/sub switch) reflect that.

If we want to consider the order in which groups vs commands are declared in the input source (a.k.a. keep "sub" as first in the example above), then we'll likely looking at a bigger refactor because we'll have to rethink how `group_info` stores its data.

docs/tutorial/commands/index.md
```diff
@@ -257,6 +257,39 @@ Commands:
 
 </div>
 
+
+## Sorting of the commands
+
+Note that by design, **Typer** shows the commands in the order they've been declared.
```
svlandeg, line 263: Note: if we keep the current behavour of command/group sorting, then we probably want to comment on that in the docs as well, because otherwise it may be confusing. I can add this after we decide how to move forward.

tiangolo (approved):
This looks great! :rocket: :tada: 

@svlandeg feel free to merge it. :sunglasses:

liquidcarbon:
sorry, does this work?  I'm seeing sorted commands.

```python
import typer

print(typer.__version__)

app = typer.Typer(no_args_is_help=True)


@app.command()
def say(what: str):
    """Say it"""
    typer.echo(f"say {what}")


@app.command()
def hi():
    """Say hi to pup"""
    typer.echo("🐶 says: woof!")


if __name__ == "__main__":
    app()
```

Output:
```
pixi run python typer_sort.py
0.12.5

 Usage: typer_sort.py [OPTIONS] COMMAND [ARGS]...

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --install-completion          Install completion for the current shell.                                                                                                          │
│ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                                                   │
│ --help                        Show this message and exit.                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ hi    Say hi to pup                                                                                                                                                              │
│ say   Say it                                                                                                                                                                     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

svlandeg:
Hi @liquidcarbon,

This PR was merged on Aug 26, while v. 0.12.5 was released on Aug 24. So this PR hasn't yet made it to the latest release of `typer`.

svlandeg:
This is now available in release 0.13.0 🎉

PR #964, opened 2024-09-02
✅ Add more tests for the behaviour of `rich_markup_mode` 

svlandeg:
There's a few open PRs around markdown formatting and parsing that I want to review, but first I would like to establish what exactly is the desired outcome for different docstrings and different settings of `rich_markup_mode`.

This PR is supposed to capture the ideal output for each test, using a `pytest.mark.xfail` marker for those cases that currently fail on `master`. Any potential fixes will then be contributed in subsequent PRs.

### DIFF
--- a/tests/test_rich_markup_mode.py
+++ b/tests/test_rich_markup_mode.py
@@ -1,3 +1,6 @@
+from typing import List
+
+import pytest
 import typer
 import typer.completion
 from typer.testing import CliRunner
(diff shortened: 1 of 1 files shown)


tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,186 @@ def main(arg: str):
...
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        ("markdown", ["First line", "Line 1 Line 2 Line 3", ""]),
+        ("rich", ["First line", "Line 1", "", "Line 2", "", "Line 3", ""]),
+        ("none", ["First line", "Line 1", "Line 2", "Line 3", ""]),
+    ],
+)
+def test_markup_mode_newline_pr815(mode: str, lines: List[str]):
```
svlandeg, line 54: Here, the question is mostly about the newlines: should there be one after "First line" or not?

Further, should the `markdown` parser output "Line 1 Line 2 Line 3" as one string, or split them into multiple strings, similar to the current behaviour for the option `none`?

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,186 @@ def main(arg: str):
...
+            [
+                "This header is just pretty long really",
+                "Line 1 of a very extremely super long line",
+                "And a short line 2",
+                "",
+            ],
+        ),
+    ],
+)
+def test_markup_mode_newline_issue447(mode: str, lines: List[str]):
```
svlandeg, line 117: Here, the splitting of the strings is different for all 3 options, with `rich` keep each input line as such, `markdown` merging everything that is not header, and `none` keeping all double newlines as separators.

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,186 @@ def main(arg: str):
...
+
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        pytest.param(
+            "markdown",
+            ["First line", "", "• 1", "• 2", "• 3", ""],
+            marks=pytest.mark.xfail(),
+        ),
```
svlandeg, line 156: This test currently fails on `master`, with the output being 
```
['First line', '', '• 1 - 2 - 3', '']
```
-> this happens mostly because the newlines are not parsed correctly, after which the bullet points aren't recognized correctly. PR #815 attempts to fix this.

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,186 @@ def main(arg: str):
...
+
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        pytest.param(
+            "markdown",
+            ["First line", "", "• 1", "• 2", "• a", "• b", "• 3", ""],
+            marks=pytest.mark.xfail(),
+        ),
```
svlandeg, line 195: This test currently fails on `master`, with the output being 
```
['First line', '', '• 1 - 2   - a   - b - 3', '']
```
-> this happens mostly because the newlines are not parsed correctly, after which the bullet points aren't recognized correctly. PR #815 attempts to fix this.

tests/test_rich_markup_mode.py
tiangolo: Yep! In all 3 modes the lines *should* be:

* `First line`
* new line
* `Line 1`
* new line
* `Line 2`
* new line
* `Line 3`
* new line

```
["First line", "", "Line 1", "", "Line 2", "", "Line 3", ""]
```

I just tested with pure Rich:

```Python
from rich.console import Console
from rich.markdown import Markdown

console = Console()

content = """
First line

Line 1

Line 2

Line 3
"""

md = Markdown(content)


print("--- Markdown ---")

console.print(md)

print("--- Rich markup ---")

console.print(content, markup=True)

print("--- plain ---")

print(content)
```

And the output is almost always the same:

```console
$ python demo.py
--- Markdown ---
First line                                                                                                                                                                                                               

Line 1                                                                                                                                                                                                                   

Line 2                                                                                                                                                                                                                   

Line 3                                                                                                                                                                                                                   
--- Rich markup ---

First line

Line 1

Line 2

Line 3

--- plain ---

First line

Line 1

Line 2

Line 3


```

The main difference at least in this script is that for Markdown Rich doesn't add an extra line on top and below. But the main question we have right now are lines in between, and in all cases those are printed consistently.

So yep, that's a bug here in Typer.

tests/test_rich_markup_mode.py
tiangolo: Hum, this gets very tricky. :thinking: 

Because I see how I would intuitively expect the lines to be split, visually it's just a list.

But then, that would go against the automatic text unwrapping that Click would do and that I was mentioning above. :thinking: 

One option is to expect it all to be unwrapped, so that if a user wants a list they would separate the items by new lines.

The other option is to challenge the first idea of automatically unwrapping new lines into a single paragraphs, and instead render things as-is, even if they have internal splits.

A third option I wouldn't want is adding a special case for lists, but that would open many edge cases and many other possible edge cases to handle, so I think I wouldn't like this idea, maybe better to decide on one format or the other.

tiangolo (commented):
Great idea to have more tests and a more established and expected behavior!

I added some comments in your comments, thinking about the best way to proceed with unwrapping lines or not.

Now I was just reading the docstrings PEP: https://peps.python.org/pep-0257/, it mentions that the first line should be a "summary", followed by a blank line, and that it should fit on a single line.

So I guess that for the first line it's up to us.

The docstrings PEP doesn't say anything about wrapping long lines or not, or how to consider it.

But PEP 8 (https://peps.python.org/pep-0008/#maximum-line-length) does, and explicitly mentions docstrings:

> [...] it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

So I guess it's kind of expected that people would wrap docstring content with internal new lines without signifying that it means new paragraphs.

So, although I don't like it too much, I'm inclined to not support lists separated by a single line and require them to have blank lines in between.

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,304 @@ def main(arg: str):
...
+        pytest.param(
+            "markdown",
+            ["First line", "", "Line 1", "", "Line 2", "", "Line 3", ""],
+            marks=pytest.mark.xfail,
+        ),
+        pytest.param(
+            "rich",
+            ["First line", "", "Line 1", "", "Line 2", "", "Line 3", ""],
+            marks=pytest.mark.xfail,
+        ),
```
svlandeg, line 58: This one fails on `master` only because of a missing `""` right after the header

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,304 @@ def main(arg: str):
...
+    assert result_lines[help_start:arg_start] == lines
+
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        ("markdown", ["First line", "", "• 1 - 2 - 3", ""]),
+        pytest.param(
+            "rich", ["First line", "", "- 1 - 2 - 3", ""], marks=pytest.mark.xfail
+        ),
```
svlandeg, line 174: Note that currently on `master`, the output for this one is
> ['First line', '- 1', '- 2', '- 3', '']

This may mean that some users would see it as a "breaking" change once we change the code to succeed this test.

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,304 @@ def main(arg: str):
...
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        ("markdown", ["First line", "", "• 1", "• 2", "• 3", ""]),
+        pytest.param(
+            "rich",
+            ["First line", "", "- 1", "", "- 2", "", "- 3", ""],
+            marks=pytest.mark.xfail,
+        ),
```
svlandeg, line 216: This one fails on `master` only because of a missing `""` right after the header

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,304 @@ def main(arg: str):
...
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        ("markdown", ["First line", "", "• 1 - 2   - a   - b - 3", ""]),
+        pytest.param(
+            "rich",
+            ["First line", "", "- 1 - 2  - a  - b - 3", ""],
+            marks=pytest.mark.xfail,
+        ),
```
svlandeg, line 261: Note that currently on `master`, the output for this one is
> ['First line', '- 1', '- 2', '- a', '- b', '- 3', '']

This may mean that some users could see it as a "breaking" change once we change the code to succeed this test.

tests/test_rich_markup_mode.py
```diff
@@ -38,3 +41,304 @@ def main(arg: str):
...
+
+@pytest.mark.parametrize(
+    "mode,lines",
+    [
+        ("markdown", ["First line", "", "• 1", "• 2", "• a", "• b", "• 3", ""]),
+        pytest.param(
+            "rich",
+            ["First line", "", "- 1", "", "- 2", "", "- a", "", "- b", "", "- 3", ""],
+            marks=pytest.mark.xfail,
+        ),
```
svlandeg, line 306: This one fails on `master` only because of a missing `""` right after the header

svlandeg:
Hi @tiangolo,

Thanks for the detailed feedback!

I agree that the trade-off is difficult here, because some users may "feel like" single lines should be unwrapped, while others may feel like they should remain separated, e.g. when using lists. I don't think we'll be able to satisfy all use-cases, but at least aiming for consistency will be good.

I've edited all tests to try and reflect your feedback - could you please go through them one final time to double check this is how you intend it?

Two important points that I've also highlighted in individual comments:
- For the tests with lists & double lines, `rich` formatting is mostly OK already on `master`, except that there's currently a missing newline right after the header. From your very first comment on the first test `test_markup_mode_newline_pr815` I gather that you always want this newline to be present in the output after the header, correct?
- For the tests with lists & single lines, note that the `rich` formatting option is currently parsing the lists and not unwrapping the single newline. If I understood you correctly, we want to reverse this behaviour, glue the lines together, and only support lists when there's double lines. This is what the tests currently reflect. However, I could imagine some users taking issue with this decision, because it will mean that their specific use-case (lists with single line) may break in a future version of Typer. If this is what we decide, we will need to communicate about this clearly.

tests/test_rich_markup_mode.py
svlandeg: Note: changed the expected behaviour so that this wouldn't be "breaking" anymore.

tests/test_rich_markup_mode.py
svlandeg: Note: changed the expected behaviour so that this wouldn't be "breaking" anymore.

svlandeg:
Ok, after internal discussion with @tiangolo, we've decided on the state of things as they're currently represented by the tests. In contrast to the earlier discussion, we've decided to support lists with single line breaks anyway, as `rich` is already capable of doing this and we wouldn't need complicated logic on our end.

With respect to the (new) test with docstring 
```
        """
        First line

        Line 1
        Line 2
        Line 3
        """
```
the lines of the body will be merged together for `markdown` and `None` mode, but not for Rich mode, mimicking more closely what native Rich does.

tiangolo (approved):
Thank you @svlandeg ! :rocket: :cake:

svlandeg:
Also, this PR now already includes the quick bug fix that ensures there's a newline between the header and the body.

tiangolo:
This is available now in Typer 0.15.3, just released :rocket:

PR #979, opened 2024-09-08
Fixed issue with custom classes

AryazE:
Similar to [This issue in `rich`](https://github.com/Textualize/rich/issues/2875), comparing types with `==` could be problematic. The canonical `is` would fix this problem.

### DIFF
--- a/tests/test_annotated.py
+++ b/tests/test_annotated.py
@@ -76,3 +76,30 @@ def test_annotated_option_with_argname_doesnt_mutate_multiple_calls():
     result = runner.invoke(app, ["--force"])
     assert result.exit_code == 0, result.output
     assert "Forcing operation" in result.output
+
+def test_default_with_class_with_custom_eq():
+    app = typer.Typer()
+
+    from typer.models import ParamMeta
+    class StupidClass:
+        def __init__(self, a):
+            self.a = a
+
+        def __eq__(self, other) -> bool:
+            if other is ParamMeta.empty:
+                return True
+            try:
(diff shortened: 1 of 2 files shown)


svlandeg (approved):
Thanks for the PR!

Changing the check from equality to identity check seems correct to me, also considering that the official Python [documentation](https://docs.python.org/3/library/inspect.html#inspect.Parameter) on `inspect.Parameter` also includes this snippet:
```
if param.default is param.empty
```

The test you included fails on `master` and succeeds on this PR. I only added 2 more lines to ensure 100% code coverage.

tests/test_annotated.py
```diff
@@ -76,3 +76,35 @@ def cmd(force: Annotated[bool, typer.Option("--force")] = False):
     result = runner.invoke(app, ["--force"])
     assert result.exit_code == 0, result.output
     assert "Forcing operation" in result.output
+
+
+def test_default_with_class_with_custom_eq():
```
svlandeg, line 81: Tiangolo: we might want this test in a different location in the test suite?

tiangolo: Good point! I just moved it to another file. Not sure about the name, but it will work for now. :shrug:

tiangolo (approved):
Cool, thanks! :rocket: 

I don't see an obvious use case where this would be a problem as types in parameters for Typer are from the default Python types, but anyway, it doesn't hurt having it. :nerd_face: 

Thanks @AryazE!

Thanks @svlandeg for the help! :bow:

PR #983, opened 2024-09-09
📝 Update optional CLI argument section in tutorial with `Annotated`

gkeuccsr:
To match the example, description is updated


svlandeg (commented):
Thanks for the PR! The text has indeed gotten a bit out of date as the example code now uses `Annotated`. 

I've rewritten the text again, because we shouldn't focus on the `Optional` part of the type annotation, as that is not what is making the CLI argument optional 🙃  We need to focus the reader on the fact that there is a default value; whether that is `None` or not and whether or not the type is `Optional[X]` is irrelevant. We could also have had 
```
name: Annotated[str, typer.Argument()] = "Rick"
```
for example. This is sufficiently explained in various parts of the documentation so doesn't need repeating here, but we just need to make sure to avoid confusion with the rephrasing.

Anyway - thanks again for spotting this!

docs/tutorial/arguments/optional.md
```diff
@@ -113,7 +113,7 @@ name: str
 
 Now, finally what we came for, an optional *CLI argument*.
 
-To make a *CLI argument* optional, use `typer.Argument()` and pass a different "default" as the first parameter to `typer.Argument()`, for example `None`:
+To make a *CLI argument* optional, use `typer.Argument()` and make sure to provide a "default" value, for example `None`:
```
svlandeg, line 116: I've now rewritten it so that it covers both `Annotated` and non-Annotated code snippets:
```
name: Optional[str] = typer.Argument(default=None)
```
&
```
name: Annotated[Optional[str], typer.Argument()] = None
```

gkeuccsr: From the point of someone, who were reading the tutorial for the first time (me), that's still confusing if you go from the previous examples:
> ```Python
> name: Annotated[str, typer.Argument()]
> ```
> ...that works exactly the same as
> ```Python
> name: str
> ```

where you only see `typer.Argument()` used inside the type definition. And peeking in the next example from the *phrase* in subject, you see only:

> Now we have:
> ```Python
> name: Annotated[Optional[str], typer.Argument()] = None
> ```


So I'd suggest adding both variants you mentioned in tutorial, like:

> ```Python
> name: Annotated[Optional[str], typer.Argument()] = None
> ```
> ...that works exactly the same as
> ```Python
> name: Optional[str] = typer.Argument(default=None)
> ```

And placing the *Annotated* version first as it is recommended in the rest of the tutorial.

svlandeg: Hm, I see.

I wonder whether it would be less confusing if the first example of an optional argument doesn't actually use `None` as default, but some other random value. I think that would help enforce the idea that the default is all that matters, and that `Optional` is only used to support `None` as a potential value. What do you think?

svlandeg (approved):
I've gone ahead and edited this PR further, and I chose to use a default argument `"World"` in this part of the docs instead of `None`, to more clearly separate the meaning of "an optional argument" from the type `Optional`. This meant changing `tutorial002` and simplifying it - I think this should make the section more clear. 

I'll leave the final review with Tiangolo.

docs/tutorial/arguments/optional.md
```diff
@@ -109,21 +109,27 @@ name: str
...
 /// tip
 
-By using `Optional` your editor will be able to know that the value *could* be `None`, and will be able to warn you if you do something assuming it is a `str` that would break if it was `None`.
+If you want the default value to be `None`, you have to additionally declare the parameter to be of type `Optional`, which will tell your editor that the value of this parameter can be `None`:
+
+```Python
+name: Annotated[Optional[str], typer.Argument()] = None
+```
+
+Your editor can then warn you if you do something assuming it is a `str` that would break if it was `None`.
```
svlandeg, line 132: I'm not even sure we need all this additional explanation about `None` - it may still cause some confusion and we're probably already explaining `Optional` in various other places.

tiangolo: Agreed! I'm gonna remove it. 🤓

tiangolo (approved):
Great, thank you @gkeuccsr! :rocket: :coffee:

PR #984, opened 2024-09-09
Make links to tutorial and virtual-environment relative in index.md

gkeuccsr:
Otherwise running a custom build refers to a official published versions and not the ones build (in PRs, for ex.)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -52,7 +52,7 @@ The key features are:
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


gkeuccsr:
Yikes, readme.md must be equal to docs index.md :(

YuriiMotov:
> Yikes, readme.md must be equal to docs index.md :(

Yes, I think this is the reason why we have absolute links in `index.md`.
`readme.md` is generated automatically from `index.md`. And links in the `readme` should be absolute.

I suggest we close this PR.

@gkeuccsr, thanks for your interest and efforts!

Issue #130, opened 2020-06-28
Support parameter type typing.Dict[str, typing.Any]


tekumara:
My CLI app can use a profile key, which is used to load the appropriate config:

```
def load_profile(profile: str) -> Dict[str, Any]:
    ...

@app.command()
def describe(
    config: Dict[str, Any] = typer.Option(None, "--profile", help="Profile in the config file to use", callback=load_profile),
) -> List[Dict[str, Any]]:
```

However, this errors with:
```
RuntimeError: Type not yet supported: typing.Dict[str, typing.Any]
```

### The solution you would like

The ability to support Dict as a type, when returned from a callback function.

tekumara:
The workaround, for now, is to remove the type hint from the `config` param (which also disables type checking).

pavan-uppari:
# Handling Dict type.

Issue - If we pass dict directly from terminal, it will be coming as string to us. 

Fix: We can use pydantic `Json` with given dict type and use pydantic `TypeAdapter` to validate data and get the actual value

Usage

```py
import json
from pydantic import TypeAdapter
from pydantic.types import Json

dict_type = dict[str, int]  #type from method annotations
data = '{"key": 1, "key2": 2}'  #input from terminal - json data

type_adapter = TypeAdapter(Json[dict_type])
t2 = type_adapter.validate_python(data, strict=True)
print(t2) #pass this value to function

```

Usage with `typer`

```py

def main(data: dict[str, int]):
    print(f"data is {data}")

```

```console
$ typer main.py run {"key": 1, "key2": 2}
data is {"key": 1, "key2": 2}
```

> NOTE: User should provide json data as input

@tiangolo If this is ok, we can follow the same for classes as types

pavan-uppari:
> # Handling Dict type.
> Issue - If we pass dict directly from terminal, it will be coming as string to us.
> 
> Fix: We can use pydantic `Json` with given dict type and use pydantic `TypeAdapter` to validate data and get the actual value
> 
> Usage
> 
> ```python
> import json
> from pydantic import TypeAdapter
> from pydantic.types import Json
> 
> dict_type = dict[str, int]  #type from method annotations
> data = '{"key": 1, "key2": 2}'  #input from terminal - json data
> 
> type_adapter = TypeAdapter(Json[dict_type])
> t2 = type_adapter.validate_python(data, strict=True)
> print(t2) #pass this value to function
> ```
> 
> Usage with `typer`
> 
> ```python
[...]

PR #985, opened 2024-09-09
✨ Support `json` (dict) as parameter type

mhkarimi1383:
Hi

Sometimes there is a need of decoding entire value as an json object (useful when some params are too dynamic or getting that option in a prompt). I wanted to also add pydantic support (for both single option as json and all of the options in a single model) there and also use orjson (by checking if that's installed as an optional dependency), If that's Ok.


mhkarimi1383:
@svlandeg it's ok to ignore coverage on newly created class?

mhkarimi1383:
@svlandeg

I have added tests for newly created class

mhkarimi1383:
If changes are Ok so far I can work on Pydantic support, Or move Pydantic support in another PR?

@svlandeg

svlandeg:
HI @mhkarimi1383! 

We haven't yet been able to review this PR in detail. When we do, we'll update here. No need to ping individual maintainers in the meantime 😉 

In terms of adding more functionality, it's always best to open separate PRs for distinct functionality. It makes the review process much easier if one PR deals with one "atomic" change.

Do have a look at existing PRs - if I recall correctly there's already a few PRs open with proposals for Pydantic support.

svlandeg:
Thanks again for this PR @mhkarimi1383! I'm going to start reviewing it in detail and will push some changes directly to your branch as I'm working on this. I'll put this PR in draft while I do so.

svlandeg:
Related issue: https://github.com/fastapi/typer/issues/130

svlandeg (commented):
I think the status of this PR should now be good enough for me to pass it onwards to Tiangolo for further review 🙏

mhkarimi1383:
We are calling it dict, but cli arg should be provided as JSON since we are using json to parse input

svlandeg:
> We are calling it dict, but cli arg should be provided as JSON since we are using json to parse input

Yea, I'm a little on the fence as to what to call it best. In the tutorial it's about parameter types and there it is typed as a `dict` though... Will let Sebastián weigh in on this one.

mhkarimi1383:
Yeah, but I think in the repr we can call it JSON, So CLi users that they want to use our application should know that the input has to be JSON

mrfroggg:
8 months after, would be nice to have this merged and released! I would use it!

typer/main.py
```diff
@@ -711,8 +712,10 @@ def get_click_type(
     elif parameter_info.parser is not None:
         return click.types.FuncParamType(parameter_info.parser)
 
-    elif annotation is str:
+    elif annotation in [str, bytes]:
```
svlandeg: Let's remove this edit from this PR, to focus here on the `dict` addition. There's also another PR that deals with `bytes`: https://github.com/fastapi/typer/pull/1190. This particular change is untested in this PR anyway.

svlandeg:
```suggestion
    elif annotation is str:
```

mhkarimi1383:
I have updated the description to not include bytes support

tests/test_tutorial/test_parameter_types/test_dict/test_tutorial001.py
```diff
@@ -0,0 +1,49 @@
...
+
+import typer
+from typer.testing import CliRunner
+
+from docs_src.parameter_types.dict import tutorial001 as mod
+
+runner = CliRunner()
+
+app = typer.Typer()
+app.command()(mod.main)
```
svlandeg:
```suggestion
app = mod.app
```

tests/test_tutorial/test_parameter_types/test_dict/test_tutorial001_an.py
```diff
@@ -0,0 +1,49 @@
...
+
+import typer
+from typer.testing import CliRunner
+
+from docs_src.parameter_types.dict import tutorial001_an as mod
+
+runner = CliRunner()
+
+app = typer.Typer()
+app.command()(mod.main)
```
svlandeg:
```suggestion
app = mod.app
```

mhkarimi1383:
@svlandeg if this will not get merged I will close it...

PR #987, opened 2024-09-13
🗑️ Deprecate support for `is_flag` and `flag_value` parameters

### LINKED ISSUE #986: Deprecate support for `is_flag` and `flag_value` parameters
### Privileged issue

### Issue Content

Typer's code base currently contains `is_flag` and `flag_value` parameters like Click does. However, the current behaviour is inconsistent with Click's behaviour, and not documented.

In discussion with @tiangolo, we've come to the conclusion that these parameters are pretty confusing overall to work with. There are essentially two defaults (one set in code, and one set by `flag_value`) and then an option to set an arbitrary value to something that is a "flag". We feel that any behaviour obtained by this combination of parameters can probably be achieved easily with some internal logic and more straightforward (and properly typed) options/arguments.

As such, we've decided to remove these options from the code base as they're not really functional right now anyway.

svlandeg:
Fixes #986

- Raise a `DeprecationWarning` when `OptionInfo` receives a value for `is_flag` and/or `flag_value`
- `TyperOption` doesn't use `flag_value` anymore and its `is_flag` is now defined only by looking at whether or not `main_type is bool`

These parameters weren't actually functional (cf. https://github.com/fastapi/typer/discussions/873) and they weren't documented, so I don't think this should be breaking. But it's always difficult to say - users could be relying on unspecified internals... 🫤


typer/completion.py
```diff
@@ -68,15 +68,13 @@ def _install_completion_placeholder_function(
...
         "--install-completion",
-        is_flag=True,
         callback=install_callback,
         expose_value=False,
         help="Install completion for the current shell.",
     ),
     show_completion: bool = Option(
         None,
         "--show-completion",
-        is_flag=True,
```
svlandeg, line 79: I'm not sure why we had these `is_flag` specifications here. I think we can just remove them, because we removed the `parameter_info.is_flag is not False` condition.

tiangolo (approved):
Great! Looks good! :rocket:

Plenitude-ai:
Hello !
Sorry to come a bit late, we previously relied on another typer version (~0.12.5), and just changed for the newer (~0.15.0)
We did use this feature for handling boolean values with explicit values :
```python
import typer


def main(my_bool: bool = typer.Option(True, is_flag=False)):
    print(f"my_bool : {my_bool}")


if __name__ == "__main__":
    typer.run(main)
```
Which could be used with :
```sh
python main.py --my-bool true
my_bool : True
type(my_bool) : <class 'bool'>
```
OR :
```sh
python main.py --my-bool false
my_bool : False
type(my_bool) : <class 'bool'>
```

For more details, we use argo workflows to submit scripts, therefore we cannot dynamically create the --my-bool or --not-my-bool string in the argo workflow template.
For the moment we implemented a workaround like :
my_bool = True if my_bool.lower()=="true else False
But this forces the definition of the argument to be string in the signature, and then to cast it to bool, which is not the best.

This would be wonderful if we could get this feature back !

twitthoefft:
I think the original conclusion that `is_flag` was non-functional is incorrect.  When the option type is `bool`, setting the `is_flag=False` parameter changes the default behavior of being _flag-like_ to an `Option` that requires a passed argument.

In version `0.12.5` you can see this when checking `--help` as shown in the example above:

When `is_flag` is set to the default value of `True` the opposite flag argument `--no-my-bool` is created automatically.

```bash
 Usage: typertest.py [OPTIONS]                                               
                                                                             
╭─ Options ─────────────────────────────────────────────────────────────────╮
│ --my-bool    --no-my-bool      [default: my-bool]                         │
│ --help                         Show this message and exit.                │
╰───────────────────────────────────────────────────────────────────────────╯
```

But when `is_flag` is set to `False` the flag `--no-my-bool` is not created.

```
                                                                                 
 Usage: typertest.py [OPTIONS]                                                    
                                                                                  
╭─ Options ──────────────────────────────────────────────────────────────────────╮
│ --my-bool          [default: True]                                             │
│ --help             Show this message and exit.                                 │
╰────────────────────────────────────────────────────────────────────────────────╯
```

Failing to pass an argument to `--my-bool` now behaves like any other `Option`

```bash
╭─ Error ────────────────────────────────────────────────────────────────────────╮
│ Option '--my-bool' requires an argument.                                       │
╰────────────────────────────────────────────────────────────────────────────────╯
```

svlandeg:
Thanks for your reports! Let me look into this (again) and get back to you 🙏

riklopfer:
@svlandeg, thanks for taking another look at this. I'm having a similar issue porting a pure click script to typer. Would be nice to support passing booleans "by value" from the command line and I believe that `is_flag=False` was the only way to do this. 

As it stands, it appears this click annotations not supported out of the box, 

```python
@click.option(
    "--my-bool",
    type=bool,
    default=False,
)
```

where you pass the value like `--my-bool=true`

PR #988, opened 2024-09-13
🐛 Allow colon in zsh autocomplete values and descriptions

### LINKED ISSUE #162: [BUG] Completions that include colons get separated in zsh
### Describe the bug

I have completions (via the `autocompletion` keyword argument in Arguments and Options) that include a colon (`:`) in the string. In zsh, these completions get split into completion candidates and help texts at the colon.

### To Reproduce

```Python
import typer

images = ["alpine:latest", "nvidia/cuda:10.0-devel-ubuntu18.04"]

def _complete(value: str) -> str:
    for image in images:
        if image.startswith(value):
            yield value

app = typer.Typer()

@app.command("image")
def hello(name: str = typer.Option(..., autocompletion=_complete)):
    typer.echo(name)

if __name__ == "__main__":
    app()
```

* Trigger autocompletion with [TAB] on ZSH

```bash
$ <app_entrypoint> image --name [TAB]
```

* It outputs:

```
$ <app_entrypoint> image --name [TAB]
alpine       -- latest
nvidia/cuda  -- 10.0-devel-ubuntu18.04
```
When I cycle through the completion candidates, only `alpine` and `nvidia/cuda` gets filled in.

```
$ <app_entrypoint> image --name [TAB]
alpine:latest
nvidia/cuda:10.0-devel-ubuntu18.04
```

### Environment

* OS: Ubuntu 18.04
* Typer Version: 0.3.2
* Python version: 3.7.9
* Zsh version: 5.4.2 (x86_64-ubuntu-linux-gnu)

### Additional context

I tried the exact same thing in Bash, and it worked as expected (the output in "But I expected it to output").

snapbug:
Allow for colons in zsh completion values and descriptions, related to #162 

Note that because of the [`startswith` check](https://github.com/fastapi/typer/blob/master/typer/core.py#L90) escaping the completion options in the client side is not an option, as a value that includes a `:` for `incomplete` in will not pass this check.

Note that we need two levels of esacping, one for zsh, and one to preserve the escape for zsh

Tested by updating the `demo.py` example in the linked issue to the following:

```python
import typer

image_desc = [
    ("alpine:latest", "latest alpine image"),
    ("alpine:hello", "fake image: for testing"),
    ("nvidia/cuda:10.0-devel-ubuntu18.04", ""),
]

def _complete(incomplete: str) -> str:
    for (image, desc) in image_desc:
        if image.startswith(incomplete):
            yield image, desc

app = typer.Typer()

@app.command()
def image(name: str = typer.Option(autocompletion=_complete)):
    typer.echo(name)

if __name__ == "__main__":
    app()
```

use the provided docker image/setup to test, first that the colons show as values and descriptiosn correctly

```
0252e3324b42# typer demo.py run --name [TAB]
alpine:hello                        -- fake image: for testing
alpine:latest                       -- latest alpine image
nvidia/cuda:10.0-devel-ubuntu18.04
```

and that an incomplete value with a colon generates options

```
0252e3324b42# typer demo.py run --name alpine:[TAB]
alpine:hello   -- fake image: for testing
alpine:latest  -- latest alpine image
```

### DIFF
--- a/tests/test_completion/test_completion_option_colon.py
+++ b/tests/test_completion/test_completion_option_colon.py
@@ -5,6 +5,16 @@ import sys
 from . import colon_example as mod
 
 
+def test_script():
+    result = subprocess.run(
+        [sys.executable, "-m", "coverage", "run", mod.__file__, "--name", "DeadPool"],
+        capture_output=True,
+        encoding="utf-8",
+    )
+    assert result.returncode == 0
+    assert "DeadPool" in result.stdout
+
+
 def test_completion_colon_bash_all():
     result = subprocess.run(
         [sys.executable, "-m", "coverage", "run", mod.__file__, " "],
--- a/tests/test_completion/test_completion_option_colon.py
+++ b/tests/test_completion/test_completion_option_colon.py
@@ -1,0 +1,209 @@
+import os
+import subprocess
+import sys
+
+from . import colon_example as mod
+
+
+def test_completion_colon_bash_all():
+    result = subprocess.run(
+        [sys.executable, "-m", "coverage", "run", mod.__file__, " "],
+        capture_output=True,
+        encoding="utf-8",
+        env={
+            **os.environ,
+            "_COLON_EXAMPLE.PY_COMPLETE": "complete_bash",
+            "COMP_WORDS": "colon_example.py --name ",
+            "COMP_CWORD": "2",
+        },
+    )
+    assert "alpine:hello" in result.stdout
+    assert "alpine:latest" in result.stdout
+    assert "nvidia/cuda:10.0-devel-ubuntu18.04" in result.stdout
+
+
+def test_completion_colon_bash_partial():
+    result = subprocess.run(
+        [sys.executable, "-m", "coverage", "run", mod.__file__, " "],
+        capture_output=True,
+        encoding="utf-8",
+        env={
+            **os.environ,
+            "_COLON_EXAMPLE.PY_COMPLETE": "complete_bash",
+            "COMP_WORDS": "colon_example.py --name alpine ",
+            "COMP_CWORD": "2",
+        },
+    )
+    assert "alpine:hello" in result.stdout
+    assert "alpine:latest" in result.stdout
+    assert "nvidia/cuda:10.0-devel-ubuntu18.04" not in result.stdout
+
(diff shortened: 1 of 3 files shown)


svlandeg:
Thanks for the PR! We'll review this and get back to you. Ideally, we'd love to also add a unit test for this (but that's not always easy with the autocompletion functionality)

snapbug:
> Thanks for the PR! We'll review this and get back to you. Ideally, we'd love to also add a unit test for this (but that's not always easy with the autocompletion functionality)

Agree, I took a look at the current completion tests, but they're tightly integrated into the tutorial, and couldn't think of a change for those that wasn't somewhat contrived. Open to guidance here as well.

svlandeg (approved):
I could confirm the bug in a `zsh` shell.

On `master`:
```
typer main.py run --name [TAB]
alpine       -- hello:fake image: for testing
nvidia/cuda  -- 10.0-devel-ubuntu18.04
```
Note that the options are wrongly split and the `alpine:latest` option is missing because it got collapsed into `alpine:hello`

With this PR:
```
typer main.py run --name [TAB]
alpine:hello                        -- fake image: for testing
alpine:latest                       -- latest alpine image
nvidia/cuda:10.0-devel-ubuntu18.04
```
Which is what you would expect.

I added tests for this use-case, the `zsh` ones are failing on `master` and succeeding on this branch.

I'll leave the final review up to Tiangolo.

tests/test_completion/test_completion_option_colon.py
```diff
@@ -0,0 +1,219 @@
...
+
+def test_completion_colon_bash_single():
+    result = subprocess.run(
+        [sys.executable, "-m", "coverage", "run", mod.__file__, " "],
+        capture_output=True,
+        encoding="utf-8",
+        env={
+            **os.environ,
+            "_COLON_EXAMPLE.PY_COMPLETE": "complete_bash",
+            "COMP_WORDS": "colon_example.py --name alpine:hell ",
```
tiangolo, line 60: I have struggled with Alpine hell too :joy:

tiangolo (approved):
Awesome, thank you @snapbug! :rocket:

And thank you @svlandeg for the help, guidance, tests, review. :bow:

PR #990, opened 2024-09-15
small clarification in the tutorial docs for cli arguments

mina58:
In the optional cli arguments section, it said "To make a CLI argument optional, use typer.Argument() and pass a different "default" as the first parameter to typer.Argument(), for example None:" and the example was:
```python
from typing import Optional

import typer
from typing_extensions import Annotated


def main(name: Annotated[Optional[str], typer.Argument()] = None):
    if name is None:
        print("Hello World!")
    else:
        print(f"Hello {name}")


if __name__ == "__main__":
    typer.run(main)
```

I thought it was confusing because when I read it for the first time i though i should write it like this 
```python
def main(name: Annotated[Optional[str], typer.Argument(default=None)]) -> None
```
which is wrong, so i clarified the text a little.

### DIFF
--- a/docs/tutorial/arguments/optional.md
+++ b/docs/tutorial/arguments/optional.md
@@ -113,7 +113,7 @@ name: str
 
 Now, finally what we came for, an optional *CLI argument*.
 
-To make a *CLI argument* optional, use `typer.Argument()` and pass a different "default" as the first parameter to `typer.Argument()`, for example `None`:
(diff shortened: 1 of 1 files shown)


svlandeg:
Hi, thanks for the PR!

This particular sentence in the documentation is already being addressed by https://github.com/fastapi/typer/pull/983, so I'll go ahead and close this one. Thanks again though!

PR #994, opened 2024-09-17
:art: Add devenv/direnv configs

mhkarimi1383:
Making contibutions easier, Also adds `test` and `format` commands to user shell

 Also I can make devenv to manage pre-commit and add pre commit config will be auto generated

For nix users it will be so easier to work on typer


tiangolo:
Thanks for the interest! But I don't use these tools, at least not now. If I start using them, I'll then add configs for them. But for now, I'll pass on this. Thanks! :coffee:

mhkarimi1383:
@tiangolo 
So, as a Nix User I can't contribute on this project without them :)
Since NixOS is so strict, also it will make it easy for user who want to contribute, since it will automatically makes environment ready

PR #1006, opened 2024-09-29
✨ Support `Parameter` in the CLI Option autocompletion functionality

### LINKED ISSUE #949: Improve support for autocompletion: `shell_complete` vs `autocompletion`
### Privileged issue

## In Short

We want a single parameter to handle CLI parameter autocompletion, based on type annotations.

## Current State

Right now the Typer docs only talk about the function for the parameter `autocompletion`.

And Typer has (proper) support for that function based on type annotations.

It checks the types for a `Context`, `list[str]` for other args, and `str` for the current incomplete args: https://typer.tiangolo.com/tutorial/options-autocompletion/#types-types-everywhere

The support for autocompletion is based on the type annotations (like the rest of Typer), not on the order or name of the arguments.

Now with Click 8.x the underlying parameter `autocompletion` was removed and there's a new one `shell_complete` (https://click.palletsprojects.com/en/8.1.x/api/#click.ParamType.shell_complete).

In Typer, `autocompletion` raises a warning already: https://github.com/fastapi/typer/blob/master/typer/core.py#L65

But there are no docs nor proper support for `shell_complete` based on type annotations.

## We Want

We want to have support for completion based on type annotations. The parameter should be able to expect a `Context` and a `str` with the incomplete parameter. But I'm not sure of the parameter name for this function.

**Note**: we probably don't need to support `list[str]` for extra args as that's available from [`context.args`](https://click.palletsprojects.com/en/8.1.x/api/#click.Context.args).

We should have that as the main documented feature. We could have a note at the end mentioning that there's a (now deprecated) `autocompletion` parameter, with the basics of how it worked, but the main functionality would be with the new parameter.

## Hardest Problem in CS: Naming

Writing this I'm realizing that in Typer I'm using the same parameter name that was available in Click: `autocompletion`, but in reality, it has different semantics.

In Click, the meaning of the parameters is based on the position/order.

In [Click 7.x](https://click.palletsprojects.com/en/7.x/bashcomplete/#completion-help-strings) it had to be in the exact order of `ctx`, `args`, `incomplete`:

```Python
def get_colors(ctx, args, incomplete):
    ...
```

In [Click 8.x](https://click.palletsprojects.com/en/8.1.x/shell-completion/#overriding-value-completion) it has to be the exact order of `ctx`, `param`, `incomplete`:

```Python
def complete_env_vars(ctx, param, incomplete):
    ...
```

In Typer, `autocompletion` it's based on the type annotations, which also means that the order doesn't matter, if all of the parameters are present or not doesn't matter, the name of the parameters doesn't matter. The only thing that matters is the type annotations: https://typer.tiangolo.com/tutorial/options-autocompletion/#types-types-everywhere

This means that the name of the parameter `autocompletion` is the same as in Click, but the semantics are actually not the same. :thinking: 

That makes me think that maybe the parameter should have a different name, so that people don't get confused with it assuming it's the same as in Click, and it really is not, at least not in the same way.

Maybe we should just keep the same name `autocompletion` just because that's the main name we've had documented in Typer. :thinking: 

My original thought was to support a parameter `shell_complete` (like the one in Click) but based on type annotations. But now thinking about it, I think maybe it makes more sense to have a different name (or just keep the old `autocompletion` name). And have it based on type annotations.

I think the new/final parameter should be able to request (via type annotations) the `Context` and the `incomplete` `str`. I don't think we should support it requesting the `param` Click `Parameter` until we find a reason to do so (I can't find any yet).

The other thing is, the old Click `autocompletion` supported returning `str` values or tuples. The new `shell_complete` expects objects of type `CompletionItem`, but I wouldn't think requiring users to import that class and create instances of it would be a great developer experience, so I would prefer to keep supporting the older syntax too.

## Note on `CompletionItem`

From the source, I understand that the main thing it provides apart from the previously supported data is a new `type` that tells the completion parts if the thing to complete is a directory or file... I'm not sure in which cases that type is useful, maybe internally, but not sure for completion functions. And also, if there's a use case where that makes sense, we can probably know the type from the type annotation (e.g. `Path`).

bckohan:
Cf issue: #949

Note, this <s>includes (and supersedes) the changes from #974 and</s> implements these suggestions https://github.com/fastapi/typer/issues/949#issuecomment-2345047182, with a few differences.

I attempted to push these changes to #974 but was [asked to open a new PR here instead](https://github.com/svlandeg/typer/pull/1#issuecomment-2357820955).

This PR does 3 things:

1. Allows autocompletion functions to accept a ``click.core.Parameter`` argument. I think this is necessary because without it you can't define generic completer functions that don't know what CLI options or arguments they may be attached to and that need to alter their behavior based on how click has processed the arguments before the incomplete argument. Concrete example: you have an option or argument that accepts multiple values but you do not wish those values to be repeated.

2. Allows CompletionItems to be returned in addition to strings and 2-tuples from autocompletion functions.

3. Fixes arg: **edit** this change was separated out in #1134 

**A reusable completer function tutorial has been added to the autocompletion tutorials**


svlandeg:
Thanks for the PR! We'll get this reviewed and get back to you 🙏

svlandeg:
@bckohan: the tests are failing because there isn't a 100% code coverage with the new commits. Could you look into this? 🙏

bckohan:
I've updated the docs and added the necessary tests. Should be good to go!

svlandeg:
Ok so looking at the diffs of these two PRs (this one and #949) it appears that all the changes from #949 are still included here verbatim. As such I think it's actually easier to review these changes separately, i.e. one PR at a time. I've also pulled out the typo changes and will submit those separately.

@bckohan: do you see any reason not to merge #949 first, then continue working on the functionality from this PR?

bckohan:
@svlandeg as long as the two PRs are merged in the same release that should be ok. Otherwise #949 alone will break downstream software.

bckohan:
@tiangolo Its not exactly broken - but deprecation warnings are being thrown with no recourse to fix them because autocompletion does not pass the Parameter objects. There are also downstream completers that expect to be able to pass a CompletionItem back because they're setting the type field.

tiangolo:
I would think now there would be fewer warnings as the `autocompletion` parameter is now undeprecated. :sweat_smile: 

And the other `shell_complete` parameter was never documented, and was not completed, so I would expect people would not be using it yet. :thinking: 

> There are also downstream completers that expect to be able to pass a CompletionItem back because they're setting the type field.

I'm not sure I got this, could you share an example?

It might also be useful to start in a GitHub Discussion with a minimal example we can copy paste that shows the problem (or if there are multiple problems, multiple examples).

bckohan:
HI!

I think its reasonable to expect significant usage of shell_complete for the following reasons:

1. The deprecation message for autocompletion pointed to it.
2. It's documented by Click and was in the named parameter lists of ``Argument``/``Option``.
3. Without the ``Parameter`` object there are things you simply can't do in generic completer functions (i.e. functions that are written without knowing which parameter they are going to be asked to complete).
4. It's much more reliable to fetch the parsed argument values for variadic arguments from the ``Context`` using the parameter name. The autocompletion interface as documented would have you manually parse the list of argument strings to determine if you need to exclude previously provided argument values, but see 5.
5. The argument list passed to autocompletion is usually empty because it's derived from ``ctx.args`` [which only includes unparsed leftover arguments](https://click.palletsprojects.com/en/stable/api/#click.Context.args). This PR fixes the behavior and the test that is currently erroneously passing.

I'm [extensively using shell_complete with Parameter objects downstream](https://django-typer.readthedocs.io/en/latest/shell_completion.html#provided-completers). And there's [evidence here](https://github.com/fastapi/typer/issues/334) of others trying. I also added example/explanation of why this is needed to the [tutorial in the docs](https://989516f8.typertiangolo.pages.dev/tutorial/options-autocompletion/#reusing-generic-completer-functions). And here's [real world production code](https://github.com/django-commons/django-typer/blob/2c1d22020bf063f30aa46714abe162183f952537/django_typer/completers.py#L351-L358) where I'm using ``Parameter`` to filter out previously provided model objects for a generic model field completer.

``CompletionItem`` is somewhat less important than all of the above. Click uses it because it's shell specific scripts are using inbuilt file/directory completer logic. To trigger these you have to be able to pass back meta info. I can think of 4 reasons to support this:

1. Click does, so there are likely generic completer functions in the wild that people might want to reuse on typer apps.
2. Its a trivial addition to the code.
3. Gives future implementors of other shells a hook to add meta information that may be useful for enabling whatever super awesome shell specific completer feature they want to use. The arbitrariness of type is important. It could be something other than file or directory.
4. Existing code is probably already passing these objects back (mine is at least!)

bckohan:
I rebased - hopefully it will be easier to review!

svlandeg (commented):
Hi @bckohan,

Apologies for the delay in reviewing this and the slow iterations on our end. While we've split off some of the functionality to separate PRs already, I wonder whether perhaps we should go even further with this. In particular:

> I think numbers 1 and 2 are important. Number 3 is mostly for backwards compatibility but I don't think its that critical - I just wanted to show that there is a way to do it using public Click 8 interfaces.

With Number 3 referring to the change of behaviour of `args`, as reflected by the change in the unit test in `test_others.py`. While I can follow your reasoning, I can't attest to how breaking this would be for other users. It looks like on the other hand, your first two contributions are clear feature additions that wouldn't change any current behaviour (right)?

Knowing the work load that the team, including Tiangolo, already has on the maintenance of Typer and FastAPI (and friends), I would advice to further split this PR into separate parts, and moving the breaking change 3 with respect to `args` to a separate PR that might involve more discussion. When doing so, I would further advice to keep all existing unit tests the same as much as possible, to remove any doubt that there would be breaking changes in existing behaviour.

I realise that you've already put quite a bit of time and effort into this work, and totally understand if you don't have the bandwidth to follow up on this right now. If so - just let me know, and I will have a stab at breaking up this PR further myself next week.

bckohan:
@svlandeg I'm happy to split out the args bug fix into a separate PR. I'll do that ASAP. To be clear its not really a breaking change, its a bug fix to bring behavior in line with [what the current docs say](https://typer.tiangolo.com/tutorial/options-autocompletion/?h=autocompletion#getting-the-raw-cli-parameters). Args is documented to include all of the CLI options before the incomplete string. Currently args is pulled from Context.args which [only contains unrecognized/leftover options](https://click.palletsprojects.com/en/stable/api/#click.Context.args).

docs_src/options_autocompletion/tutorial009.py
```diff
@@ -12,9 +13,8 @@
 err_console = Console(stderr=True)
 
 
-def complete_name(ctx: typer.Context, args: List[str], incomplete: str):
-    err_console.print(f"{args}")
-    names = ctx.params.get("name") or []
+def complete_name(ctx: typer.Context, param: Parameter, incomplete: str):
+    names = ctx.params.get(param.name) or []
```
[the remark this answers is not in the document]
bckohan: Reading closer I see what I was doing. I'm changing the "[Getting the Context and the raw CLI parameters](https://typer.tiangolo.com/tutorial/options-autocompletion/?h=autocompletion#getting-the-context-and-the-raw-cli-parameters)" tutorial to "[Getting the Context and the Parsed CLI parameters](https://27478c2f.typertiangolo.pages.dev/tutorial/options-autocompletion/#getting-the-context-and-the-raw-cli-parameters)". The reasoning for this is just to limit the size of the tutorials to keep it from becoming unwieldy. Instead of a second tutorial illustrating usage of args - which should probably be avoided in place of using the context we instead illustrate using the context and the parameter object to get prior parsed options.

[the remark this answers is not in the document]
bckohan: Ok ok, ignore all previous comments. I think my change here was OBE by my addition of the [Reusing Generic Completer Functions](https://27478c2f.typertiangolo.pages.dev/tutorial/options-autocompletion/#reusing-generic-completer-functions) section. I can revert this tutorial to its prior state. It will, however, as with the first "raw CLI" parameter section require the args bug fix to be accurate.

bckohan:
I've separated out the fix to args here: #1134

docs_src/options_autocompletion/tutorial009.py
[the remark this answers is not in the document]
bckohan: Sorry I forgot to leave a note about this. When I pulled #1134 out I did a run through of the tutorials to make sure everything still flowed well and made sense. There is one tutorial change and one additional tutorial that are part of this PR:

1. In main, tutorial009 is the last tutorial and the point of it is to illustrate that you can receive all the arguments if you would like. Since we're adding param as another argument I simply extended this tutorial to show that you can get all 4.
2. tutorial010 is new and illustrates using the param object to determine which parameter the generic completer function is being asked to complete.

* OBE == Overcome By Events

docs/tutorial/options-autocompletion.md
```diff
@@ -342,7 +372,7 @@ But it's probably useful only in very advanced use cases.
 
 ## Getting the Context and the raw *CLI parameters*
 
-Of course, you can declare everything if you need it, the context, the raw *CLI parameters*, and the incomplete `str`:
+Of course, you can declare everything if you need it, the context, the raw *CLI parameters*, the Parameter and the incomplete `str`:
```
svlandeg:
```suggestion
Of course, you can declare everything if you need it, the context, the raw *CLI parameters*, the `Parameter` and the incomplete `str`:
```

docs/tutorial/options-autocompletion.md
```diff
@@ -381,6 +411,7 @@ You can declare function parameters of these types:
 
 * `str`: for the incomplete value.
 * `typer.Context`: for the current context.
+* `click.Parameter`: for the CLI parameter being completed.
 * `List[str]`: for the raw *CLI parameters*.
 
 It doesn't matter how you name them, in which order, or which ones of the 3 options you declare. It will all "**just work**" ✨
```
svlandeg:
```suggestion
It doesn't matter how you name them, in which order, or which ones of the 4 options you declare. It will all "**just work**" ✨
```

docs/tutorial/options-autocompletion.md
```diff
@@ -264,6 +264,36 @@ It's quite possible that if there's only one option left, your shell will comple
 
 ///
 
+## Reusing generic completer functions
+
+You may want to reuse completer functions across CLI applications or within the same CLI application. If you need to filter out previously supplied parameters the completer function will first have to determine which parameter it is being asked to complete.
+
+We can declare a parameter of type <a href="https://click.palletsprojects.com/en/stable/api/#click.Parameter" class="external-link" target="_blank">click.Parameter</a> along with the `click.Context` in our completer function to determine this. For example, lets revisit our above context example where we filter out duplicates but add a second greeter argument that reuses the same completer function:
```
svlandeg:
```suggestion
We can declare a parameter of type <a href="https://click.palletsprojects.com/en/stable/api/#click.Parameter" class="external-link" target="_blank">click.Parameter</a> along with the `click.Context` in our completer function to determine this. For example, lets revisit our above example where we filter out duplicates but add a second greeter argument that reuses the same completer function:
```

svlandeg (commented):
This feature makes sense to me - e.g. being able to access `param.name` unlocks new functionality I can see being relevant to users.

I have some concerns about this part of the documentation historically using `name` as the parameter to be autocompleted. Because now using `Parameter`, we would go from accessing `ctx.params.get("name")` to `ctx.params.get(param.name)` with `param.name` being `"name"` in this particular case, which just adds unnecessary confusion at this point.

So, while I don't like adding more edits to a PR, I suggest to refactor all tutorial examples here to use `user` instead of `name` as the option we want to autocomplete. Just to 100% avoid this confusion.

I'll put this PR in draft and make those edits directly on the branch to then hopefully get this PR in a good state for merging.

docs_src/options_autocompletion/tutorial009.py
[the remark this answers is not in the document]
svlandeg: Ok, this all looks good now :-)

docs/tutorial/options-autocompletion.md
```diff
@@ -264,11 +264,43 @@ It's quite possible that if there's only one option left, your shell will comple
 
 ///
 
+## Reusing generic completer functions
+
+You may want to reuse completer functions across CLI applications or within the same CLI application. In this case, you need to first determine which parameter is being asked to complete.
+
+This can be done by declaring a parameter of type <a href="https://click.palletsprojects.com/en/stable/api/#click.Parameter" class="external-link" target="_blank">click.Parameter</a>, and accessing its `param.name` attribute.
+
+For example, lets revisit our above example and a second greeter argument that reuses the same completer function, now called `complete_user_or_greeter`:
```
svlandeg:
```suggestion
For example, lets revisit our above example and add a second greeter argument that reuses the same completer function, now called `complete_user_or_greeter`:
```

docs/tutorial/options-autocompletion.md
```diff
@@ -264,11 +264,43 @@ It's quite possible that if there's only one option left, your shell will comple
...
+
+
+Check it:
+
+<div class="termy">
+
+```console
+$ typer ./main.py run --user Sebastian --greeter Camila --greeter [TAB][TAB]
+
+// Our function returns Sebastian too because it is completing 'greeter'
```
svlandeg:
```suggestion
// Our function returns Sebastian too because it is completing 'greeter', not 'user'
```

svlandeg (approved):
Ok, even though this has exploded the diff, I really do think it's more clear to users this way, going from `ctx.params.get("user")` to `ctx.params.get(param.name)`. I've reviewed everything in detail and think this looks good - I'm moving this to Tiangolo's internal TODO list to review.

Thanks again for all this work, @bckohan !

svlandeg (approved):
Ok, the merge conflict after https://github.com/fastapi/typer/pull/1459 was a bit of a challenge, but all should be in good order again, and the upside is that there are now less files changed (from 50 orso to 37). Should again be good to review/merge.

svlandeg:
Aaaaah, this is one of those edge cases Yurii warned about where editing `pyproject.toml` on a user PR is valid and the (new) bot is trigger happy.

That said, I need to revisit this PR now that we've vendored Click. I'll keep this on my TODO list and will look into what's the best course of action here.

PR #1018, opened 2024-10-12
🐛 Ensure that `Optional[list]` values work correctly with callbacks

solesensei:
This pull request refactors the `generate_list_convertor` function to handle the case when the `value` parameter is `None`.

Fixes: https://github.com/fastapi/typer/discussions/762


solesensei:
cc: @svlandeg

svlandeg:
Hi @solesensei, thanks for the PR! We've got this on our internal queue and will let you know once we've been able to review it 🙏

solesensei:
@svlandeg I'm sorry, but could you please take a look at this PR? It's a **simple one-line fix** that shouldn't take much time.

svlandeg:
@solesensei: like I said, we have this on our queue and will look at it soon. Please refrain from pinging maintainers directly, as this makes our Github mentions very difficult to use and we don't just give priority to whoever pings the most 😉

docs_src/options/callback/tutorial005.py
```diff
@@ -0,0 +1,27 @@
+from typing import List, Optional
+
+import typer
+
+
+def names_callback(ctx: typer.Context, param: typer.CallbackParam, value: str):
```
svlandeg, line 6:
```suggestion
def names_callback(ctx: typer.Context, param: typer.CallbackParam, value: str):
```
`value` should be of type `Optional[List[str]]` here

tests/test_tutorial/test_options/test_callback/test_tutorial005_an.py
```diff
@@ -0,0 +1,56 @@
...
+    assert "Hello Camila, Victor" in result.output
+
+
+def test_2():
+    result = runner.invoke(app, ["--name", "rick", "--name", "Victor"])
+    assert result.exit_code != 0
+    assert "Invalid value for '--name': Camila must be in the list" in result.output
+
+
+def test_3():
```
svlandeg, line 29: This is the main test failing on `master`.

typer/main.py
```diff
@@ -651,7 +651,7 @@ def generate_list_convertor(
     convertor: Optional[Callable[[Any], Any]], default_value: Optional[Any]
 ) -> Callable[[Sequence[Any]], Optional[List[Any]]]:
     def internal_convertor(value: Sequence[Any]) -> Optional[List[Any]]:
-        if default_value is None and len(value) == 0:
+        if default_value is None and (value is None or len(value) == 0):
```
svlandeg, line 654: If `value` can be `None`, then the type of `value` should also be updated to `Optional[Sequence[Any]]`.

svlandeg (commented):
Thanks for the PR! I've now had time to go through it in detail.

The tests you've added clearly showcase the issue on `master` when a callback is combined with a list convertor - giving an error when `value` is `None` at L655 of `main.py`.

In the test files, both `test_tutorial005.py` and `test_tutorial005_an.py` only contain one failing test, and both represent the same behaviour. Generally, tutorial files should also be included in the Typer documentation. In this case however, I don't think that adding more explanation to the documentation adds much. As such, I suggest to drastically reduce the unit test files and only test for the edge case mentioned above.

I'll go ahead and make those changes accordingly, to get the PR in a good shape to merge.

These tutorial files also haven't been added to the Typer documentation

typer/main.py
solesensei: Sure. See you've already made the change. Thanks

svlandeg: Yep - and making this change uncovered one more related issue that I'm fixing now.

svlandeg (approved):
This should now be ready for a final review by Tiangolo!

solesensei:
Hey! 👋

Any updates on this?

solesensei:
Hi, @tiangolo! Do you have a moment to look at this PR?

PR #1028, opened 2024-10-23
Fix a few typos

kkirsche:
This merge request fixes typos in various locations in the application.

### DIFF
--- a/docs/help-typer.md
+++ b/docs/help-typer.md
@@ -147,7 +147,7 @@ Here's what to have in mind and how to review a pull request:
 
 * Don't worry too much about things like commit message styles, I will squash and merge customizing the commit manually.
 
-* Also don't worry about style rules, there are already automatized tools checking that.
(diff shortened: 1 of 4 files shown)


svlandeg (commented):
Hi, thanks for the contribution! We appreciate it.

The typo fixes in the documentation and source comments look fine. 👍 

However, I would refrain from changing `convertor` to `converter` as this is used as an argument in various public functions in `main.py`, and it would break people's code who call these functions with `convertor=x`. And while `convertor` is probably less frequently used than `converter`, I still think it's a valid term, so as such I would just keep that one across all occurrences. Can you make that edit? Thanks! 🙏

svlandeg (approved):
I've gone ahead and reverted the `convertor` changes, to avoid unnecessary breaking changes. This PR should now be good to merge 🎉

tiangolo (approved):
Great, thanks @kkirsche! :coffee: 

And thank you @svlandeg! :bow: :raised_hands:

PR #1030, opened 2024-10-26
fix: don't set excepthook unless pretty_exceptions_enable

charmoniumQ:
Previously, pretty_exceptions_enable means "set sys.excepthook to the default exception handler". This commit changes the interpretation to "simply don't change sys.excepthook at all". If it was the default handler before, it will remain the default handler. If it was a customized excepthook before, it will remain a customized excepthook.

This interpretation gives more control to users when pretty_exceptions_enable is False, while maintaining the same out-of-the-box experience for users who leave it at its default (True).

Since sys.excepthook is for uncaught exceptions, it will be read directly by humans, I don't expect this change to break any users of Typer. Most users would leave pretty_exceptions_enable to the default (True), and that behavior did not change at all in this commit.

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -70,7 +70,6 @@ def except_hook(
     if (
         standard_traceback
         or not exception_config
-        or not exception_config.pretty_exceptions_enable
     ):
         _original_except_hook(exc_type, exc_value, tb)
         return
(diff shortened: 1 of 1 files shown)


svlandeg:
Hi @charmoniumQ, thanks for the contribution! Currently this PR is failing the CI - there appear to be some failing tests. Could you look into those? It would be much more straightforward to review this PR and its proposed changes once all unit tests are green. Thanks! 🙏

PR #1035, opened 2024-10-30
✨ Update `launch` to not print anything when opening urls

patrick91:
It's not launch that does this, but the actual browsers that print text when starting, see:

image

This PR updates launch to handle URLs and swallow the output being printed by browsers (when possibile)

It now looks like this:

image


patrick91:
Just double checked and click might have this already, I remember checking though 😅

patrick91:
Actually, this might still be useful: 
CleanShot 2024-10-30 at 13 59 21@2x

patrick91:
yup, this fixes the issue on ubuntu: 
CleanShot 2024-10-30 at 15 13 28@2x

tiangolo (approved):
Awesome, thank you @patrick91! :rocket: :cake:

paketb0te:
@patrick91 just out of curiosity, which combination of python / OS / Browser did you find to be spammy?
You mention Chrome in the docstring of `launch()`, but at least on ubuntu I see no terminal output when using `webbrowser.open()` or `webbrowser.get().open()` (py3.8 to py3.13).
So I guess it is happening on macOS?

patrick91:
@paketb0te yes, it was a fresh install of ubuntu, and I think it was Chromium 😊

PR #1039, opened 2024-11-04
✨ Handle `KeyboardInterrupt` separately from other exceptions

patrick91:


tiangolo (approved):
Nice! Good idea! :rocket: :tada: 

Thanks @patrick91 :bow:

Issue #385, opened 2022-04-17
Testing Documentation w/ stderr


euan-cowie:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
runner = CliRunner(mix_stderr=False) # mix_stderr must to be set to False to access result.stderr
```

### Description

In the [Testing documentation](https://typer.tiangolo.com/tutorial/testing/) there is a Tip 🔥 Box with the following Tip:

You could also check `result.stderr` for "standard error".

This is true, but only if one sets the extra arguments on the `CliRunner` - specifically `runner = CliRunner(mix_stderr=False)`. I believe this should be made clear in the documentation, or otherwise the CliRunner can prevent the default mixing of stderr and stdout as provided by click.

### Operating System

macOS

### Typer Version

0.4.1

### Python Version

3.9.9

mrchrisadams:
For those curious, I've created a PR to merge this change in now, as it bit me last week, and I found this issue when searching for a solution.

YuriiMotov:
It has been changed.

For now we don't need `mix_stderr`. And, `result.output` contains both, `stdout` and `stderr`.
If you need, you can access them separately using `result.stdout` and `result.stderr`:

```py
import typer
from rich.console import Console
from typer.testing import CliRunner

err_console = Console(stderr=True)


app = typer.Typer()


@app.command()
def main():
    err_console.print("Here is something written to standard error")


runner = CliRunner()


def test_app():
    result = runner.invoke(app)
    assert result.exit_code == 0

    # result.output contains both
    assert "Here is something written to standard error" in result.output

    # result.stdout doesn't contain text printed to stderr
    assert "Here is something written to standard error" not in result.stdout

    # result.stderr does contain text printed to stderr
    assert "Here is something written to standard error" in result.stderr
```

PR #1045, opened 2024-11-08
📝 Clarify the need for `mix_stderr` when accessing the output of `stderr` in tests

mrchrisadams:
This pull request includes a small change to the `docs/tutorial/testing.md` file. The change updates the documentation to clarify how to check for output sent to standard error when using `CliRunner`.

Documentation update:

* [`docs/tutorial/testing.md`](diffhunk://#diff-6993c3fc60f6e1d2352cf2d6ebe40284b97c7159a5fd20a6424554568a240e45L82-R82): Clarified that the `CliRunner` instance must be created with the `mix_stderr=False` argument to access `result.stderr` in tests.

### Why would this change be necessary?

I've made this because I was left scratching my head trying to test the output of `stderr` today in a project using Typer, and I only found the solution when I came across https://github.com/fastapi/typer/issues/385, an issue that was filed in 2022.

I looked for a better `Exception` to raise, but it looks like this more a problem with Click's `Result` class than Typer; after looking into the codebase there didn't seem to be an obvious place in Typer to create a new, clearer exception class.

So, I've tried to go for some clearer text in the docs instead.

This is the _before_ version:

Screenshot 2024-11-08 at 18 44 06

This is the _after_ version:

Screenshot 2024-11-08 at 18 43 43


docs/tutorial/testing.md
```diff
@@ -79,7 +79,7 @@ Then we check that the text printed to "standard output" contains the text that
 
 /// tip
 
-You could also check `result.stderr` for "standard error" independently from "standard output" if your `CliRunner` instance is created with the `mix_stderr=False` argument.
+**Checking for output sent to standard error?** Make sure your `CliRunner` instance is created with the `mix_stderr=False` argument. You need this to be able to access `result.stderr` in tests.
```
svlandeg, line 82:
```suggestion
You could also check output sent to "standard error" (`stderr`) instead of "standard output" (`stdout`). To do so, make sure your `CliRunner` instance is created with the `mix_stderr=False` argument. You need this setting to be able to access `result.stderr` in tests.
```

svlandeg (commented):
I see your point for making this more clear in the docs. I've proposed a slightly different phrasing, including a reference to `stdout` to refer to the code snippet that is shown in the docs right before.

mrchrisadams:
hi @svlandeg I have no objections to at that change 👍

svlandeg (approved):
Thanks for the feedback @mrchrisadams! This should now be good to merge, I'll just run it by Tiangolo for final approval.

tiangolo (approved):
Cool, thank you! :coffee: :cake:

PR #1048, opened 2024-11-12
✨ Avoid the unnecessary import of `typing_extensions` in newer Python versions

horta:
Relevant links:

- https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.get_args (since Python 3.8)
- https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.get_origin (since Python 3.8)
- https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.get_type_hints (since Python 3.?)
- https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.Annotated (since Python 3.9)
- https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.Literal (since Python 3.8)

The reason for this is that typing_extensions is extensive and can cause slowdowns in compiled Python scripts.


svlandeg (commented):
Thanks for the PR! This makes sense. We do have the check in other places where we're importing things like `Literal`, though we're checking on v3.8 in those instances. From the link you shared, I see that `typing_extensions` has an updated definition for 3.8, so it makes sense to only import from `typing` from 3.9 onwards. Perhaps we should even make that change across the codebase to be consistent.

horta:
I dont understand that linting error for the `rich_utils.py` file. If I import Literal directly from typing_extensions or typing, it works, but not if I import it from _typing?

svlandeg:
Hm, that's weird. I can have a look sometime later this week if you don't figure it out!

svlandeg:
> I dont understand that linting error for the `rich_utils.py` file. If I import Literal directly from typing_extensions or typing, it works, but not if I import it from _typing?

Looks like this is expected behaviour when importing `Typing` from an indirect import 😕 : https://github.com/astral-sh/ruff/issues/128#issuecomment-1496289504

svlandeg (approved):
Should be good to merge!

tiangolo (approved):
Nice, thank you! :coffee: :cake:

PR #1037, opened 2024-11-01
✨ Add support for extending typer apps without passing a name, add commands to the top level

patrick91:

PR #1052, opened 2024-11-18
🔥 Remove auto naming of groups added via `add_typer` based on the group's callback function name

patrick91:
While working on #1037 (allowing `add_typer` to extend an app), we found that callbacks are sometimes used to implicitly get a group's name.

This would happen in cases like this:

```python
import typer

app = typer.Typer()
users_app = typer.Typer()

app.add_typer(users_app)


@users_app.callback()
def callback():
    ...

@users_app.command()
def create(name: str):
    print(f"Creating user: {name}")
```

<details>
<summary>Output for the above</summary>

```text
[14:47:30] ➜ python example.py --help

 Usage: example.py [OPTIONS] COMMAND [ARGS]...

╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --install-completion          Install completion for the current shell.                                                                                    │
│ --show-completion             Show completion for the current shell, to copy it or customize the installation.                                             │
│ --help                        Show this message and exit.                                                                                                  │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ callback                                                                                                                                               │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

This is documented here: https://typer.tiangolo.com/tutorial/subcommands/name-and-help/#inferring-name-and-help-from-appcallback

With this PR, I'm proposing to remove this implicit behavior, this would make #1037 possible and without issue when using callbacks[1]

I did a quick code search and, while unfortunately GitHub only returns [5 pages of searches](https://github.com/search?q=%22.add_typer%22+path:*.py&type=Code&l=&l=&p=5), it seems that almost no one is using this behaviour.

There's one interesting case, where they are pretty much using callbacks to do what #1037 should make it easier to implement:

https://github.com/pbdtools/xfds/blob/cf6f2a618ca8137a96eeb55eecadf6a61779d38c/src/xfds/_reset.py#L18-L19

And in cases where `name` isn't passed to `add_typer`, the `Typer` instance has a name, like here:

https://github.com/emergentmethods/flowdapt/blob/0be625437041b3b9c919f751d18969d7ce08c6b7/flowdapt/cli/db.py#L13-L18

[1] though they won't be supported in the first implementation

### DIFF
--- a/docs_src/subcommands/name_help/tutorial007.py
+++ b/docs_src/subcommands/name_help/tutorial007.py
@@ -18,7 +18,7 @@ def new_users():
     """
 
 
-app.add_typer(users_app, callback=new_users)
+app.add_typer(users_app, callback=new_users, name="users")
 
 
 @users_app.callback(help="Help from callback for users.")
--- a/docs_src/subcommands/name_help/tutorial007.py
+++ b/docs_src/subcommands/name_help/tutorial007.py
@@ -21,6 +21,7 @@ def new_users():
 app.add_typer(users_app, callback=new_users)
 
 
+# TODO: do we want this to work?
 @users_app.callback("call-users", help="Help from callback for users.")
 def users():
     """
--- a/docs_src/subcommands/name_help/tutorial007.py
+++ b/docs_src/subcommands/name_help/tutorial007.py
@@ -21,8 +21,7 @@ def new_users():
 app.add_typer(users_app, callback=new_users)
 
 
-# TODO: do we want this to work?
-@users_app.callback("call-users", help="Help from callback for users.")
+@users_app.callback(help="Help from callback for users.")
 def users():
     """
     Manage users in the app.
--- a/docs_src/subcommands/name_help/tutorial007.py
+++ b/docs_src/subcommands/name_help/tutorial007.py
@@ -9,7 +9,7 @@ def old_callback():
     """
 
 
-users_app = typer.Typer(callback=old_callback, name="exp-users", help="Explicit help.")
+users_app = typer.Typer(callback=old_callback, name="users", help="Explicit help.")
 
 
 def new_users():
@@ -18,7 +18,7 @@ def new_users():
     """
 
 
-app.add_typer(users_app, callback=new_users, name="users")
+app.add_typer(users_app, callback=new_users)
 
 
 @users_app.callback(help="Help from callback for users.")
--- a/typer/main.py
+++ b/typer/main.py
@@ -498,7 +498,7 @@ def get_group_from_info(
 
             warnings.warn(
                 "The 'callback' parameter is not supported by Typer when using `add_typer` without a name",
-                stacklevel=2,
+                stacklevel=5,
             )
     solved_info = solve_typer_info_defaults(group_info)
     (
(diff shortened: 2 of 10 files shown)


typer/main.py
```diff
@@ -390,17 +390,17 @@ def get_command(typer_instance: Typer) -> click.Command:
 
 
 def get_group_name(typer_info: TyperInfo) -> Optional[str]:
-    if typer_info.callback:
```
patrick91, line 393: If we remove this behaviour I'll update this code and the docs too :)

patrick91:
@svlandeg do you have any opinion on this? 😊

tiangolo:
Great write-up and investigation, thank you @patrick91! :bow: 

I think this would be fine, we would release it in a new 0.x version. And it would allow new features that are worth it. I'll still wait for @svlandeg's opinion in case she sees something obvious we aren't seeing yet. :sweat_smile:

docs_src/subcommands/name_help/tutorial007.py
```diff
@@ -21,6 +21,7 @@ def new_users():
 app.add_typer(users_app, callback=new_users)
 
 
+# TODO: do we want this to work?
 @users_app.callback("call-users", help="Help from callback for users.")
```
patrick91, line 25: @tiangolo I think we might want to remove support for this too 😊

svlandeg (commented):
I had a detailed look into this and it does feel sensible to remove this behaviour. Great digging by Patrick to check the current usage of this in downstream applications, which appears to be pretty limited. We'd definitely have to warn users for the breaking change though.

It feels like it would make sense to also remove the help text taken from the callback, just for consistency.

I guess one final question I have is whether there's no other way to enable #1037, so we wouldn't have to touch existing behaviour.

docs_src/subcommands/name_help/tutorial007.py
tiangolo: Yep, I think so.

tiangolo:
Thanks for the thorough review @svlandeg!

> It feels like it would make sense to also remove the help text taken from the callback, just for consistency.

I think it's fine to keep support for the help from the callback, as it would only apply when there's an actual Click Group added instead of just the commands. And it would only be supported in the cases where the callback is also supported.

> I guess one final question I have is whether there's no other way to enable https://github.com/fastapi/typer/pull/1037, so we wouldn't have to touch existing behaviour.

I was also thinking a bunch about this, but I think there wouldn't be a way, it was a design error from my side at the beginning, too few constraints on what is supported. :sweat_smile: 

Fortunately, it also didn't make much sense to many people as it's not too used. :joy: 

I think it's fine to remove it. We'll release a minor version with this (as Typer is pre-1.x that's equivalent to a major version) and add this in the breaking changes, with an example of how to upgrade. I think we are close to good to go.

@patrick91 let me know when you're ready for me to give it a final review and merge. :nerd_face:

tiangolo (approved):
Awesome! I tweaked a tiny bit the docs to remove places that mentioned the old name, and this is now ready. :rocket:

PR #1055, opened 2024-11-18
🐛 Fix printing HTML from Rich output

tiangolo:
🐛 Fix printing HTML from Rich output

It seems capturing was not working, I think it did at https://github.com/fastapi/typer/pull/847 because tests were passing, but it seems it's no longer working.

Using `io.StringIO()` seems to be the suggested way in an issue: https://github.com/Textualize/rich/issues/247#issuecomment-678699266

### DIFF
--- a/typer/rich_utils.py
+++ b/typer/rich_utils.py
@@ -1,6 +1,7 @@
 # Extracted and modified from https://github.com/ewels/rich-click
 
 import inspect
+import io
 import sys
 from collections import defaultdict
 from gettext import gettext as _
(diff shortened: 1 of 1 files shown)


bollwyvl:
It looks like not all of the whl/sdist releases of 0.13.1 went through:
 - https://pypi.org/project/typer-slim/#history
 - https://pypi.org/project/typer-cli/#history

tiangolo:
Thanks @bollwyvl! Yep, it seems there was a hiccup in the actions, I re-ran them and then it worked. :nerd_face:

PR #877, opened 2024-07-03
:sparkles: Remove Rich tags when showing completion text

svlandeg:
Upon completion, print the help text with Rich if it's available. Fixes #822.

Original behaviour, as reported in #822: hitting TAB TAB shows the autocompletion with literal Rich formatting, e.g. 
> goodbye  -- Say [bold]goodbye[/bold] to the user.

With this PR, the Rich tags are removed:
> goodbye  -- Say goodbye to the user.

In my experiments, this does remove the Rich tags from the output, though it does not yet actually format the text with the right colors etc. Not sure whether that's due to the shells/OS's I've tried - further input from others using different systems would be appreciated!

## Powershell on Windows

- [x] Shows Rich tags in completion text with `master`
- [x] Removes Rich tags in completion text with this PR
- [x] Shell is able to show Rich formatting in simple script
- [ ] Shows Rich formatting in completion text with this PR

## GitBash using `zsh` shell on Windows

I couldn't yet get this shell to work with Rich formatting - probably because it's an emulated shell.

- [x] Shows Rich tags in completion text with `master`
- [x] Removes Rich tags in completion text with this PR
- [ ] Shell is able to show Rich formatting in simple script
- [ ] Shows Rich formatting in completion text with this PR

PR #1069, opened 2024-11-29
🐛 Fix shell completions for the fish shell

goraje:
As described in this discussion #1068, recent changes aimed at sanitizing the help text from rich tags during autocompletion generation (#877)  have caused problems with proper generation of autocompletions for fish shell, when rich is installed in the environment. The main culprit most likely being the `rich.Console.print` method that converts the tabs into spaces which results in malformed autocompletes. I propose a different approach to sanitizing the help text, one that allows to perform the removal of tags before the output of `format_completion` gets returned, thus allowing to generate a rich tag-free help text, while still being able to pass it to a formatting-preserving output via `click.echo` in `typer.completion` (basically reverting this part of the source to the way it was before the #877 changes).

An example of the current behaviour while using `rich.Console.print` invoked in `typer.rich_utils.print_with_rich`:
```
testapp example1\ \ \ Just\ some\ help\ for\ example1
```

with an example app of the following structure:
```python
from typing import Annotated

import typer
from rich.markup import escape

app = typer.Typer(no_args_is_help=True, rich_markup_mode="rich")

@app.command(name="example1", help="Just [bold]some[/] help for example1")
def example1(name: Annotated[str, typer.Option(help="name")]) -> None:
    pass

@app.command(name="example2", help=f"Just some help for example2 {escape('[I'm just text in brackets]')}")
def example2() -> None:
    pass

def main() -> None:
    app()
 ```


svlandeg:
Thanks for the detailed report and PR! I'll look into this 🙏

svlandeg (commented):
Hi @goraje,

Thanks so much for this PR and the detailed report! The issue is clear and it does look like a regression for the fish shell.

In the PR #877, I should have added some more tests to check this behaviour for different shells. I've now gone in and added these tests to your PR: I've created a (minimal) version of `docs_src.commands.help.tutorial001` that has Rich tags in its subcommand descriptions, and ran completion tests in the new file `test_completion_complete_rich.py` to double check that the Rich formatting isn't there. These tests seem to be succeeding just fine on this PR. I double checked on `master`, and there the `test_completion_complete_subcommand_fish` fails because of the tab substitution you mentioned.

I saw that you also added tests to double check the functionality of the `_sanitize_help_text` function. I think that's great - I've just gone ahead and pulled that into its own separate test file for clarity.

So far the behaviour and code look fine to me, but I'll yet perform a final detailed review of all the code changes and will comment here again when I do so.

svlandeg (approved):
This looks good to me, leaving this for a final review by Tiangolo. Thanks again, @goraje!

goraje:
Thank you for looking into that :) I completely agree that the tests didn’t quite belong but I didn’t feel confident enough to modify the project’s structure by adding a test module on my own so thank you for fixing that for me :)

ndrew222:
It would be great if this can be merged soon, it's been undermining my use of the `nala` program.
Thanks so much for your good work

pilucca:
Hi, is it possible to have any update about this issue? Thanks

tiangolo (approved):
Awesome, thank you! :rocket: :tada: 

I was able to confirm the bug with the included Docker Compose and then confirm the fix the same way, using the included test file.

PR #1072, opened 2024-12-03
♻️ When printing error tracebacks with Rich, default to not showing locals, which are sometimes verbose

tiangolo:
♻️ When printing error tracebacks with Rich, default to not showing locals, which are sometimes verbose

### DIFF
--- a/typer/main.py
+++ b/typer/main.py
@@ -151,7 +151,7 @@ class Typer:
         rich_markup_mode: MarkupMode = Default(DEFAULT_MARKUP_MODE),
         rich_help_panel: Union[str, None] = Default(None),
         pretty_exceptions_enable: bool = True,
-        pretty_exceptions_show_locals: bool = True,
+        pretty_exceptions_show_locals: bool = False,
         pretty_exceptions_short: bool = True,
     ):
         self._add_completion = add_completion


docs/tutorial/exceptions.md
```diff
@@ -67,52 +63,48 @@ TypeError: can only concatenate str (not "int") to str
 
 </div>
 
-## Disable Local Variables for Security
+## Show Local Variables for Detailed Debugging
 
-If your Typer application handles **delicate information**, for example a **password**, a **key**, a **token**, then it could be problematic if the automatic errors show the value in those <abbr title="a variable that lives only inside a function, its value is only visible inside of it">local variables</abbr>.
+When using Rich, you can get more verbose output by printing the values of the <abbr title="a variable that lives only inside a function, its value is only visible inside of it">local variables</abbr> as part of the error message.
 
+By default, this setting is disabled (since Typer v0.18.0) to avoid showing **delicate information**, for example a **password**, a **key** or a **token**.
```
svlandeg: TODO: update version number here when/if we do decide to release this change

svlandeg:
I've updated this PR by changing the documentation to explain first the (new) default setting, showing no local variables. Then, showing how to enable them. Updated the tests accordingly as well.

svlandeg (commented):
In theory, this is a breaking change in behaviour, as the default case will now show less information. But it's not breaking in the sense of "applications downstream will be breaking", and we're now error'ing on the side of caution, so that's better IMO.

tiangolo:
I'll mark it as "breaking change" just in case, I don't think it will be any problem, but just to be safe. :nerd_face: 

This should be available in Typer 0.23.0, released in the next few hours. :rocket:

PR #1082, opened 2024-12-04
🔥 Remove tests for old compat code

tiangolo:
🔥 Remove tests for old compat code


tiangolo:
It seems these tests handle coverage for corner cases, like `auto_envvar_prefix`.

I'll rename them here: https://github.com/fastapi/typer/pull/1083

PR #1089, opened 2024-12-12
🐛 Fix escaping in help text when `rich` is installed but not used

svlandeg:
With PR #847, we implemented functionality to ensure that help descriptions with Rich tags (e.g. `[bold]`) would get rendered properly in Markdown (instead of just showing the "raw" brackets). This uses Rich's functionality from `console.export_html`.

Unfortunately, a problem occurred at the time because meta information like `[required]` looks like Rich tags, but with no known formatting they would just get stripped off by `console.export_html()`. That then meant we had to escape those tags at the point in the code where they're being added to the rest of the help string - this happens in `get_help_record` for both `TyperOption` and `TyperArgument` :
```
if rich is not None:
    # This is needed for when we want to export to HTML
    extra_str = rich.markup.escape(extra_str).strip()

help = f"{help}  {extra_str}" if help else f"{extra_str}"
```

Now, as pointed out in #1058, there's now an issue when the escaping happens if `rich` is installed (triggering the `if` clause in the code above) BUT the Typer app actually has `rich_markup_mode=None`. The help text will in this case show the backslash from the escaping, which is clearly unwanted behaviour.

To remedy, we need to know at the time of escaping what the value is of `rich_markup_mode`. This is not trivial, as `TyperOption` and `TyperArgument` are unaware of this information. And we can't just pass it through from the `Typer` app either, because `get_click_param()` creates empty/default `OptionInfo`/`ArgumentInfo` objects which will lead to default values downstream. I don't see a way to make things work via this route.

Instead, this PR proposes to store a custom value in `ctx.obj`. This object is being passed around anyway, and as far as I understood, `ctx.obj` may be used for these kind of application-specific settings. The code I've written tries to be super careful not to mess up any current usages of this object:

```
    if hasattr(typer_obj, "rich_markup_mode"):
        if not hasattr(ctx, "obj") or ctx.obj is None:
            ctx.ensure_object(dict)
        if isinstance(ctx.obj, dict):
            ctx.obj['TYPER_RICH_MARKUP_MODE'] = typer_obj.rich_markup_mode
```

Then, right before the escaping we can check this value and refrain from escaping if Rich is not enabled.

## WIP

Opening this in draft for now, as there's a few more things I want to check out and test before collecting feedback on this approach.

### DIFF
--- a/typer/cli.py
+++ b/typer/cli.py
@@ -301,6 +301,11 @@ def docs(
     if not typer_obj:
         typer.echo("No Typer app found", err=True)
         raise typer.Abort()
+    if hasattr(typer_obj, "rich_markup_mode"):
+        if not hasattr(ctx, "obj") or ctx.obj is None:
+            ctx.ensure_object(dict)
+        if isinstance(ctx.obj, dict):
+            ctx.obj['TYPER_RICH_MARKUP_MODE'] = typer_obj.rich_markup_mode
     click_obj = typer.main.get_command(typer_obj)
     docs = get_docs_for_click(obj=click_obj, ctx=ctx, name=name, title=title)
     clean_docs = f"{docs.strip()}\n"
--- a/tests/test_rich_markup_mode.py
+++ b/tests/test_rich_markup_mode.py
@@ -20,6 +20,7 @@ def test_rich_markup_mode_none():
     assert "Hello World" in result.stdout
 
     result = runner.invoke(app, ["--help"])
+    assert "ARG  [required]" in result.stdout
     assert all(c not in result.stdout for c in rounded)
 
 
--- a/typer/core.py
+++ b/typer/core.py
@@ -371,7 +371,10 @@ class TyperArgument(click.core.Argument):
         if extra:
             extra_str = "; ".join(extra)
             extra_str = f"[{extra_str}]"
-            if rich is not None:
+            rich_markup_mode = None
+            if hasattr(ctx, "obj") and isinstance(ctx.obj, dict):
+                rich_markup_mode = ctx.obj.get('TYPER_RICH_MARKUP_MODE', None)
+            if rich is not None and rich_markup_mode == "rich":
                 # This is needed for when we want to export to HTML
                 extra_str = rich.markup.escape(extra_str).strip()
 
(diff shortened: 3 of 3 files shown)


tests/assets/cli/multi_app_norich.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@sub_app.command()
+def bye():
+    """
+    Say bye
+    """
+
+
+app = typer.Typer(help="Demo App", epilog="The end", rich_markup_mode=None)
```
svlandeg, line 29: The point of this new test is this: checking that `rich_markup_mode=None`, with `rich` installed, still produces correct results (showing [required], default values, etc)

typer/cli.py
```diff
@@ -272,10 +273,13 @@ def get_docs_for_click(
...
-    if not has_rich:  # pragma: no cover
-        return input_text
-    return rich_utils.rich_to_html(input_text)
+def _parse_html(ctx: click.Context, input_text: str) -> str:
+    rich_markup_mode = None
+    if hasattr(ctx, "obj") and isinstance(ctx.obj, dict):
+        rich_markup_mode = ctx.obj.get("TYPER_RICH_MARKUP_MODE", None)
+    if has_rich and rich_markup_mode and rich_markup_mode == "rich":  # pragma: no cover
+        return rich_utils.rich_to_html(input_text)
+    return input_text
```
patrick91, line 282: 🤔 I think we should move the computation of `rich_markup_mode` in `get_docs_for_click` so we don't do it more than necessary (and we keep this function simple)

svlandeg: Yea fair enough, I had it there first 😁  Somehow that felt like cluttering the `get_docs_for_click` function, and conceptually it's nicer to pass around `ctx` everywhere. But I agree that it means repeating the same computation (though they should be simple, fast checks/retrievals)

fornwall:
This should also fix issues (such as `[linktext](link)` being rendered to markdown as `(link)`) when `rich_markup_mode="markdown"`, right?

typer/cli.py
svlandeg: Moved parts of the function to avoid duplicate retrieval operations.

typer/cli.py
```diff
@@ -272,10 +273,13 @@ def get_docs_for_click(
...
 
 
-def _parse_html(input_text: str) -> str:
-    if not has_rich:  # pragma: no cover
-        return input_text
-    return rich_utils.rich_to_html(input_text)
+def _parse_html(ctx: click.Context, input_text: str) -> str:
+    rich_markup_mode = None
+    if hasattr(ctx, "obj") and isinstance(ctx.obj, dict):
+        rich_markup_mode = ctx.obj.get("TYPER_RICH_MARKUP_MODE", None)
```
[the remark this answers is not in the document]
svlandeg: Done

tiangolo (approved):
Nice, thank you! 🙌 🍰 

~This will be available in Typer 0.20.2, released in the next few hours. 🎉 ~ Updating from `master` and running CI it seems coverage is failing, so this will be available later. :smiling_face_with_tear:

tiangolo:
Awesome, thanks! This will be available in Typer 0.21.1, released in the next few hours. :tada:

PR #1091, opened 2024-12-14
🔧 Add support for Python 3.13, tests in CI and add PyPI trove classifier

edgarrmondragon:


.github/workflows/test.yml
```diff
@@ -32,6 +32,8 @@ jobs:
             python-version: "3.10"
           - os: macos-latest
             python-version: "3.11"
+          - os: macos-latest
+            python-version: "3.12"
```
svlandeg, line 36:
```suggestion
          - os: windows-latest
            python-version: "3.12"
```
We alternate the OS systems to test more different combinations 😉

svlandeg (approved):
Thanks! Looks good - just making sure we test a grid of OS/Python versions.

.github/workflows/test.yml
edgarrmondragon: Nice, thanks for updating it!

tiangolo (approved):
Nice, thanks! :rocket:

PR #1096, opened 2024-12-20
💄 Add option for panels to fit to content

harkabeeparolus:
Provide an option for using Rich `Panel.fit()` aka `Panel(expand=False)` for help panels and errors. This was mentioned in discussion #607 and #680. To opt into this new, non-default behavior, just use:

```python
app = typer.Typer(rich_expand=False)
```

I think this can look nicer on very wide terminals, especially for CLIs with few parameters, or short parameter descriptions, so I would prefer to have the option to choose this style sometimes.

### Status

Ready to merge.

* The code worked fine on all the examples I tried from the Typer tutorial
* Docs and tests are finished, thanks to svlandeg!

### Example

rich_expand_example


mshonichev:
Any chances to merge this PR soon? Can someone please restart CI?

harkabeeparolus:
> Any chances to merge this PR soon?

I believe what's missing is test coverage for my new code and logic, which is not my strong suit. I would appreciate any help or suggestions on how to approach writing tests for this.

harkabeeparolus:
I updated the PR to remove merge conflicts, and:

* Fixed a bug.
* Added rich_expand to `typer.run()` for convenience. 😊

Remaining issues:

* I still need help with test coverage. 😬 I'm not familiar enough with the codebase to understand what's going on.
  * I have manually tested this code on many of the code samples in the Typer documentation, and some of my own code, with no issues so far.
  * As far as I can tell, the only failing CI tests right now are on 3.8 and 3.9, which are end of life. Should we address these? 🤨
* I have updated the text of the documentation, but I haven't made any screenshots, because I don't know what procedure you use to make them consistent with the rest of the docs.

svlandeg:
> the only failing CI tests right now are on 3.8 and 3.9, which are end of life. Should we address these?

Yes, of course, as long as the CI is running 3.8/3.9 tests, that means we're still officially supporting these versions of Python, even if they've become EOL. We'll likely drop them at some point in the near future, but nevertheless every PR should always pass the full test suite.

typer/rich_utils.py
```diff
@@ -234,7 +234,8 @@ def _get_parameter_help(
     param: Union[click.Option, click.Argument, click.Parameter],
     ctx: click.Context,
     markup_mode: MarkupMode,
-) -> Columns:
+    rich_expand: bool = True,
+) -> Table | Columns:
```
svlandeg, line 245:
```suggestion
) -> Union[Table, Column]:
```

typer/rich_utils.py
```diff
@@ -234,7 +234,8 @@ def _get_parameter_help(
     param: Union[click.Option, click.Argument, click.Parameter],
     ctx: click.Context,
     markup_mode: MarkupMode,
-) -> Columns:
+    rich_expand: bool = True,
+) -> Union[Table, Column]:
```
svlandeg:
```suggestion
) -> Union[Table, Columns]:
```

typer/rich_utils.py
```diff
@@ -457,7 +467,7 @@ class MetavarHighlighter(RegexHighlighter):
         options_table = Table(
             highlight=True,
             show_header=False,
-            expand=True,
+            expand=False,
```
svlandeg: Why is this set to `False` and not to `expand`?

harkabeeparolus: That's a very good question. I'm not sure! I wrote that change almost a year ago, but honestly, It seems to have no effect on the result, after the last bugfix I made. Let me test it some more with the current code. That change should probably be removed!

svlandeg: Thanks, I appreciate you following up after so much time! 🙈

typer/main.py
```diff
@@ -1071,8 +1081,8 @@ def wrapper(ctx: click.Context, args: List[str], incomplete: Optional[str]) -> A
     return wrapper
 
 
-def run(function: Callable[..., Any]) -> None:
-    app = Typer(add_completion=False)
+def run(function: Callable[..., Any], rich_expand: bool = True) -> None:
+    app = Typer(add_completion=False, rich_expand=rich_expand)
```
svlandeg: I see why you'd add this here, but all the other settings are also not available from the `run` command. If you want to set `rich_markup_mode` or `rich_help_panel` etc, you'd have to go through calling the actual `Typer()` constructor. I think it can be open for discussion whether or not we want that, but at the very least things should be consistent. So, I propose to drop this change from this PR and potentially discuss in follow-up work whether we want to expose some of these parameters directly to the `run` function.

harkabeeparolus: I'm fine with that.

Basically I needed it for my manual testing with a lot of different code samples from the docs, so I thought it'd be a cute idea. But I'm happy to drop it. ☺️

typer/rich_utils.py
harkabeeparolus: I'm just happy you're looking at it. 😊

typer/rich_utils.py
```diff
@@ -234,7 +234,8 @@ def _get_parameter_help(
     param: Union[click.Option, click.Argument, click.Parameter],
     ctx: click.Context,
     markup_mode: MarkupMode,
-) -> Columns:
+    rich_expand: bool = True,
```
svlandeg:
```suggestion
    rich_expand: bool,
```
There's only one place in the code base where we call `_get_parameter_help`, and there we do provide the `rich_expand` setting so we don't need this default value. It's always best to have the default value in as little different places as possible to avoid inconcistencies in future refactoring.

typer/rich_utils.py
```diff
@@ -691,7 +711,7 @@ def rich_format_help(
         console.print(Padding(Align(epilogue_text, pad=False), 1))
 
 
-def rich_format_error(self: click.ClickException) -> None:
+def rich_format_error(self: click.ClickException, expand: bool = True) -> None:
```
svlandeg:
```suggestion
def rich_format_error(self: click.ClickException, expand: bool) -> None:
```
Same reasoning about setting default values as above ^

typer/rich_utils.py
harkabeeparolus: Good point! That's cleaner.

typer/core.py
```diff
@@ -166,6 +166,7 @@ def _main(
     standalone_mode: bool = True,
     windows_expand_args: bool = True,
     rich_markup_mode: MarkupMode = DEFAULT_MARKUP_MODE,
+    rich_expand: bool = True,
```
svlandeg:
```suggestion
    rich_expand: bool,
```

typer/main.py
svlandeg: I know what you mean, I've had similar issues in the past trying to test existing code examples with different Rich markup settings. So I definitely see the value. But I'll raise this question as a separate issue internally, so we can keep it off this PR.

harkabeeparolus: Reverted this change.

svlandeg: FYI, I've briefly discussed with Tiangolo, and as expected he wants to keep `typer.run` minimal. It's only supposed to be used for quick & dirty scripts; actual proper applications should use `Typer()` directly. I'm going to work on another PR to make this more explicit in the docs and to convert most of the tutorial examples to use `Typer()` instead of `typer.run`. So, we can keep this reverted 🙂

typer/rich_utils.py
harkabeeparolus: I have restored this parameter to what's in master. It seems completely unnecessary to change this, because it makes no difference to the output, as far as I can tell after a bunch of manual tests.

harkabeeparolus:
@svlandeg thanks for your review and suggestions today! 😊

I guess my remaining points are:

* I don't really know how to write the tests to restore 100% coverage. I can try to figure it out.
* I also don't know how to test that the new behavior with `rich_expand=False` actually works. 😉
* There's no screenshot for the docs. Do we need one? How are those normally generated in this project?

svlandeg:
Hi @harkabeeparolus, thanks for the status update! I'll try to look into these issues sometime in the next 10 days.

docs/tutorial/commands/help.md
```diff
@@ -462,7 +462,7 @@ $ python main.py --help
 <font color="#A5A5A1">│ </font><font color="#A1EFE4"><b>create          </b></font> <font color="#A6E22E">Create</font> a new user. ✨                            │
 <font color="#A5A5A1">╰───────────────────────────────────────────────────────────────────╯</font>
 <font color="#A5A5A1">╭─ Utils and Configs ───────────────────────────────────────────────╮</font>
-<font color="#A5A5A1">│ </font><font color="#A1EFE4"><b>config         </b></font> <font color="#66D9EF">Configure</font> the system. 🔧                          │
+<font color="#A5A5A1">│ </font><font color="#A1EFE4"><b>config         </b></font> <font color="#66D9EF">Configure</font> the system. ⚙                          │
```
svlandeg, line 464:
```suggestion
<font color="#A5A5A1">│ </font><font color="#A1EFE4"><b>config         </b></font> <font color="#66D9EF">Configure</font> the system. 🔧                          │
```

svlandeg (approved):
This now looks good to me, though another review from someone who hasn't worked on it directly would be good 🙏 

(Also @harkabeeparolus: feel free to check and report whether the docs look good to you after my edits!)

Current docs link can be found [here](https://github.com/fastapi/typer/pull/1096#issuecomment-3520528501).

harkabeeparolus:
Yeah, the docs look great to me! The example really helps to illustrate the different look.

@svlandeg, thank you so much! You've been super helpful and I'm really happy. 😊

harkabeeparolus:
Seems like I accidentally left a `Union` type hint in there. I'll fix it.

PR #1103, opened 2024-12-31
Allow custom styles for commands in help output

TheTechromancer:
This PR makes a small tweak to the command highlighting in help output to use the global `STYLE_OPTION` instead of being hardcoded as `bold cyan`:

image

This allows users to more easily customize colors in their help output.

### DIFF
--- a/typer/rich_utils.py
+++ b/typer/rich_utils.py
@@ -488,7 +488,7 @@ def _print_commands_panel(
     # Define formatting in first column, as commands don't match highlighter
     # regex
     commands_table.add_column(
-        style="bold cyan",
+        style=STYLE_OPTION,
         no_wrap=True,
         width=cmd_len,
     )


svlandeg (commented):
Thanks for the PR! 
Good catch - that style shouldn't be hardcoded like this. 

We could indeed reuse `STYLE_OPTION` as you propose, but that would mean that the color of this first column would always be the same as the style chosen for the console's "option" theme. I think these may also be separate things?

So basically I think we could take this one step further and define a new var, something like `STYLE_COMMANDS_TABLE_FIRST_COLUMN` and set that to `"bold cyan"` to default? What do you think?

TheTechromancer:
Sounds good @svlandeg, I made the change.

svlandeg (approved):
Look good to me! I'll leave this for Tiangolo to merge.

tiangolo (approved):
Makes sense, thank you! :rocket:

PR #1116, opened 2025-01-08
💚 Fix test matrix for Python 3.7

svlandeg:
Limit the ubuntu version in the test suite to `22.04`, which still supports Python 3.7 while the current `ubuntu-latest` image doesn't anymore.


tiangolo (approved):
Oh! Cool, thank you! :rocket:

PR #1128, opened 2025-01-19
⚡️ Lazy-load `rich_utils` to reduce startup time

oefe:
Lazy-load `rich_utils` (and `rich.traceback`) to reduce import time.

Fixes #744

There was some discussion in that issue about using `lazyasd` and `if TYPE_CHECKING:` in order to make the import known to type checkers and build tools that perform import analysis.

To keep it simple, I didn't go that route.
There are already a few lazy imports, so I think this doesn't change the big picture.
And `lazyasd` appears to be no longer maintained.

Let me know if you rather would use `lazyasd` and/or `if TYPE_CHECKING:`.

## Benefit

This cuts down startup overhead by 45%. Below are some measurements (best of 5 runs).
Measured on a Mac mini (M4 pro).
 
### Before

```
~/R/typer►time uv run docs_src/first_steps/tutorial001.py
Hello World

________________________________________________________
Executed in   94.19 millis    fish           external
   usr time   71.72 millis    0.15 millis   71.57 millis
   sys time   19.54 millis    1.76 millis   17.77 millis

```

before

### After

```
~/R/typer►time uv run docs_src/first_steps/tutorial001.py
Hello World

________________________________________________________
Executed in   51.99 millis    fish           external
   usr time   37.36 millis    0.10 millis   37.26 millis
   sys time   12.77 millis    1.37 millis   11.39 millis
```

after


bonastreyair:
I am also looking forward for this quick win, CLI should be faster than it is right now...

N-Demir:
+1

stinovlas (approved):
Looks good to me (I'm not a maintainer though).

oefe:
Originally, I thought the same. While the speedup is significant, it would be difficult to test reliably in a unit test, and I didn't want to introduce a flaky test. 

But thinking about it again, I realize that it would be possible to assert that rich is not imported. This is deterministic and should be testable.

oefe:
OK, added a test

tests/test_rich_import.py
```diff
@@ -0,0 +1,21 @@
...
+        capture_output=True,
+        encoding="utf-8",
+    )
+    modules = result.stdout.splitlines()
+    modules = [
+        module
+        for module in modules
+        if module not in ACCEPTED_MODULES and module.startswith("rich")
+    ]
+    assert not modules
```
svlandeg, line 21: This is a matter of personal preference, but for clarity of code I would prefer 
```suggestion
    assert len(modules) == 0
```

svlandeg (commented):
Thanks for the clear analysis and PR proposal @oefe!

I can definitely see the appeal of doing the lazy load to improve startup time when `rich` isn't used. It clutters the code a little bit, but the trade-off may be worth it.

One question I have is for when you do have Rich installed. Do the multiple `import` statements at different points in the code base result in an efficiency loss somehow? Because I can imagine that in the background, it is being checked what is already imported and what isn't. I'm not sure if that check leaves a performance trace or not.

tests/test_rich_import.py
oefe: This is arguably the most idiomatic style, see also [PEP8](https://peps.python.org/pep-0008/#programming-recommendations):

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false:

```python
# Correct:
if not seq:
if seq:
# Wrong:
if len(seq):
if not len(seq):
```

But I can change it if you prefer.

oefe:
The performance impact is negligible. Essentially, it boils down to one dictionary lookup (to test whether the module is already in `says.modules`) and one dictionary insertion (add the module to the local namespace). These are both very quick. 

Also, most of the affected lines are executed at most once per run.

tests/test_rich_import.py
svlandeg: I'll leave the final decision with Tiangolo 🙏 

UPDATE: Note that I'm not suggesting to check truthness of `len(seq)` as in your code above, I'm suggesting to check `len(seq) == 0`.

svlandeg (approved):
> The performance impact is negligible.

Thanks for double checking!

This PR looks good to merge to me, I'll leave it to Tiangolo for a final review.

onyx-and-iris:
I just tested these changes, it makes a noticeable difference for my CLI.

https://gist.github.com/onyx-and-iris/987ad066cb9baa03b58f5f446f7ea837

Thank you for investigating this issue.

tests/test_rich_import.py
basnijholt: There are even pylint / Ruff rules that enforce this behavior: https://docs.astral.sh/ruff/rules/len-test/

`assert not modules` is the right thing.

basnijholt (approved):
Awesome! Came here after I found out that Typer is the one package that is the slowest part of ([my CLI](https://github.com/basnijholt/agent-cli)).

tests/test_rich_import.py
svlandeg: I'm not reading this with the same interpretation. I'm not suggesting to check truthness of `len(x)`, I'm suggesting to explicitely check the length being zero, as suggested on the page you cited:
> You can either remove the call to len or **compare the length against a scalar**.

basnijholt, line 21: Fair point!

```suggestion
    assert not any(
        module in ACCEPTED_MODULES and module.startswith("rich")
        for module in modules
    )
```
is even better I guess.

oefe: These are all good suggestions, however in addition to the readability of the code, we should also consider the readability of the error messages.
This is important for example if someone introduces a regression by inadvertently importing `rich` again.
They should see right away which module caused the regression.

If an assertion fails, `pytest` does a good job printing meaningful error messages based on the condition. But this works only with relatively simple conditions.

To demonstrate this, I changed ACCEPTED_MODULES to an empty set to cause a test failure. With `assert not modules`, I get:

```
E       AssertionError: assert not ['rich._extension', 'rich']
```

This tells me immediately which modules were unexpectedly imported, and thus helps finding the cause.

With `assert len(modules) == 0`, I get:

```
E       AssertionError: assert 2 == 0
E        +  where 2 = len(['rich._extension', 'rich'])
```

This is a bit more lengthy and indirect, but I can still see which modules were unexpectedly imported.

With `assert not any (...)`, I get:

```
E       assert not True
E        +  where True = any(<generator object test_rich_not_imported_unnecessary_any.<locals>.<genexpr> at 0x101803510>)
```

This is rather cryptic and unhelpful.

I therefore prefer the first variant. I find the second variant (with `len`) acceptable too, but it wouldn't be my choice. I don't like the third variant because of the cryptic error message.

However, in the end, this is @tiangolo's project, so I will use whatever he prefers.

basnijholt:
@tiangolo, it seems like this PR is [blocked by your review](https://github.com/fastapi/typer/pull/1128#discussion_r2175316844), would you mind giving it a few minutes of your time?

In [my application `agent-cli`](https://github.com/basnijholt/agent-cli), startup time is extremely important and typer is responsible for 60% of all of it. This few line change will solve most of the problem!

image

chrisgoddard:
+1 on this

N-Demir:
+1 !!!

tiangolo (approved):
Awesome! Thank you @oefe 🙌 

In particular, for the detailed description, benchmarks, and explaining the train of thought in all the points.

Thanks all for the help!

And thanks @svlandeg for the help maintaining as always. 😄 

---

This will be available in Typer 0.17.0 in the next few hours. 🎉

dwreeves:
@tiangolo There is still one additional performance improvement missing, not on CLI execution but for CLI help text rendering.

Specifically-- lazy-loading `from rich.markdown import Markdown` significantly improve performance of help text rendering for all users not using markdown because the downstream `markdown_it` import takes a ton of time. Most users don't set help text rendering to `"markdown"` mode so for most users this is a major performance improvement for `--help`.

There are many libraries in the wild which take a long time to load their CLIs (for reasons unrelated to Typer) and so every fraction of a second shaved off can make things feel snappier basically for free.

svlandeg:
@dwreeves: thanks for the note. Please start a new discussion for this, as this PR has been merged and closed. When opening the discussion, it'd be great if you can quantify the improvement you get from lazy loading `rich.markdown`.

PR #1134, opened 2025-01-24
Fix arg parameter of autocompletion

bckohan:
Separated from #1006 

Bug documented in #1133 

One note on the implementation. Because the CLI options during an autocompletion run are passed through shell-specific mechanisms (usually environment variables) and are not available in sys.args we have to resort to adding this information to the context in the completer classes. Click provides a mechanism to do this using ctx_args and the "obj" parameter. Obj is seldom used, but it can be anything per the click docs. Right now I do not believe the click code path sets it to anything. This PR sets it to a dictionary if it isn't already and if it is dictionary it sets the args key to be the arguments passed by the shell specific mechanism. This would be brittle to obj being set somewhere upstream to anything but a dictionary. I think this is unlikely to happen, but if it does it will not brick the code, it'll just revert to the current behavior (empty list) and the CI will fail.

### DIFF
--- a/tests/test_tutorial/test_options_autocompletion/test_tutorial008.py
+++ b/tests/test_tutorial/test_options_autocompletion/test_tutorial008.py
@@ -23,7 +23,7 @@ def test_completion():
     assert '"Camila":"The reader of books."' in result.stdout
     assert '"Carlos":"The writer of scripts."' in result.stdout
     assert '"Sebastian":"The type hints guy."' in result.stdout
-    assert "[]" in result.stderr
+    assert "--name" in result.stderr
 
 
 def test_1():
--- a/tests/test_tutorial/test_options_autocompletion/test_tutorial008_an.py
+++ b/tests/test_tutorial/test_options_autocompletion/test_tutorial008_an.py
@@ -23,7 +23,7 @@ def test_completion():
     assert '"Camila":"The reader of books."' in result.stdout
     assert '"Carlos":"The writer of scripts."' in result.stdout
     assert '"Sebastian":"The type hints guy."' in result.stdout
-    assert "[]" in result.stderr
+    assert "--name" in result.stderr
 
 
 def test_1():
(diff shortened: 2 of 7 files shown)


svlandeg:
Related comment with links to the relevant code places: https://github.com/fastapi/typer/pull/1006#issuecomment-2613071848

tests/test_tutorial/test_options_autocompletion/test_tutorial008.py
```diff
@@ -23,7 +23,7 @@ def test_completion():
     assert '"Camila":"The reader of books."' in result.stdout
     assert '"Carlos":"The writer of scripts."' in result.stdout
     assert '"Sebastian":"The type hints guy."' in result.stdout
-    assert "[]" in result.stderr
+    assert "--name" in result.stderr
```
svlandeg, line 26:
```suggestion
    assert "['--name']" in result.stderr
```

tests/test_tutorial/test_options_autocompletion/test_tutorial008_an.py
```diff
@@ -23,7 +23,7 @@ def test_completion():
     assert '"Camila":"The reader of books."' in result.stdout
     assert '"Carlos":"The writer of scripts."' in result.stdout
     assert '"Sebastian":"The type hints guy."' in result.stdout
-    assert "[]" in result.stderr
+    assert "--name" in result.stderr
```
svlandeg, line 26:
```suggestion
    assert "['--name']" in result.stderr
```

svlandeg (approved):
Thanks again for the contribution @bckohan, and for your patience while we're moving these different PRs to the review process! 🙏 

I agree that this behaviour is not what is expected/documented, and the change in unit tests reflects the correct behaviour. 

While I'm approving this PR, I will request an additional review from Tiangolo to get his feedback on the change, as he might have more information on why this was implemented originally the way it was.

PR #1138, opened 2025-01-30
🐛 Ensure that autocompletion works for `Path` arguments/options

### LINKED ISSUE #951: Autocompletion for `Path`
### Privileged issue

### Issue Content

It seems autocompletion for `Path` types is not working correctly.

```Python
from pathlib import Path

import typer

app = typer.Typer()


@app.command()
def f(p: Path):
    print(p)


if __name__ == "__main__":
    app()
```

Running that, for example with Typer CLI:

```console
$ typer demo.py run <TAB>
```

that should show the available files in the current directory. In Zsh it doesn't show the files, in Bash it shows the files.

Then, let's say it's run from the same Typer directory, where the `LICENSE` file is located:

```console
$ typer demo.py run L<TAB>
```

That should autocomplete `LICENSE`, but it only adds `TAB`/spaces characters.

svlandeg:
Resolves #951

## Issue

The issue on `master` is that parameters (either an `Argument` or an `Option`) specifically typed as `Path` wouldn't be automatically completed on the commandline. For instance, having two files `test_1.py` and `test_2.py` in a directory, and typing `"t"` for an option typed as `Path`, would simply put a space after the `"t"`, not autocompleting at all.

The desired behaviour instead would be to first autocomplete the `"test_"` part, and then after another TAB show the two options `"test_1.py"` and `"test_2.py"`.

## Proposed solution

After quite a bit of digging and reading @tiangolo's elaborate background information on issue #951, I've come to the conclusion that this unexpected behaviour is likely caused by an unwanted and inconsistent interplay between Typer and Click.

Stepping back for a second - if you would type the `Argument` or `Option` as `str`, ironically then the autocompletion would actually work out of the box. Things go haywire (IMO) in the `get_click_type` function where the parameter is converted to the corresponding `click.Path` when a type annotation of `pathlib.Path` is found. This [`click.Path`](https://github.com/pallets/click/blob/8.1.8/src/click/types.py#L768) class has a specific [`shell_complete`](https://github.com/pallets/click/blob/8.1.8/src/click/types.py#L934C9-L934C23) function that:

> Return a special completion marker that tells the completion system to use the shell to provide path completions for only directories or any paths.

But Typer's autocompletion functionality doesn't use that marker like Click's scripts (e.g. the [Bash one](https://github.com/pallets/click/blob/8.1.8/src/click/shell_completion.py#L93)) does. If instead we return the empty list (i.e. the default behaviour of `click.ParamType`), then it all seems to work again.

## Testing

I could confirm the original bug as described in #951 on Bash and Zsh, and I could confirm that the autocompletion functionality works again with `Path` parameters with this code change.

If this is the route forward we want to take, I'll test on more systems.

However, I'm entirely unsure how to actually test this in the test suite, given the dependency on the actual console for completion.

## Open questions

Not sure whether this is the kind of fix/resolution that @tiangolo has in mind, so I'm writing this up to solicit feedback. Also the naming of "TyperPath" is probably not ideal - open to suggestions 😁


svlandeg:
The coverage test is failing, because there are no autocompletion tests with `Path` parameters. I can add one, but I'd rather add one that actually exhibits the wrong behaviour on `master`, and this is tricky... 🤔

bckohan:
You might have a look at [how I do this in django-typer](https://github.com/django-commons/django-typer/tree/v3/tests/shellcompletion).

I run the shells in a subprocess in a pseudo terminal. This allows me to test end-to-end like you want here. There are a few downsides though:

1). I haven't been able to get this to work with fish yet (when I run it in a pty it decides its not an interactive shell and terminates prematurely - I'm sure theres a way around it I just havent found it yet). It does work with powershell, bash and zsh.
2) There doesn't seem to be a way to avoid sleeps to handle timing issues in tests. This can make the tests take a bit of time if you run a lot of them.

tiangolo (commented):
Amazing! Beautiful job! :clap: 

The investigation and report make me understand what's happening.

And I like this is the minimal change to fix the problem. :sunglasses: 

I consider finding short and simple (but correct) solutions tend to be much harder than finding more complicated ones, so this is great. :brain: 

I think this could go in, we make a release with the fix, and then later we could refactor Typer's completion logic having all the rest in mind, maybe having the equivalent of the new logic in Click but written in Python... but this would fix the bug right away so we can start with this which solves the most important/urgent part.

About the name `TyperPath` I didn't find a better alternative, there's already other `Typer<X>` classes like `TyperArgument` and `TyperGroup`, so it seems fine.

The other words I was playing with were `Param` or `Type` (as Click's `Path` inherits from `ParamType`), but I don't think it's any clearer, so I think `TyperPath` is fine. :nerd_face:

svlandeg (commented):
I've fixed the coverage issue by adding a test, though this test does not (yet) replicate the bug on `master`. We may want to think about having a more extended test suite of end-to-end tests for the autocompletion behaviour in different shells, but we could defer that work to follow-up PRs to avoid blocking the release of this fix.

tiangolo (approved):
Nice, thank you! :rocket:

MarshalX:
Hi! Could you please tell me when a new release of Typer is planned that will include this PR?

tiangolo:
This is available now in Typer 0.15.3, just released :rocket:

PR #1144, opened 2025-02-06
Add click.exception to __init__.py

VikalpRusia:
Added more exceptions from click.exception to typer.__init__ so that it has all exceptions.

### DIFF
--- a/typer/__init__.py
+++ b/typer/__init__.py
@@ -5,8 +5,14 @@ __version__ = "0.15.1"
 from shutil import get_terminal_size as get_terminal_size
 
 from click.exceptions import Abort as Abort
+from click.exceptions import BadArgumentUsage as BadArgumentUsage
+from click.exceptions import BadOptionUsage as BadOptionUsage
 from click.exceptions import BadParameter as BadParameter
-from click.exceptions import Exit as Exit
+from click.exceptions import ClickException as ClickException
+from click.exceptions import FileError as FileError
+from click.exceptions import MissingParameter as MissingParameter
+from click.exceptions import NoSuchOption as NoSuchOption
+from click.exceptions import UsageError as UsageError
 from click.termui import clear as clear
 from click.termui import confirm as confirm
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Could you explain more why we'd need this?

VikalpRusia:
> Could you explain more why we'd need this?

As there are few exceptions which are in __init__ file while others are not there.

What I am suggesting is to have all or have None for uniformity

VikalpRusia:
> > > Could you explain more why we'd need this?
> > 
> > 
> > As there are few exceptions which are in **init** file while others are not there.
> > What I am suggesting is to have all or have None for uniformity
> 
> Well, the ones that are currently in there are there because they are being used in Typer's core codebase, e.g. to `raise typer.Abort()`. So we definitely can't remove those.
> 
> Conversely, I'm not sure it adds anything to import exceptions that are not used.

Won't it be better to import from typer instead of click?

tiangolo:
Thanks for the interest @VikalpRusia, I don't expect users to capture and use those exceptions in their code, so I don't think there's a situation where it would make sense to import them from Typer. Given that I'll pass on this one. Thanks! :coffee:

VikalpRusia:
@tiangolo , @svlandeg 
I am using these exceptions for few exceptions

```from click.exceptions import MissingParameter, FileError, UsageError```

Could you provide me alterantives to use similar/same exceptions using typer?
Accordingly I will refactor my code.

Thanks in advance

svlandeg:
Hi @VikalpRusia, there's no need to refactor your code - you can simply import these from Click directy as you're already doing.

VikalpRusia:
> Thanks for the interest @VikalpRusia, I don't expect users to capture and use those exceptions in their code, so I don't think there's a situation where it would make sense to import them from Typer. Given that I'll pass on this one. Thanks! ☕

@svlandeg 
but @tiangolo mentioned this, he isn't expecting to use them in their code?
Could you explain it more what does it mean then?

PR #1169, opened 2025-03-01
🔨  Add --showlocals to test.sh

rickwporter:
When tests fail, this provides more debug information. Having the extra debug is particularly useful when working in different  environments than your local setup (e.g. different Python versions in the CI pipeline).


tiangolo (approved):
Thanks! 🚀

PR #1183, opened 2025-03-20
✨ Allow annotated parsing with a subclass of `Path`

emfdavid:
Please allow using custom Path subclasses such as [UPath](https://github.com/fsspec/universal_pathlib/tree/main)

Here is an example
```python
import logging
from pathlib import Path
from typing import Annotated, Optional

import typer
from upath import UPath
app = typer.Typer()

@app.command()
def path_func(arg: Annotated[UPath, typer.Argument(parser=UPath)]):
    logger.info(f"got arg: '{arg}'; type: {type(arg)}")

if __name__ == "__main__":
    logging.basicConfig(level=logging.INFO)
    logger = logging.getLogger(__name__)
    app()
```

But as you can see in my debugger, the parsed value will get converted to a Path by the param_path_convertor.

Screenshot 2025-03-20 at 2 18 16 PM

Allowing that function to return anything that is already a Path (or maybe a PurePath?) would fix this issue, though, it maybe a problem for other types as well?

I can hack around this by setting the annotated type to None but using UPath as a parser, however that does not seem to be documented behavior or good practice.
```python
@app.command()
def path_func(arg: Annotated[None, typer.Argument(parser=UPath)]):
    logger.info(f"got arg: '{arg}'; type: {type(arg)}")
```

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -21,7 +21,7 @@ jobs:
     outputs:
       docs: ${{ steps.filter.outputs.docs }}
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
     # For pull requests it's not necessary to checkout the code but for the main branch it is
     - uses: dorny/paths-filter@v3
       id: filter
@@ -52,7 +52,7 @@ jobs:
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
         run: echo "$GITHUB_CONTEXT"
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -23,7 +23,7 @@ jobs:
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
         run: echo "$GITHUB_CONTEXT"
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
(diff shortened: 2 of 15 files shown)


svlandeg (commented):
Hi, thanks for the contribution and feature request!

What you're proposing makes sense to me, and at a first glance the code looks good. Surprisingly, some of the CI tests do fail. Do you have time to look into those?

emfdavid:
Well - [that](https://discuss.python.org/t/support-subclassing-pathlib-path/29095) was a rabbit hole.
Let's see if my hack works for older versions of python... if not can we just skip this test for older versions?

svlandeg:
> @svlandeg what happens now?

@emfdavid: we've put this on our internal reviewing board, but we're currently facing quite a bit of a backlog in PRs (also on `fastapi` etc), so it may take a while before we make a final decision. We'll keep you up-to-date here!

emfdavid:
> @emfdavid: we've put this on our internal reviewing board, but we're currently facing quite a bit of a backlog in PRs (also on `fastapi` etc), so it may take a while before we make a final decision. We'll keep you up-to-date here!

Thank you for the update - I have monkey patched my applications for now.

delsner:
@svlandeg any idea when this will land? Thanks!

svlandeg:
@delsner & @emfdavid : I've been engaged on other projects for the last few months, but I'll try and review this in detail this week! Thanks for your patience 🙏

svlandeg (approved):
This is a clean fix that supports subclassing `Path` cleanly. I verified that the test fails on `master` and otherwise everything looks good to me.

I'll leave this for Tiangolo for a final check.

Issue #536, opened 2023-01-18
Support for bytes in Options and Arguments


AlexanderPodorov:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer


def main(name: bytes):
    typer.echo(name)


if  __name__ == '__main__':
    typer.run(main)
```

### Description

It would be nice if we could add support for bytes in order to use it in contexts where bytes are expected (e.g. base64 encode/decode, password hashing, etc.)

### Wanted Solution

Support for bytes type. Currently `typer` says:
`RuntimeError: Type not yet supported: <class 'bytes'>`

### Wanted Code

```python
import typer


def main(name: bytes):
    typer.echo(name)


if  __name__ == '__main__':
    typer.run(main)
```

### Alternatives

Always use encode/decode methods.

### Operating System

macOS

### Typer Version

0.7.0

### Python Version

3.11.1

marcosschroh:
Hi,

is there any news  on this issue?

ESPR3SS0:
Hello, I could look into adding this! 

However just so you know, custom types can be made easily. A type for bytes could be made like:
```python
import typer
from typing_extensions import Annotated
app = typer.Typer()
class Bytes(self, value):
    def __init__(self, value):
        self.value = value.encode()
def parse_bytes(val):
    return Bytes(val).value

def func(input: Annotated[Bytes, typer.Argument(parser=parse_bytes)]):
    ....
    ....
    return
```

The greatest hassle with this solution is remembering to add `parser=parse_bytes' to all the Bytes arguments. So, could be worth natively supporting it.

JPHutchins:
This would also be useful for my use cases (firmware / transport protocols).  Though, instead of supporting "bytes" directly, wouldn't we need to support "encodings"?  

For example, a type specified as:

- utf-8 would take a string end encode it to utf-8 bytes
- base64 would encode base64 bytes
- hex would use bytes from_hex (or whatever that method is called
- etc

[...]

PR #1190, opened 2025-04-03
✨ Support bytes in Options and Arguments

doubledare704:
- Added support for bytes type in Typer by:
- Creating a custom BytesParamType class in typer/main.py
- Modifying the get_click_type function to handle bytes type
- Created comprehensive tests:
    - Basic tests for bytes type in tests/test_bytes_type.py
    - Advanced tests for base64 and hex encoding/decoding in tests/test_bytes_encoding.py
- Created example applications:
    - Simple bytes example in examples/bytes_type_example.py
    - Comprehensive encoding/decoding example in examples/bytes_encoding_example.py

Attempt to solve https://github.com/fastapi/typer/issues/536


svlandeg:
Thanks for the PR, appreciate the time and effort spent on writing tests as well!

We'll get this reviewed by the team and will get back to you 🙏

YuriiMotov (commented):
@doubledare704, thank you for your interest and efforts!

As I see it, the main reason to introduce `bytes` type support is to have the ability to use the same function outside the Typer, right?
Otherwise I don't see a point in this.. We can just accept `str` and decode it in one line. That would even be easier in cases where we want to let users specify the encoding as a CLI option

This implementation assumes that the argument value is encoded with system default encoding. It doesn't allow to specify the encoding.
This implementation assumes that the argument value is encoded with the system default encoding. It doesn’t allow specifying the encoding.
I believe we should at least allow specifying the encoding as a parameter to `Argument`/`Option`.

But first, I would wait for Sebastian to approve the idea

saky-semicolon (commented):
A few improvement opportunities:

Consider adding more edge case tests to further validate error handling.

In convert(), extracting decoding/encoding logic into helper functions could improve readability and reuse.

Documentation could be expanded with usage examples to make it easier for new users to try the features.

doubledare704:
Summary of bytes support changes and request for approval

Primary use case:
- Allow the same function to work both inside Typer (CLI parsing) and outside Typer (regular Python code) with bytes parameters.

API and implementation:
- Reused existing ParameterInfo.encoding and ParameterInfo.errors to control how CLI inputs are encoded to bytes.
- Implemented encoding-aware BytesParamType(encoding=None, errors=None) with small helper methods for readability.
- Wiring: get_click_type(annotation is bytes) now returns a configured BytesParamType based on the Argument/Option’s encoding/errors. Defaults are UTF-8 + "strict", matching Python’s str.encode() behavior.
- Backward compatibility: unchanged when encoding/errors aren’t provided; typing bytes still works as before.

Tests:
- Added edge cases: non-ASCII defaults (UTF-8), latin-1 for Argument/Option, errors="replace" with ascii, invalid encoding name error messaging.
- Full test suite passes locally.

Docs/examples:
- New docs page: Tutorial → CLI Parameter Types → Bytes, with examples from docs_src.
- Updated examples to demonstrate default UTF-8, latin-1, ascii+replace.

Questions:
- Please confirm the approach of reusing ParameterInfo.encoding/errors for bytes and the default policy (UTF-8, "strict") is acceptable.
- If you prefer a different default or API shape, I’m happy to adjust.

Thanks!

tests/test_bytes_encoding.py
```diff
@@ -0,0 +1,160 @@
+import base64
```
svlandeg, line 1: As a quick comment, if we do decide to progress with this PR, we should add the correct tests for the tutorial files in `test_tutorial/test_parameter_types/test_bytes`.

docs/tutorial/parameter-types/bytes.md
```diff
@@ -0,0 +1,38 @@
+# Bytes
```
svlandeg, line 1: This page should also be added to `mkdocs.yml`

doubledare704: seems like already has a link in mkdocs
mkdocs.yml:118

tests/test_bytes_encoding.py
doubledare704: I've just added tests for tutorials. Please check.

svlandeg (commented):
@doubledare704: as Yurii said, we'll wait for further reviews until Tiangolo approves the idea. We have got this on our internal board, so no need for further pings 🙏

PR #1222, opened 2025-05-13
⬆️ Add compatibility with Click 8.2

tiangolo:
⬆️ Add compatibility with Click 8.2


thewhaleking:
Also, the [docs](https://typer.tiangolo.com/tutorial/testing/) reference using `mix_stderr=False` for `CliRunner`, but this is not longer available in Click 8.2.0

typer/core.py
```diff
@@ -375,15 +375,25 @@ def get_help_record(self, ctx: click.Context) -> Optional[Tuple[str, str]]:
...
+        # TODO: When Click < 8.2, remove this
+        signature = inspect.signature(self.type.get_metavar)
+        if "ctx" in signature.parameters:
+            # Click >= 8.2
+            type_var = self.type.get_metavar(self, ctx=ctx)  # type: ignore[arg-type]
+        else:
+            # Click < 8.2
+            type_var = self.type.get_metavar(self)  # type: ignore[call-arg]
+        # TODO: /When Click < 8.2, remove this, uncomment the line below
+        # type_var = self.type.get_metavar(self, ctx=use_ctx)
```
lord-haffi, line 396: Did you mean
```suggestion
        # TODO: When deprecating Click < 8.2, remove this
        signature = inspect.signature(self.type.get_metavar)
        if "ctx" in signature.parameters:
            # Click >= 8.2
            type_var = self.type.get_metavar(self, ctx=ctx)  # type: ignore[arg-type]
        else:
            # Click < 8.2
            type_var = self.type.get_metavar(self)  # type: ignore[call-arg]
        # TODO: /When deprecating Click < 8.2, remove this, uncomment the line below
        # type_var = self.type.get_metavar(self, ctx=ctx)
```
?

davidism:
I haven't looked over this PR, but we did just add a new doc page about supporting multiple versions in Click, with an example for the `get_metavar` signature change. https://click.palletsprojects.com/en/stable/support-multiple-versions/ If there are other parts that you think we should document, or need help figuring out compatibility with, please let us (Click maintainers) know.

tiangolo:
Thanks a lot for offering help @davidism and especially for writing those docs! :raised_hands: 

---

I'm currently debugging the output showing errors for required parameters, with the update it doesn't show what was the (missing) required parameter, I probably have something wrong here... still debugging. :nerd_face:

PR #1240, opened 2025-06-05
🐛 Ensure exit code 0 when `no_args_is_help` shows help

Fabian665:
When `no_args_is_help` is True and no arguments are provided to a Typer application, the help message is displayed. However, an additional empty error panel also currently appears, and the application exits with status code 2. This issue affects applications using Click version 8.2.0 or later.

The root cause is a change in Click (pallets/click@d8763b93) where `Command.parse_args` now raises `click.exceptions.NoArgsIsHelpError` when `no_args_is_help` is true and no arguments are given. Previously, Click would print help and call `ctx.exit()` directly within `parse_args`.

Typer's generic `except click.ClickException as e:` block in `core.py._main` catches this `NoArgsIsHelpError`. The subsequent call to `rich_utils.rich_format_error(e)` (or `e.show()` if Rich is not used) results in the help message being displayed. This is because `NoArgsIsHelpError` is a subclass of `click.UsageError`, and `UsageError.show()` prints `ctx.get_help()`. However, this process also leads to an attempt to format a non-existent specific error message (as `NoArgsIsHelpError.format_message()` is empty), causing the blank error panel. Finally, `sys.exit(e.exit_code)` uses the `UsageError`'s exit code of 2.

This commit addresses the issue by checking if the `click.ClickException` is of instance `NoArgsIsHelpError` and then raising `click.exceptions.Exit(0)`

This ensures that after the help message is displayed, the application exits cleanly with status code 0 and without displaying the erroneous empty error panel, aligning with the intended behavior of `no_args_is_help`.

### DIFF
--- a/typer/core.py
+++ b/typer/core.py
@@ -210,7 +210,7 @@ def _main(
             raise click.exceptions.Exit(130) from e
         # TODO: When deprecating Click < 8.2 uncomment the next two lines
         # except click.exceptions.NoArgsIsHelpError as e:
-            # raise click.exceptions.Exit(0) from e
+        # raise click.exceptions.Exit(0) from e
         except click.ClickException as e:
             # TODO: When deprecating Click < 8.2 remove this section [start]
             _no_args_is_help_error = getattr(


DHUKK:
Should we also update the test that was changed [here](https://github.com/fastapi/typer/pull/1222/files#diff-fc17c61ebf30b2e29e0cdc1fcc2045f18214624f7fca68023ed73969dd177ef6R13-R17) `tests/test_tutorial/test_commands/test_index/test_tutorial003.py - test_no_arg`


to add back the 
```python
assert result.exit_code == 0
```

Fabian665:
@DHUKK Thanks, I have restored the test. 

I also added commented code for dropping Click<8.2 support.

SeedyROM:
I'd love this to be merged, I think what's breaking the pipeline is that this issue is missing a label? 

For now I'm just doing something akin to this!

First remove `no_args_is_help` from the options, then add this:
```python
@app.callback(invoke_without_command=True)
def main(ctx: typer.Context):
    """My awesome CLI tool"""
    if ctx.invoked_subcommand is None:
        print(ctx.get_help())
        raise typer.Exit()
```

Fabian665:
@SeedyROM It doesn't let me put a label, I guess only maintainers can do so.

typer/core.py
```diff
@@ -208,7 +208,19 @@ def _main(
             raise click.Abort() from e
         except KeyboardInterrupt as e:
             raise click.exceptions.Exit(130) from e
+        # TODO: When deprecating Click < 8.2 uncomment the next two lines
+        # except click.exceptions.NoArgsIsHelpError as e:
+        # raise click.exceptions.Exit(0) from e
         except click.ClickException as e:
+            # TODO: When deprecating Click < 8.2 remove this section [start]
+            _no_args_is_help_error = getattr(
```
diablodale, line 216: Rather than probing attrs and instance, I suggest using the exception we want to catch.
Click's __init__.py doesn't pull in the `NoArgsIsHelpError` exception into itself. So instead, we can easily do the same at the top of this file with;

```python
from click.exceptions import NoArgsIsHelpError
```

and then here simply the code to be

```python
...
except KeyboardInterrupt as e:
    raise click.exceptions.Exit(130) from e
except NoArgsIsHelpError as e:
    raise click.exceptions.Exit(0) from e
except click.ClickException as e:
...
```

tests/test_tutorial/test_commands/test_index/test_tutorial003.py
```diff
@@ -12,6 +12,7 @@
 
 def test_no_arg():
     result = runner.invoke(app)
+    assert result.exit_code == 0
     assert "[OPTIONS] COMMAND [ARGS]..." in result.output
     assert "Commands" in result.output
     assert "create" in result.output
```
ncoghlan, line 18: It would also be good to add a `assert not result.stderr` to ensure the doubled output isn't displayed

typer/core.py
ncoghlan: To continue supporting older versions of click (which don't define the exception), the exception clause can be made conditional with code like:

```python
CatchClickNoArgsIsHelpError: Exception|tuple[()]
try:
    # click >= 8.2.0
    CatchClickNoArgsIsHelpError = click.exceptions.NoArgsIsHelpError
except AttributeError:
    # click < 8.2.0
    CatchClickNoArgsIsHelpError = ()
```

And then:
```python
    except CatchClickNoArgsIsHelpError as e:
        raise click.exceptions.Exit(0) from e
```

Empty tuples are permitted in except clauses, and mean that the clause will never actually trigger at runtime.

adam-moss:
> I'd love this to be merged, I think what's breaking the pipeline is that this issue is missing a label?
> 
> For now I'm just doing something akin to this!
> 
> First remove `no_args_is_help` from the options, then add this:
> 
> ```python
> @app.callback(invoke_without_command=True)
> def main(ctx: typer.Context):
>     """My awesome CLI tool"""
>     if ctx.invoked_subcommand is None:
>         print(ctx.get_help())
>         raise typer.Exit()
> ```

This works great for the app object, does anyone have a solution for the command object?  Since ctx.invoked_subcommand is not none in that scenario then you get an error about missing params rather than the help text (assuming the command has required params, of course!).

ruben-mena:
Hello everyone, I was wondering if there were any updates regarding this issue?

svlandeg:
Hi all!

Thanks for all the work on this @Fabian665 and others for chiming in.

I've been a few months MIA on this repo and have started going through the backlog. While doing so, came across #1262 which focuses on resolving the additional output printed to the console. I ended up making a clean version of this in #1278 and this has been released as Typer v0.16.1. Unfortunately at the time I didn't realise this PR is super related, otherwise I would have dealt with them simultaneously - apologies!

Now, the remaining issue here is whether or not the exit code should be 0 or 2, and on Click's github there's been quite some discussion about this as well. From Sebastián's edit [here](https://github.com/fastapi/typer/pull/1222/files#diff-fc17c61ebf30b2e29e0cdc1fcc2045f18214624f7fca68023ed73969dd177ef6R13-R17) I gather that he prefers to follow the breaking change Click introduced, and in a way I can follow the reasoning that just typing the app's main command and no subcommand is a user error, EVEN if we're friendly and displaying the help text. 

So, TLDR:
- the bug of displaying additional information on the console is fixed and released, so please update to 0.16.1
- the fact whether or not the exit code should be reverted back to `0` is unclear: I'll check one final time with Sebastián and will report back here.

Fabian665:
@svlandeg, I appreciate the response and the work you put into this repo in general. 

I guess the name of that PR is a bit misleading, and that is what led to the confusion. Anyway, I am glad to see things solved. As to the exit code itself, I am impartial on whether it should be 0 or 2. I went with 0 in this PR purely for backward compatibility, though I did not put much time or thought into this, and in hindsight, there is a case to be made in favor of the exit code 2.

Thanks again for your time and effort

svlandeg:
Ok, so just to wrap this up: I discussed with Sebastián (Tiangolo) internally, and we've decided to go with Click's behaviour, i.e. changing the exit code from 0 to 2 from Click 8.2.0 onwards.

As such, I think we can close this PR as the output issues have already been resolved. Thanks again everyone! 🙏

Issue #346, opened 2021-12-27
No autocompletion and syntax highlighting in callback for typer.Context but they work using click.Context


Minipada:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import click
import typer

def validate_name_typer(ctx: typer.Context, param: typer.CallbackParam, value: str) -> str:
    params = ctx.command.params # Autocompletion and syntax highlighting don't work
    return value

def validate_name_click(ctx: click.Context, param: typer.CallbackParam, value: str) -> str:
    params = ctx.command.params # Autocompletion and syntax highlighting work
    return value

def main(name: str= typer.Option(...,callback=validate_name_typer)):
    typer.echo(f"Hello {name}")


if __name__ == "__main__":
    typer.run(main)
```

### Description

On VSCode, I don't have autocompletion for the context in a callback. I am not sure where it comes from but this definitely seems a typer issue since it is working with Click.

![image](https://user-images.githubusercontent.com/1220912/147466352-6c2edc02-15c2-4e98-a908-ed3c8178890f.png)
![image](https://user-images.githubusercontent.com/1220912/147466413-c249884b-a82d-4504-affe-657eceddde7f.png)

I added both in the example.

### Operating System

Linux

### Operating System Details

```bash
$ cat /etc/os-release 
````
NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://manjaro.org/"
BUG_REPORT_URL="https://bugs.manjaro.org/"
LOGO=manjarolinux

### Typer Version

0.4.0

### Python Version

3.9.9

patricksurry:
having a similar weird issue where autocompletion is working fine in bash/osx, but in pwsh/win11 some of my autocomplete works, but others fail in the same script. 

[...]

PR #1257, opened 2025-07-20
 🐛 Fix `Context.command.params` type hinting for better IDE support

ashupal86:
# Fix IDE autocompletion for typer.Context

**Fixes:** #346

## What was broken

When using `typer.Context` in callbacks, IDEs couldn't autocomplete parameter attributes:

```python
def my_callback(ctx: typer.Context, param, value):
    params = ctx.command.params
    param.name  # ❌ No autocompletion
```

## What's fixed

Now IDEs provide full autocompletion:

```python
def my_callback(ctx: typer.Context, param, value):
    params = ctx.command.params
    param.name  # ✅ Full autocompletion!
```

## How it works

Added proper type hints to the `Context` class:

```python
# New in typer/models.py
class TypedCommand(click.Command):
    params: List[click.Parameter]

class Context(click.Context):
    command: TypedCommand
```

## Files changed

- `typer/models.py` - Added type hints
- `tests/test_ctx_params_typing.py` - Added tests

### DIFF
--- a/tests/test_ctx_params_typing.py
+++ b/tests/test_ctx_params_typing.py
@@ -2,50 +2,50 @@
 Test for improved type hints in Context.command.params
 
 This test verifies that the Context class now has proper type hints
-for the command.params attribute, allowing IDE autocompletion and 
+for the command.params attribute, allowing IDE autocompletion and
 type checking tools like mypy to work correctly.
 """
+
 from typing import List
 
 import click
-
 import typer
 from typer.models import Context
 
 
 def test_context_command_params_typing():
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Seems that this is not an issue in current version.

I checked in VSCode and it shows me all autocompletions:

----
image

----

image

svlandeg:
The tests introduced in this PR also succeed on `master`, so it's unclear what this PR would be fixing. I'll go ahead and close this PR, but we do appreciate the effort, @ashupal86!

PR #1262, opened 2025-07-28
Fix the output for no_args_is_help=True

mmakaay:
Since typer 0.16.0, running a CLI with the no_args_is_help feature enabled displays an error block. The patch from the PR suppresses the duplicate `Usage:` info and the `Error` output block, in case the processed exception is a NoArgsIsHelperror.

**Example with invalid output (based on the `no_args_is_help` example code)**

image

**Output with this patch applied:**

image

### DIFF
--- a/typer/core.py
+++ b/typer/core.py
@@ -20,6 +20,7 @@ from typing import (
 
 import click
 import click.core
+from click.exceptions import NoArgsIsHelpError
 import click.formatting
 import click.parser
 import click.shell_completion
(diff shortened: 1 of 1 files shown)


j-standfast:
+1

j-standfast:
+1 (as in, I'm seeing the same issue using Typer v0.16.0)

AleksBelytskyi:
+1 (the same issue here)

svlandeg:
Note that in the meantime, you can avoid this bug by restricting Click to < 8.2.

svlandeg:
@mmakaay: thanks again for your work on this! 

I'm closing this PR in favour of https://github.com/fastapi/typer/pull/1278 which has a unit test, some additional explanations and a different fix with a more narrow scope (in `rich_utils.py` instead of in `core.py`) which also keeps compatibility with older Click versions. 

I'd appreciate your review on this new PR if you have time! [edit: it's already merged in the meantime 🙈]

PR #1264, opened 2025-07-30
🐛 Fix evaluating stringified annotations in Python 3.9 and less (also `from __future__ import annotations`)

sisp:
I've fixed evaluating stringified annotations in Python 3.9 and less when using the `Annotated[<type>, typer.{Argument,Option}(...)]` syntax. When removing the `@needs_py310` decorator from the `test_annotated_argument_in_string_type_with_default` test, it fails with the following error:

```
――――――――――――― test_annotated_argument_in_string_type_with_default ――――――――――――――
[gw0] linux -- Python 3.9.20 .../typer/.venv/bin/python3

    def test_annotated_argument_in_string_type_with_default():
        app = typer.Typer()
    
        @app.command()
        def cmd(val: "Annotated[int, typer.Argument()]" = 0):
            print(f"hello {val}")
    
        result = runner.invoke(app)
        assert result.exit_code == 0, result.output
        assert "hello 0" in result.output
    
        result = runner.invoke(app, ["42"])
>       assert result.exit_code == 0, result.output
E       AssertionError: Usage: cmd [OPTIONS]
E         Try 'cmd --help' for help.
E         
E         Error: Got unexpected extra argument (42)
E         
E       assert 2 == 0
E        +  where 2 = <Result SystemExit(2)>.exit_code

tests/test_annotated.py:38: AssertionError
```

This problem typically when `from __future__ import annotations` is used with Python prior to 3.10.

This PR fixes the problem for Python versions prior to 3.10 by evaluating string annotations with `typing_extensions.get_annotations(...)` (added in v4.13.0) instead of `inspect.signature(..., eval_str=True)`, which is only available for Python 3.10+.


svlandeg:
Support for Python 3.8 has now also been dropped, so I think this PR will need to be thoroughly reviewed.

tiangolo:
As now Python 3.9 is no longer supported, we can close this one. Thanks for the effort here! :coffee:

PR #1267, opened 2025-08-02
Treat keyword only parameters as OptionInfo even without default

wouterzwerink:
Reviving #270 to make Typer treat keyword only parameters as options even when they dont have a default.

E.g. 
```python
@app.command()
def my_command(*, my_param):
     pass
```

Will now result in the following CLI with a flag instead of positional parameter:
`python my_module --my-param value`

Supports #269

### DIFF
--- a/tests/test_ambiguous_params.py
+++ b/tests/test_ambiguous_params.py
@@ -229,3 +229,21 @@ def test_forbid_default_and_default_factory_with_default_param(param, param_info
 )
 def test_error_rendering(error, message):
     assert str(error) == message
+
+
+def test_keyword_only_without_default_becomes_option():
+    app = typer.Typer()
+
+    @app.command()
+    def cmd(*, my_param):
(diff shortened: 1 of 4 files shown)


wouterzwerink:
I'm not sure where to add documentation for this, i.e. what the correct page would be. If anyone from the team could make a suggestion I'd be happy to document this.

YuriiMotov:
@wouterzwerink, thanks for your interest!

I think the drawback of such approach is that it might break the existing code bases.
It's common to declare all parameters as keyword-only in order to avoid having to place parameters with defaults after parameters without defaults. This feature might break such code..

```py
@app.command()
def hello(
    *,
    formal: bool = False,
    name: str,  #  <=  This will become option..
    color: str = "green",
):
    pass
```

https://github.com/fastapi/typer/issues/269#issuecomment-3302330010

Let's wait for Sebastian to approve the idea

PR #1278, opened 2025-08-18
🐛 Avoid printing additional output with `no_args_is_help=True` and Click 8.2

svlandeg:
Follow-up of #1262

The core issue: running a CLI with the `no_args_is_help` feature enabled displays an error block and a duplicated "Usage" line, whenever Rich ánd Click >= 8.2.0 is used (supported since Typer 0.16.0). The error does not occur without Rich, or with an older version of Click.

To replicate, run `python docs_src/commands/index/tutorial003.py` and inspect the console output.

There's two aspects to consider here:

### Bug fix 
Ensure this duplicate print & error message is avoided. Note that we can't import `NoArgsIsHelpError` from Click directly because that would break compatibility with Click 8.0 and 8.1, as this new error type was only introduced in 8.2.0. 

### Potentially breaking behaviour
Decide on the **exit code** of a function call without arguments but with `no_args_is_help=True`. Should it be 0 (as before) or 2 (as Click is doing since https://github.com/pallets/click/pull/1489)?
- If we want the exit code to be the same for any version of Typer/Click, we can test for its value explicitely in `test_tutorial003.test_no_arg`. We'll then need to add a custom workaround in Typer to change the behaviour compared to what Click does.
- If we're happy to stick with whatever Click does, we should see this as a breaking change introduced since Click 8.2. (see also [this discussion](https://github.com/pallets/click/issues/2909)). This is what the current state of the PR assumes.

### DIFF
--- a/tests/test_tutorial/test_commands/test_index/test_tutorial003.py
+++ b/tests/test_tutorial/test_commands/test_index/test_tutorial003.py
@@ -19,7 +19,7 @@ def test_no_arg():
 
 
 def test_no_additional_output():
-    """With Click 8.2, additional output was generated (cf. PR #1262)"""
+    """Ensure that no additional output was generated (cf. PR #1262)"""
     result = runner.invoke(app)
     assert result.output.count("Usage") == 1
     assert "Error" not in result.output


svlandeg:
A little bit more explanation as to why/how this happens:

At L211-220 we have this in `core.py`:
```
        except click.ClickException as e:
            if not standalone_mode:
                raise
            # Typer override
            if rich and rich_markup_mode is not None:
                rich_utils.rich_format_error(e)
            else:
                e.show()
            # Typer override end
            sys.exit(e.exit_code)
```
When Rich isn't used, the `else` structure goes to `e.show()` which shows the help message as before, after the changes from https://github.com/pallets/click/pull/1489, where `NoArgsIsHelpError.show()` is calling `echo(self.format_message(), ...)` and thus the output looks as before.

But when Rich is used, this gets bypassed and in `rich_format_error` we get an additional:
- `console.print(ctx.get_usage())` -> the duplicated Usage line
- `console.print(RICH_HELP.format(...))` -> the reference to `--help`
- `console.print(... self.format_message() ...)` -> the (empty) error message

mmakaay:
Thanks @svlandeg 
This definitely trumps my PR for fixing the issue,
and for completeness sake: works like a charm 😄👍

PR #1286, opened 2025-09-01
👷 Detect and label merge conflicts on PRs automatically

svlandeg:
- Label a PR that has merge conflicts with the new label "conflicts"
- Add a comment to the PR to alert the original poster & any subscribers to the thread
- Remove the label again when the conflict is resolved (but no further comment)


.github/workflows/conflict.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if PRs have merge conflicts
+        uses: eps1lon/actions-label-merge-conflict@v3
+        with:
+          dirtyLabel: "conflicts"
+          repoToken: "${{ secrets.GITHUB_TOKEN }}"
+          commentOnDirty: "This pull request has a merge conflict that needs to be resolved."
```
svlandeg, line 16: We can also remove this line, then no comment will be posted, only the label will be added.
I think posting a comment will be helpful though, both for the original poster as well as for maintainers who have worked on the PR and are subscribed to it.

PR #1290, opened 2025-09-04
🐛 Make sure `rich.markup` is imported when rendering help text

g-arjones:
rich.markup is used to escape strings when rendering help texts.

This fixes a regression introduced by https://github.com/fastapi/typer/pull/1128

### DIFF
--- a/tests/test_rich_utils.py
+++ b/tests/test_rich_utils.py
@@ -1,3 +1,6 @@
+import sys
+from typing import Annotated
+
 import typer
 import typer.completion
 from typer.testing import CliRunner
@@ -79,3 +82,21 @@ def test_rich_doesnt_print_None_default():
     assert "Hello Rick" in result.stdout
     assert "First: option_1_default" in result.stdout
     assert "Second: Morty" in result.stdout
+
+
+def test_rich_ness():
+    # Remove rich.markup if it was imported by other tests
+    if "rich" in sys.modules:
+        rich_module = sys.modules["rich"]
+        if hasattr(rich_module, "markup"):
+            delattr(rich_module, "markup")
+
+    app = typer.Typer(rich_markup_mode=None)
+
+    @app.command()
+    def main(bar: Annotated[str, typer.Argument(help="foobar")]):
+        pass
+
+    result = runner.invoke(app, ["--help"])
+    assert "Usage" in result.stdout
+    assert "bar" in result.stdout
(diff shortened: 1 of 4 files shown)


svlandeg (commented):
So just to clarify some edits, the original content of the PR imported `rich.markup` before using `escape`, because due to the lazy imports from #1128 this module wasn't available. 

I've rewritten this and moved the escaping functionality to `rich_utils`. My proposal is to never use `rich` directly in the core library, to isolate all this functionality within `rich_utils` and avoid any other future regressions.

There's a few more edits and maybe an additional test I'd like to add to this PR, so I'll put this in draft while I continue to work on it.

And thanks for the report & PR, @g-arjones ! 🙏

svlandeg:
@g-arjones: do you have a minimal example that crashes on `master`?

g-arjones:
@svlandeg Sure, here you go:

```python
from typing import Annotated

import typer


app = typer.Typer(rich_markup_mode=None)


@app.command()
def foo(bar: Annotated[str, typer.Argument(help='foobar')]) -> None:
    """foobar."""
    pass


app()
```

Then run with `--help`

svlandeg:
Thanks! This has been a bit tricky to get into the test suite, because other tests do import `rich.markup` at some point, which is why we didn't catch this regression earlier. I've now added a test that catches your originally reported regression on `master`, and that should go green on this PR.

I will add one more edit to this PR to fix a similar issue with accessing `rich.traceback` directly in `main.py`.

g-arjones:
Thanks for taking the time!

tests/test_rich_utils.py
```diff
@@ -79,3 +81,21 @@ def main(
...
     assert "First: option_1_default" in result.stdout
     assert "Second: Morty" in result.stdout
+
+
+def test_rich_markup_import_regression():
+    # Remove rich.markup if it was imported by other tests
+    if "rich" in sys.modules:
+        rich_module = sys.modules["rich"]
+        if hasattr(rich_module, "markup"):
+            delattr(rich_module, "markup")
```
svlandeg, line 91: A bit of voodoo to ensure this test fails when running on `master` as part of the larger test suite (because other tests end up importing `rich` etc)

tiangolo (approved):
Thanks for the report and fix! 🍰 

And thanks @svlandeg for the tweaks, confirmation, test... 🙌 

This will be available in Typer 0.17.4 released in the next few minutes. 🎉

PR #1297, opened 2025-09-08
✅ Use Ruff rules to ensure safe lazy-loading of `rich`

nathanjmcdougall:
To avoid regressions like #1294 occurring in the future, I propose a different way of lazy-loading rich using `importlib.util.find_spec`. I also propose using the `TID` Ruff rules to enforce the lazy-loading logic in the future.

This provides a framework to systematically apply other lazy-loading migrations safely.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -164,12 +164,14 @@ select = [
     "B",  # flake8-bugbear
     "C4",  # flake8-comprehensions
     "UP", # pyupgrade
+    "TID", # flake8-tidy-imports
 ]
 ignore = [
     "E501",  # line too long, handled by black
     "B008",  # do not perform function calls in argument defaults
     "C901",  # too complex
     "W191", # indentation contains tabs
+    "TID252", # Relative imports OK
 ]
 
 [tool.ruff.lint.per-file-ignores]
@@ -192,6 +194,8 @@ ignore = [
 "docs_src/prompt/tutorial003.py" = ["F841"]
 # Loop control variable `x` not used within loop body
 "docs_src/using_click/tutorial001.py" = ["B007"]
+# No need to avoid rich imports in docs
+"docs_src/*" = ["TID"]
 
 [tool.ruff.lint.isort]
 known-third-party = ["typer", "click"]
@@ -201,3 +205,9 @@ known-first-party = ["reigns", "towns", "lands", "items", "users"]
 [tool.ruff.lint.pyupgrade]
 # Preserve types, even if a file imports `from __future__ import annotations`.
 keep-runtime-typing = true
+
+[tool.ruff.lint.flake8-tidy-imports]
+banned-module-level-imports = ["typer.rich_utils"]
+
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+"rich".msg = "Use 'typer.rich_utils' instead of importing from 'rich' directly."
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,7 +177,7 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 # "__init__.py" = ["F401"]
 # rich_utils is allowed to use rich imports
-"typer\rich_utils.py" = ["TID251"]
+"typer/rich_utils.py" = ["TID251"]
 # This file is more readable without yield from
 "docs_src/progressbar/tutorial004.py" = ["UP028", "B007"]
 # Default mutable data structure
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -176,6 +176,8 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 # "__init__.py" = ["F401"]
+# rich_utils is allowed to use rich imports
+"typer\rich_utils.py" = ["TID251"]
 # This file is more readable without yield from
 "docs_src/progressbar/tutorial004.py" = ["UP028", "B007"]
 # Default mutable data structure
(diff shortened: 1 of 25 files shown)


svlandeg:
Thanks for working on this! I'm putting this in draft as long as the CI fails - feel free to mark as ready for review when the CI goed green!

nathanjmcdougall:
@svlandeg Alright, I think it's good now. I can't add labels so that check's failing, but everything else is green.

typer/main.py
```diff
@@ -90,7 +88,7 @@ def except_hook(
         if any(frame.filename.startswith(path) for path in internal_dir_names):
             if not exception_config.pretty_exceptions_short:
                 # Hide the line for internal libraries, Typer and Click
-                stack.append(
+                stack.append(  # pragma: no cover
```
svlandeg: Do you know why this statement has become necessary? It's worrying me somewhat, as it means that something must have changed in the testing functionality?

typer/cli.py
```diff
@@ -11,14 +11,7 @@
...
 
-try:
-    import rich
-
-    has_rich = True
-
-except ImportError:  # pragma: no cover
-    has_rich = False
-    rich = None  # type: ignore
+HAS_RICH = importlib.util.find_spec("rich") is not None
```
svlandeg: I appreciate you staying as close to the current code base as possible, and simply substituting this functionality whereever it appears. But while we're at it, should we clean up more? Have this `HAS_RICH` var only in `core` and import it from there in `cli` and `main`?

nathanjmcdougall: Yeah I was thinking that too - I'll go ahead and do that

typer/main.py
nathanjmcdougall: Agree, I think it's because I was careless in this file:

tests/assets/type_error_no_rich_short_disable.py

nathanjmcdougall: Yup, I can confirm that using the correct module to monkey patch in that file fixed the issue. I've remove the `# pragma: no cover` comment in https://github.com/fastapi/typer/pull/1297/commits/4c126f8a0f8892a0f0eafb6c7f7568b525e1ec66

pyproject.toml
```diff
@@ -201,3 +207,9 @@ known-first-party = ["reigns", "towns", "lands", "items", "users"]
 [tool.ruff.lint.pyupgrade]
 # Preserve types, even if a file imports `from __future__ import annotations`.
 keep-runtime-typing = true
+
+[tool.ruff.lint.flake8-tidy-imports]
+banned-module-level-imports = ["typer.rich_utils"]
```
svlandeg, line 213: This might just be me being confused, but shouldn't this be 
```suggestion
banned-module-level-imports = ["rich"]
```
?

nathanjmcdougall: This config is specifically to force imports of `typer.rich_utils` to stay within functions, i.e. they shouldn't be done at the top of the module at the module level (hence "`banned-module-level-imports`"). Otherwise it's possible to bypass the lazy loading.

This is the Ruff rule that's using this config with some further information:
https://docs.astral.sh/ruff/rules/banned-module-level-imports/

The `banned-api` config below is the one to prevent directly importing `rich` anywhere (but we have a per-file exception for `rich_utils`).

I will add a comment to clarify.

svlandeg: Ah, thanks for clarifying. I didn't quite get that L215-L216 both define the blocked API ànd define the error msg - I thought it did only the latter. Clear, tnx!

svlandeg (approved):
Great, so a recap.

This PR does a number of related things:
1. Avoid importing `rich_utils` at the top of any module, ensuring its lazy loading from #1128 (`flake8-tidy-imports`)
2. Avoid importing `rich` anywhere in the codebase, except in `rich_utils.py` and docs files (`flake8-tidy-imports.banned-api`)
3. Refactor how to define and store `HAS_RICH`: using `importlib.util.find_spec("rich")` instead of a `try`-`catch` import in `core.py`
4. Fetching the `HAS_RICH` var from `core.py` in both `main.py` and `cli.py` to avoid code duplication and have easier maintenance in the future
5. Refactor the tests to use `HAS_RICH` and `pytest.MonkeyPatch` instead of setting the `typer.core.rich` variable directly.

To me, 1 & 2 are the crucial points of this PR: they ensure we're not importing `rich` or `rich_utils` by accident. This would have caught #1290 as well as a similar issue we encountered in #828: it will be nice to set things up so that we can't hit ourselves with this again 🥲 

3 follows out of this because you can't import `rich` anymore, and in fact having a `HAS_RICH` variable feels like a nicer solution as well, as compared to our old solution of `rich` either being an imported library, or `None`.

4 was added to this PR by my suggestion as I think it's more clean to define this functionality only once and then import.

5, the refactoring of the tests, needs to happen anyway when switching from the `rich` variable check for `None` to the boolean `HAS_RICH`. Using `MonkeyPatch` to do this feels more elegant to me than what we had before.

**So, all in all, this looks (really) good to me.** Thanks again @nathanjmcdougall for all your work on this!

I'm passing this on internally to Tiangolo's reviewing queue, he will update here once he's had a chance to review 🙏

nathanjmcdougall:
After this PR is done, I plan do the same but with `shellingham`.

tiangolo:
Awesome, looks great! Thanks for all the work on this @nathanjmcdougall! :raised_hands: 

And thanks @svlandeg for the thorough explanation of everything and the reasons for it all. :rocket:

PR #1303, opened 2025-09-15
✨ Allow setting an environment variable to `None` in `CliRunner.invoke`

arjenzorgdoc:
The type of the Click `env` parameter is `Mapping[str, str | None] | None`. This can be used to "delete" an environment variable.


tiangolo:
Great, thanks! :rocket:

PR #1304, opened 2025-09-15
🐛 Ensure that the default `rich_markup_mode` is interpreted correctly

svlandeg:
By using `Default(DEFAULT_MARKUP_MODE)`, `rich_markup_mode` would be a `typer.models.DefaultPlaceholder` object and comparing it to other literal strings like `MARKUP_MODE_RICH` would fail. This resulted in flaky behaviour such as reported in https://github.com/fastapi/typer/discussions/976, where Rich formatting worked fine when specifying it correctly, but the default value (with Rich installed) didn't result in the same behaviour.

This PR fixes that by simply not using `Default` for this value - there's no real reason to use it anyway, as it's unimportant to the application whether it's a default or a user-set value.

### DIFF
--- a/tests/test_rich_markup_mode.py
+++ b/tests/test_rich_markup_mode.py
@@ -3,6 +3,7 @@ from typing import List
 import pytest
 import typer
 import typer.completion
+from typer.rich_utils import MARKUP_MODE_RICH
 from typer.testing import CliRunner
 
 runner = CliRunner()
@@ -284,3 +285,10 @@ def test_markup_mode_bullets_double_newline(mode: str, lines: List[str]):
     arg_start = [i for i, row in enumerate(result_lines) if "Arguments" in row][0]
     assert help_start != -1
     assert result_lines[help_start:arg_start] == lines
+
+
+def test_markup_mode_default():
+    app = typer.Typer()
+
+    assert app.rich_markup_mode == MARKUP_MODE_RICH
+
--- a/typer/main.py
+++ b/typer/main.py
@@ -138,7 +138,7 @@ class Typer:
         deprecated: bool = Default(False),
         add_completion: bool = True,
         # Rich settings
-        rich_markup_mode: MarkupMode = Default(DEFAULT_MARKUP_MODE),
+        rich_markup_mode: MarkupMode = DEFAULT_MARKUP_MODE,
         rich_help_panel: Union[str, None] = Default(None),
         pretty_exceptions_enable: bool = True,
         pretty_exceptions_show_locals: bool = True,


svlandeg:
The failing coverage test reveals that there are code paths in `rich_utils.py` that are redundant. They were only taken before by accident, because the comparisons weren't done properly with the `Default()` object.

In fact, the formatting methods in `rich_utils` shouldn't be called when `rich_markup_mode` is `None` (calling them would be problematic as `rich` would be imported!). To clarify this in code and to avoid further issues in the future, I've introduced `MarkupModeStrict` and removed the code paths that are redundant.

typer/rich_utils.py
```diff
@@ -104,6 +104,7 @@
 _RICH_HELP_PANEL_NAME = "rich_help_panel"
 
 MarkupMode = Literal["markdown", "rich", None]
```
YuriiMotov: Seems that we don't need this type (it's not used)

svlandeg: It is being used outside of `rich_utils`: in `core` and `main` we use `MarkupMode` which can be `None`.

YuriiMotov: We have one more `MarkupMode` defined in `core` and it's that type that is used in `core` and `main`.
`MarkupMode` defined in `rich_utils` is not used

https://github.com/fastapi/typer/blob/0bc7cced82ff95681ecc095d0ef461840a24862c/typer/core.py#L31

svlandeg: Oh! Ofcourse, because we shouldn't import anything from `rich_utils` until we're sure we're using Rich. Thanks. I'll go ahead and delete the duplicate defined `MarkupMode` and keep only the strict one in `rich_utils`.

svlandeg:
Feels to me like the CI is failing due to an unrelated issue - I'll investigate.

[UPDATE]: fixed with https://github.com/fastapi/typer/pull/1333

tests/test_rich_markup_mode.py
```diff
@@ -284,3 +284,9 @@ def main(arg: str):
     arg_start = [i for i, row in enumerate(result_lines) if "Arguments" in row][0]
     assert help_start != -1
     assert result_lines[help_start:arg_start] == lines
+
+
+def test_markup_mode_default():
+    # We're assuming the test suite is run with rich installed
+    app = typer.Typer()
+    assert app.rich_markup_mode == "rich"
```
svlandeg, line 292: on `master`, this fails with 
> <typer.models.DefaultPlaceholder object at 0x00000227BB50B130> != rich

PR #1306, opened 2025-09-15
➖ Only install `typing_extensions` for python '<3.10'

waketzheng:
# Description
1. Only install `typing_extensions` for python '<3.10'
2. Bump up version of ruff/mypy

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -5,8 +5,8 @@ pytest-cov >=2.10.0,<8.0.0
 coverage[toml] >=6.2,<8.0
 pytest-xdist >=1.32.0,<4.0.0
 pytest-sugar >=0.9.4,<1.2.0
-mypy ==1.4.1
-ruff ==0.12.12
+mypy ==1.18.1
+ruff ==0.13.0
 # Needed explicitly by typer-slim
 rich >=10.11.0
 shellingham >=1.3.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ classifiers = [
 ]
 dependencies = [
     "click >= 8.0.0",
-    "typing-extensions >= 3.7.4.3",
+    "typing-extensions >= 3.7.4.3; python_version < '3.10'",
 ]
 readme = "README.md"
 [project.urls]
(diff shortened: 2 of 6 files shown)


requirements-tests.txt
```diff
@@ -5,8 +5,8 @@ pytest-cov >=2.10.0,<8.0.0
 coverage[toml] >=6.2,<8.0
 pytest-xdist >=1.32.0,<4.0.0
 pytest-sugar >=0.9.4,<1.2.0
-mypy ==1.4.1
-ruff ==0.12.12
+mypy ==1.18.1
+ruff ==0.13.0
```
svlandeg: These are unrelated edits, let's remove them for simplicity (bots are pushing updates of these dependencies in separate PRs)
```suggestion
mypy ==1.4.1
ruff ==0.12.12
```

svlandeg:
Hi @waketzheng, I see you're running into the same typing error we've encountered before: https://github.com/fastapi/typer/pull/1048#issuecomment-2562379115

Basically this happens when reimporting `Literal`. For now, it's find to leave that bit of duplicated code as-is.

waketzheng:
> many of our tutorial examples use `from typing_extensions import Annotated`

How about use `from typer._typing import Annotated` instead?

svlandeg (approved):
This PR looks good to me - it's a good idea not to install `typing_extensions` when we don't need it for Python 3.9 and above.

> > What is a bit unfortunate is that many of our tutorial examples use from typing_extensions import Annotated. I worry that users won't be able to execute this code directly if they haven't got typing_extensions installed in their normal typer environment.
> 
> How about use `from typer._typing import Annotated` instead?

As I see it, our options are:
- keep the examples as they are, and update them to `from typing import Annotated` when we drop Python 3.8
- update the examples to `from typer._typing import Annotated` as you suggest but this feels like adding unnecessary complexity to the tutorial examples (exposing what should be an implementation detail)
- we can have explicit Python 3.9+ examples for all tutorial examples, that import from `typing` instead. While I wouldn't mind working on this, it feels like a lot of wasted effort, especially if we plan on dropping Python 3.8 soon.

I'll leave this decision with Tiangolo 🙏

PR #1333, opened 2025-09-19
👽️ Ensure compatibility with Click 8.3.0 by restoring the original `value_is_missing` function

svlandeg:
Yesterday's release of [Click 8.3.0](https://github.com/pallets/click/releases/tag/8.3.0) has [introduced](https://github.com/pallets/click/pull/3030) a new internal variable `UNSET` to differentiate between (truly) unset values and user-provided `None` values. As a consequence, the method `Parameter.value_is_missing` in `click.core` now checks for this `UNSET` value instead of for `None`. 

This results in bugs when using the current Typer with Click 8.3.0+, because a required, unset parameter would be `None` in Typer (and not the new `UNSET` value) and would eventually get the default value instead of throwing an exception.

For instance:
```
@app.command()
def main(name: str):
    print(f"Hello {name}")
```
would result in a print `"Hello None"` when calling it without any arguments.

## Potential solutions

1. A quick fix would be to [pin](https://github.com/fastapi/typer/pull/1336) Typer to <8.3.0, but obviously that would only be temporary.
2. This PR overwrites Click's `value_is_missing` function to revert it back to its functionality from before 8.3.0. This at least makes the test go green, and allows users to install Click 8.3.0+
3. Rewrite Typer's core more fundamentally to exploit the new options of the `UNSET` value. This should also allow us to make different choices with respect to optional flags & flag values, cf https://github.com/fastapi/typer/pull/987. Maybe with the new functionality in Click, we can restore some of that again. But it would require careful refactoring if we also want to continue support Click 8.2 and lower.

## Proposal

At this point in time (right before the weekend 🙃) I suggest to merge this PR (option 2), close #1336 (option 1) and continue working on option 3 in the next few weeks.

### DIFF
--- a/typer/core.py
+++ b/typer/core.py
@@ -609,7 +609,7 @@ def _value_is_missing(param: click.Parameter, value: Any) -> bool:
     #     return True
 
     if (param.nargs != 1 or param.multiple) and value == ():
-        return True    # pragma: no cover
+        return True  # pragma: no cover
 
     return False
 
--- a/typer/core.py
+++ b/typer/core.py
@@ -609,7 +609,7 @@ def _value_is_missing(param: click.Parameter, value: Any) -> bool:
     #     return True
 
     if (param.nargs != 1 or param.multiple) and value == ():
-        return True
+        return True    # pragma: no cover
 
     return False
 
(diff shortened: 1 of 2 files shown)


svlandeg:
This must be related to https://github.com/pallets/click/pull/3030 and using the "UNSET" value...

tiangolo (approved):
Thank you for the quick workaround and all the investigation and explanation, great as always! 🙌 

I'll release this now in Typer 0.18.0 🚀

Issue #1159, opened 2025-02-20
[BUG] help line length miscalculated when using stylized text


zmeir:
### Discussed in https://github.com/fastapi/typer/discussions/1101

<div type='discussions-op-text'>

<sup>Originally posted by **zmeir** December 23, 2024</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorials in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Annotated

import typer


app = typer.Typer()


@app.command()
def example(
    a: Annotated[str, typer.Option(help="This is A")],
    b: Annotated[str, typer.Option(help=f"This is {typer.style('B', underline=True)}")],
):
    pass


if __name__ == "__main__":
    app()
```

### Description

When using `typer.style` in the help text of a command option, the rightmost `|` in the "frame" of the Options output is misaligned. This seems to be due to a miscalculation of the help line length, caused by the invisible style characters.

Using the example script - run `python example.py --help` and you should get this output:
![image](https://github.com/user-attachments/assets/ed8dd11b-f32f-45da-8e7a-6542535b0a97)
(highlighted the misaligned part)

### Operating System

macOS

### Operating System Details

macOS version: 13.7.1 (22H221)

### Typer Version

0.15.1

### Python Version

3.10.13

### Additional Context

_No response_</div>

mahimairaja:
HI @svlandeg are you actively working on this issue?

svlandeg:
Not currently, but it's on my TODO list. If you want to work on it, please let me know, a PR is always appreciated! 🙏

mahimairaja:
Thanks @svlandeg, I would like to work on this issue. Can you please assign me

svlandeg:
We don't have a formal system of assigning tickets, other than me using assigned PRs & issues to keep track of my personal TODO list. Feel free to work on it and link the PR here when you're done - others will see from this discussion that you're working on it.

mahimairaja:
Sure Thanks! I am starting to work on it

mahimairaja:
Hi @svlandeg 

[...]

PR #1356, opened 2025-09-25
Fix `--help` text alignment when using typer.style() in option descriptions

mahimairaja:
# Problem

When using typer.style() in help text for CLI options, the help table columns become misaligned. This happens because typer.style() returns a string with ANSI escape codes, and Rich's width calculation includes these control codes, causing incorrect column sizing.

Relevant issue - #1159 

## Root Cause:
The issue occurs in _make_rich_text() function in rich_utils.py. When help text contains ANSI escape codes from typer.style(), the function was creating a plain Text object that preserved the escape codes as literal characters. Rich's width measurement then counted these control codes, causing misalignment.

## Sample Code to reproduce:
```python
from typing import Annotated

import typer

app = typer.Typer()


@app.command()
def example(
    a: Annotated[str, typer.Option(help="This is A")],
    b: Annotated[str, typer.Option(help=f"This is {typer.style('B', underline=True)}")],
):
    pass


if __name__ == "__main__":
    app()

```

## Before fix:
```
❯ python main.py --help
                                                                                                               
 Usage: main.py [OPTIONS]                                                                                      
                                                                                                               
╭─ Options ───────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  --a                         TEXT  This is A [required]                                                   │
│ *  --b                         TEXT  This is B [required]                                             │
│    --install-completion              Install completion for the current shell.                              │
│    --show-completion                 Show completion for the current shell, to copy it or customize the     │
│                                      installation.                                                          │
│    --help                            Show this message and exit.                                            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

## After fix:
```
❯ python main.py --help
                                                                                                               
 Usage: main.py [OPTIONS]                                                                                      
                                                                                                               
╭─ Options ───────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  --a                         TEXT  This is A [required]                                                   │
│ *  --b                         TEXT  This is B [required]                                                   │
│    --install-completion              Install completion for the current shell.                              │
│    --show-completion                 Show completion for the current shell, to copy it or customize the     │
│                                      installation.                                                          │
│    --help                            Show this message and exit.                                            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

## Testing
- [x] Verified with the provided example showing proper alignment
- [x] Tested with various typer.style() combinations
- [x] Confirmed no regression in existing functionality

### DIFF
--- a/typer/rich_utils.py
+++ b/typer/rich_utils.py
@@ -167,7 +167,11 @@ def _make_rich_text(
     if markup_mode == MARKUP_MODE_RICH:
         return highlighter(Text.from_markup(text, style=style))
     else:
-        return highlighter(Text(text, style=style))
+        ANSI_ESCAPE_SEQUENCE = "\x1b["
+        if ANSI_ESCAPE_SEQUENCE in text:
+            return highlighter(Text.from_ansi(text, style=style))
+        else:
+            return highlighter(Text(text, style=style))
 
 
 @group()
--- a/tests/test_rich_utils.py
+++ b/tests/test_rich_utils.py
@@ -99,3 +99,46 @@ def test_rich_markup_import_regression():
     result = runner.invoke(app, ["--help"])
     assert "Usage" in result.stdout
     assert "BAR" in result.stdout
+
+
+def test_make_rich_text_with_ansi_escape_sequences():
+    from typer.rich_utils import Text, _make_rich_text
+
+    ansi_text = "This is \x1b[4munderlined\x1b[0m text"
+    result = _make_rich_text(text=ansi_text, markup_mode=None)
+
+    assert isinstance(result, Text)
+    assert "\x1b[" not in result.plain
+    assert "underlined" in result.plain
+
+    mixed_text = "Start \x1b[31mred\x1b[0m middle \x1b[32mgreen\x1b[0m end"
+    result = _make_rich_text(text=mixed_text, markup_mode=None)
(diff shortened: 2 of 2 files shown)


mahimairaja:
Hi @svlandeg 

Here is the screenshot before fix:
Screenshot 2025-10-04 at 5 54 20 AM

after fix:
Screenshot 2025-10-04 at 5 53 06 AM

Thanks for the guidance, 

1. I have updated the ansi prefix as similar to click 
2. Yes, @svlandeg let's have a if condition, if we are passing all plain text along with ansi text, it would increase time complexity.
3. Added the test case for `|` based on the comparing all three lines as you mentioned.


Can you please review the changes.

typer/rich_utils.py
```diff
@@ -167,7 +172,11 @@ def _make_rich_text(
     if markup_mode == MARKUP_MODE_RICH:
         return highlighter(Text.from_markup(text, style=style))
     else:
-        return highlighter(Text(text, style=style))
+        # if _ANSI_ESCAPE_SEQUENCE in text:
```
svlandeg:
```suggestion
```

mahimairaja:
@svlandeg

mahimairaja:
Thank you @svlandeg!

PR #1365, opened 2025-10-03
📝 Remove unnecessary quotes in `printing.md`

svlandeg:
On currenty `master`, this now prints as:

image

So better to remove the quotes and ensure the bold actually works.


svlandeg (commented):
Ok, so it's weird, because markdown should support something like `"**f**ore**g**round"` and in fact it does when I render locally. But when our docs get built, this isn't rendered properly. When I remove the quotes, it also doesn't work.

So for now I'm suggesting to use `<strong>` tags instead.

PR #1405, opened 2025-11-12
✨ Make `epilog` formatting consistent with other parts of the help string

svlandeg:
As discussed in https://github.com/fastapi/typer/discussions/1400, the formatting of the `epilog` section of the help text is currently not consistent with the other parts of the help text.

This PR introduces a new private function `_fix_linebreaks` that has the same functionality as before as part of `_get_help_text`. It deals with linebreaks depending on the given markup mode. This function is then also used when printing the epilog, ensuring consistency. Only the formatting of the epilog is changed, nothing else.

On `master`, the new unit test fails both for "markdown" and "rich", as it joins together the first few lines into one big `"Just wrapping up: This is the first conclusion     Here is conclusion two"` line.


travishathaway (approved):
I had a chance to read the original discussion and think this is a perfectly fine solution to this problem.

EnigmaticCypher:
This looks like an elegant solution to the problem in my view, and also brings over some behaviour from the help description processing (mainly the usage of `\b` to indicate that single lines shouldn't be stripped). While the `\b` thing isn't related to my discussion, it is something I use in a typer CLI tool I maintain at work, so am happy to see that included as well.

Great PR 🚀 

I'll rebase my internal fork of typer and include these fixes until @tiangolo merges this PR. The last PR for an issue I was running into took about 1.5 years to get merged into typer master, so I'm hoping this one doesn't take quite that long 😅

svlandeg (commented):
The code could further be simplified after https://github.com/fastapi/typer/pull/1304 has been merged. This is again good to review.

medecau:
@tiangolo can we get this pull-request merged?

🙏🏼 thank you!

EnigmaticCypher:
@tiangolo I'm aware you're very busy, but this is a really minor fix that will mean I can stop maintaining a separate fork of typer that includes this fix. If you could please merge this I would be most grateful 😄

tiangolo (approved):
Thank you @svlandeg for the fix (and the ping 😅  )!

And thanks everyone for the discussion and confirmation this works.

This will be available in Typer 0.27.1 in the next few hours. 🎉

PR #1422, opened 2025-11-23
✨ Enhance command alias handling and display

### LINKED ISSUE #1242: Add support for command and subcommand aliases
## Feature Request: Add support for command and subcommand aliases

As discussed in https://github.com/fastapi/typer/issues/132#issuecomment-2417492805, there's a need for official alias support in Typer.

### Examples needed:
- Command aliases: `list` → `ls`, `remove` → `rm`
- Subcommand aliases: `versions` → `ver` → `v`, `documents` → `docs`

### Current workaround:
```python
@app.command("list")
@app.command("ls", hidden=True)  # Hidden alias
```

### Requested feature:
Official alias support that doesn't clutter help output and provides a clean API for creating command shortcuts.

This would bring Typer in line with other popular CLI tools that support aliases natively.

pipinstalled:
- Added support for positional and additional aliases in the command decorator.
- Updated `format_commands` to display command names along with their visible aliases.
- Improved command listing to maintain original order and avoid duplicates.
- Enhanced command info structure to include aliases and hidden aliases.

This improves the usability and clarity of command help output.


svlandeg (commented):
@pipinstalled: you can go ahead and set add ` # pragma: no cover` to the `pass` statements in the tests, otherwise the coverage check will keep failing.

pipinstalled:
> @pipinstalled: you can go ahead and set add ` # pragma: no cover` to the `pass` statements in the tests, otherwise the coverage check will keep failing.

@svlandeg  I added # pragma but still coverage failed

svlandeg:
> > @pipinstalled: you can go ahead and set add ` # pragma: no cover` to the `pass` statements in the tests, otherwise the coverage check will keep failing.
> 
> @svlandeg I added # pragma but still coverage failed

You'll have to look into the specifics of the coverage report. What this typically means is that there's a code path that is untested, so you'd have to implement more tests to cover it.

pipinstalled:
@svlandeg  Thanks for your help, fixed with adding new tests.

pipinstalled:
@svlandeg  Have you any idea when this PR will merge? :D

pipinstalled:
@tiangolo  Do you have any idea when this PR will merge? :)

basnijholt:
This is great! Right now I am hacking around it (https://github.com/basnijholt/compose-farm/pull/148).

basnijholt (approved):

onyx-and-iris:
This will be great, I went the route of creating a custom alias class but I recently realised it broke shell completion :/ Thanks for all the efforts you've gone to with this PR.

tests/test_tutorial/test_commands/test_name/test_tutorial002.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def test_help():
+    result = runner.invoke(app, ["--help"])
+    assert result.exit_code == 0
+    assert "Commands" in result.output
+    assert "list" in result.output or "ls" in result.output
+    assert (
+        "remove" in result.output or "rm" in result.output or "delete" in result.output
+    )
```
svlandeg, line 17: These should all be `and` instead of `or`, right?

svlandeg (commented):
Thanks for the PR, @pipinstalled!

I can see how it can be useful to define an alias for a given command. 

In this specific PR, there are two ways implemented to accomplish that: either by positional arguments, or using an `aliases=` parameter. I'm not usually keen on implementing different ways of accomplishing the same thing: it creates confusion for users and unnecessary code maintenance.

Further, there is already a way (on `master`) to register different names for the same command:
```
@app.command("list")
@app.command("ls")
@app.command("listthem", hidden=True)
def list_items():
    print("Listing items")
```

I know that that would show up slightly differently in the `help`, they would show up as separate commands. At the same time I do also worry that this PR adds a new `format_commands` function and edits the Rich formatting in `_print_commands_panel` - it feels like these are potential cases for regressions or future issues.

TLDR: I'm not convinced that this a feature we'll want to support, especially considering the amount of changes needed in `core` etc. I'll run this by Tiangolo though, to see what he thinks.

pipinstalled:
@svlandeg  Thanks for the review!

That makes sense. I didn’t mean to introduce two different ways of doing the same thing — that’s definitely not ideal long term.

My main goal was to allow defining aliases while keeping them grouped as a single logical command in the help output. The existing decorator approach works, but it shows them as separate commands, which can get a bit noisy in larger CLIs.

If having both positional aliases and `aliases=` feels redundant, I’m totally fine simplifying this and keeping just one approach.

About the changes in `format_commands` and `_print_commands_panel`: fair point. I can take another look and see if there’s a way to reduce the surface of those changes.

Happy to adjust this based on your and @tiangolo’s thoughts.

onyx-and-iris:
I definitely agree on just one way to define aliases, my vote would go towards the _aliases=_ keyword.

I would also like to see an improvement on the default way of defining them. Defining multiple decorators and then hiding them from the help output seems less than ideal.

svlandeg:
> If having both positional aliases and `aliases=` feels redundant, I’m totally fine simplifying this and keeping just one approach.
> 
> About the changes in `format_commands` and `_print_commands_panel`: fair point. I can take another look and see if there’s a way to reduce the surface of those changes.

I think doing both of these would help increase this PR's chances of getting reviewed by Tiangolo & (hopefully) merged. I agree with @onyx-and-iris, to me the `aliases` keyword also seems like the nicest approach.

PR #1473, opened 2026-01-02
docs: fix CliRunner.invoke() docstring to reference 'app' parameter

Mohamed-Elwasila:
## What I Fixed
The `CliRunner.invoke()` method had outdated documentation that referenced a `cli` parameter, but the actual parameter is called `app`.

## Why This Matters
When developers use `CliRunner` for testing, they see documentation mentioning a parameter that doesn't exist. This happens because Typer's `CliRunner` inherits the docstring from Click's version, which uses different parameter names.

Now the docs correctly show that you pass in a `Typer` application object via the `app` parameter.

## Changes Made
- Changed `:param cli:` to `:param app:` in the docstring
- Added a note explaining this is a Typer-specific wrapper around Click's CliRunner

### DIFF
--- a/typer/testing.py
+++ b/typer/testing.py
@@ -18,6 +18,28 @@ class CliRunner(ClickCliRunner):
         color: bool = False,
         **extra: Any,
     ) -> Result:
+        """Invokes a Typer application in an isolated environment.
+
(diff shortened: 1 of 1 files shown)


phalberg:
I guess this one is now relevant to #1862.
Since it seems like a minor docs improvement.

PR #1476, opened 2026-01-06
Add module docstring to colors module

Mohamed-Elwasila:
## Summary
Adds a module-level docstring to `typer/colors.py` to document its purpose and usage.

## Changes
- Added module docstring explaining color constants for terminal styling
- Included usage examples with `typer.style()` and `typer.secho()`

The colors module previously only had a comment. This adds proper documentation that will show up in IDE tooltips and generated docs.


derOtto:
Hi @Mohamed-Elwasila,

thanks a lot for your contribution!

I'm not sure if this change really helps users when working with Typer.  
There are already many different ways to print text with colors, and I was also confused myself about which colors I can use in which situation (using `rich.print`, `typer.echo`, `typer.secho`, `typer.style`).

Because of that, would it maybe be better to improve the documentation here: https://typer.tiangolo.com/tutorial/printing/  
For example, by explaining more clearly **how** and **which** colors can be used for each of the different printing options?

I found these links for the two approaches:

- For Rich: https://rich.readthedocs.io/en/stable/appendix/colors.html  
- For `typer.echo`, `typer.secho`, `typer.style`: https://click.palletsprojects.com/en/stable/api/#click.style

tiangolo:
This is currently not necessary, thanks! :coffee:

PR #1487, opened 2026-01-10
Handle Abort/Exit during shell completion

### LINKED ISSUE #286: [BUG] Abort and Exit break autocompletion
I believe these two types of exceptions should be handled internally to just abort the autocompletion (equivalent to returning an empty list), instead of leaking and flooding the terminal with a traceback.

I'm willing to try and fix this if maintainers agree :)

mpouzoup:
Fixes #286

When running shell completion, autocompletion callbacks may raise click.exceptions.Abort or
click.exceptions.Exit. This should not print a traceback; completion should abort silently and
return no completions instead.


svlandeg (commented):
Thanks for the PR! Can you look into the coverage tests that are failing?

tests/test_shell_completion_abort_exit.py
```diff
@@ -0,0 +1,65 @@
...
+        instruction=instruction,
+    )
+
+    out = capsys.readouterr()
+    assert code == 0
+    assert out.out == ""
+    assert out.err == ""
+
+
+def test_fake_completion_default_methods_are_covered():
```
svlandeg, line 62: This is just to avoid coverage issues on this particular file, right? As it's a test file, it's cleaner to just add `# pragma: no cover` to those respective lines, no need to have a test testing the test 😉

svlandeg (commented):
Makes sense to me!

PR #1498, opened 2026-01-23
✨ Suppress modules in Rich tracebacks with `pretty_exceptions_suppress`

harkabeeparolus:
I believe we should provide access to the fantastic [Rich traceback suppress parameter](https://rich.readthedocs.io/en/stable/traceback.html#suppressing-frames). This would allow developers to suppress uninteresting stack frames in tracebacks from other Python frameworks or SDKs. 😊

Why would developers want this? Well, here's a screenshot showing the difference. **With** traceback suppression (from this PR) on the left, and **without** on the right. Notice that on the right, you don't even see your own code unless you scroll up. 😳

traceback_suppress_gitlab

And if you enable `Typer(pretty_exceptions_show_locals=True)` for easier development, it becomes even worse! 😨

Normally, with Rich, you can easily exclude framework code from tracebacks. But when Typer is also involved this stops working, because Typer installs its own `sys.excepthook`, and therefore this Rich feature is no longer accessible to developers. 😟 Since it is blocked, I believe we should expose the Rich feature for suppressing frames through the Typer API instead.

For the example in the screenshot, I used the `python-gitlab` API client, which gives a verbose traceback with many frames on a simple `GitlabAuthenticationError`:

```python
#! /usr/bin/env -S uv run --script

# /// script
# dependencies = [ "python-gitlab", "typer", "rich" ]
# ///

import gitlab
import rich.traceback
import typer

# This code does nothing; Typer will override it
rich.traceback.install(suppress=[gitlab])

def main():
    gl = gitlab.Gitlab()
    print(gl.pagesdomains.list())

typer.run(main)
```

But with this pull request, you can choose to suppress modules directly with `Typer()`:

```python
#! /usr/bin/env -S uv run --script

# /// script
# dependencies = [ "python-gitlab", "typer" ]
# ///

import gitlab
import typer

app = typer.Typer(pretty_exceptions_suppress=[gitlab])

@app.command()
def main():
    gl = gitlab.Gitlab()
    print(gl.pagesdomains.list())

app()
```


svlandeg:
I'll put this in draft as long as the tests are failing 🙏

harkabeeparolus:
Whoops, I still need to rename the tutorial file to "_py310". Fixing.

svlandeg:
Thanks @harkabeeparolus! I'll try to review this week or the next 🙏

docs/tutorial/exceptions.md
```diff
@@ -192,6 +192,27 @@ $ python main.py
 
 </div>
 
+## Disable Tracebacks From Certain Modules
+
+If you are developing with Python frameworks other than **Typer** and **Click**,
```
svlandeg: As we'll be vendoring Click in the near future, I suggest to not include the reference here.
```suggestion
If you are developing with other Python frameworks besides **Typer**,
```

docs_src/exceptions/tutorial005_py310.py
```diff
@@ -0,0 +1,18 @@
+import gitlab
```
svlandeg: I see that you're mocking the installation of `gitlab` for this to run in the test suite.

What I dislike about using `gitlab` as an example here, is that users reading the tutorial would have to install the lib just to replicate this one example.

Is there a different example, with already installed libraries, you can think of to showcase this functionality?

docs/tutorial/exceptions.md
```diff
@@ -192,6 +192,27 @@ $ python main.py
 
 </div>
 
+## Disable Tracebacks From Certain Modules
```
svlandeg, line 195: Similar to the other sections in this part of the tutorial, this section should show the output before and after setting this new var.

docs_src/exceptions/tutorial005_py310.py
```diff
@@ -0,0 +1,18 @@
...
+app = typer.Typer(pretty_exceptions_suppress=[gitlab])
+
+
+@app.command()
+def main():
+    gitlab_client = gitlab.Gitlab()
+
+    # This will raise an exception if not authenticated:
+    #     GitlabAuthenticationError: 401: 401 Unauthorized
+    # But the traceback will not show any lines from the gitlab module!
```
svlandeg: In general we tend to avoid comments in the source files, as they are displayed verbatim in the tutorial, which should have all the relevant information.

```suggestion
```

tests/test_tutorial/test_exceptions/test_tutorial005.py
```diff
@@ -0,0 +1,32 @@
...
+
+from typer.testing import CliRunner
+
+# Mock the gitlab module before importing the tutorial
+_gitlab = ModuleType("gitlab")
+_gitlab.Gitlab = MagicMock  # type: ignore[attr-defined]
+_gitlab.__file__ = "gitlab/__init__.py"  # type: ignore[attr-defined]
+sys.modules.setdefault("gitlab", _gitlab)
+
+from docs_src.exceptions import tutorial005_py310 as mod  # noqa: E402
```
svlandeg: There's no need for `noqa: E402` here, just move the import to the top of the file.

tests/test_tutorial/test_exceptions/test_tutorial005.py
```diff
@@ -0,0 +1,32 @@
...
+
+runner = CliRunner()
+
+# There's no way to test a third-party package from PyPI. Also, the actual
+# feature we use is part of Rich. We just pass the flag along. So here we test
+# that the tutorial code runs without errors.
+
+# Perhaps we could find some standard library module that throws a long
+# traceback and test that instead, but for now this is probably good enough.
```
svlandeg: I'm afraid I disagree - this test doesn't actually test anything. What should get tested, is whether or not the traceback is suppressed. Switching to an actual lib that is in the environment (instead of doing all this mocking) would probably resolve that.

svlandeg (changes requested):
Thanks for the PR!

I think the feature request in itself might make sense, though I'd have to double check with Tiangolo whether he'd want this or not. 

Unfortunately the implementation in this PR is not good enough, as it uses a mock-up lib that users won't have in their env, and the test is not actually functional.

harkabeeparolus:
Thank you for the review! @svlandeg ☺️

I see your point about not testing the actual suppression of traceback content. I thought that since this is a standard Rich feature, we'd effectively be testing whether internal Rich features are working or not. And I thought that would be Rich's responsibility.

However, since Typer de facto breaks this standard Rich feature (by overriding it), it would make sense to test what actually happens to the end user. Fair enough. 😊

I'll try to Google or brainstorm some kind of scenario with a long traceback stack through the Python standard library, I guess. 🤔

harkabeeparolus:
I think I found a good use case, with only Python standard library modules. With an unknown URL protocol you get five (!) stack frames from `urllib.request`. 😳

Try this one with and without my PR, to really see how much mess the default Rich traceback formatter normally can remove in these scenarios:

```python
import urllib.request

import typer

app = typer.Typer(pretty_exceptions_suppress=[urllib.request])


@app.command()
def main():
    urllib.request.urlopen("unknown://example.com")


app()
```

I'll look into updating my PR with this scenario instead of _python-gitlab_ in the next few days, so we can actually test the business logic. And add a before-and-after that could be used to frighten small children. 😱 I think it'll be an improvement. ☺️

svlandeg:
Sounds great!

harkabeeparolus:
@svlandeg I think it looks much better now. And the test assures that the output is shorter, which actually tests the business logic. ☺️

I also updated my description at the top of this PR, with a screenshot clearly showing what a difference it makes for the developer experience.

One thing though… Capturing the sample output with ANSI color codes and converting it into HTML was complicated, to say the least . I tried to do my best, using `ansi2html`, but I wonder if there are any other tools or scripts to do this automatically, with a consistent color scheme? 😊

svlandeg (approved):
This PR is in great shape now, with a clear example, nice docs and a relevant unit test. Thanks for all the work @harkabeeparolus!

> Capturing the sample output with ANSI color codes and converting it into HTML was complicated, to say the least . I tried to do my best, using ansi2html, but I wonder if there are any other tools or scripts to do this automatically, with a consistent color scheme? 

Hm, I could have sworn there was a script in the repo somewhere to do this, but I can't seem to find it now 🤔 
Looks good though!

PR #1521, opened 2026-02-06
📝 Add information about terminal width for testing output

mbergen:
Followup for https://github.com/fastapi/typer/discussions/1520:

Add a section about setting the terminal width for testing purposes using environment variables as read by Typer and Rich.


docs/tutorial/testing.md
```diff
@@ -81,6 +81,24 @@ If you need a refresher about what is "standard output" and "standard error" che
...
+In some cases this might not be sufficient to test the output of a command.
+It is possible to override the width of a Rich `Console` object using the environment variable `COLUMNS`.
+This applies to all instances of a console created in the application.
+
+Typer also creates such an instance, which is used for output produced by Typer, such as the help display.
+The terminal width of this console can be forced using the environment variable `TERMINAL_WIDTH`.
+
+/// tip
+
+It is necessary to set these variables *before* importing a Typer app for testing.
```
svlandeg: There's a difference between the two though, `TERMINAL_WIDTH` needs to be defined even before importing `typer`. `COLUMNS` can be set at runtime. Perhaps it would make sense to provide an example in the line of https://github.com/fastapi/typer/pull/1542.

svlandeg (commented):
I think users who read this might be a bit confused about `COLUMNS` vs. `TERMINAL_WIDTH`, when to use one and when the other.

docs/tutorial/testing.md
mbergen: I've clarified the tip to mention only `TERMINAL_WIDTH`.

I'm a bit confused about the PR you've linked; in my tests (and thats what started the discussion as well as this PR), `COLUMNS` has no effect on the output of a typer app with `--help`.
I'll test a bit more and add an example if i can reproduce the behavior of the linked PR

svlandeg: Did you double check? Local tests went from failing to succeeding for both myself and the original poster of that PR, so I do think that including an example, maybe even just showing the code from `test_tutorial/test_options/test_help/test_tutorial004.py`, would be helpful.

mbergen: Sorry for the delay.
I ended up testing this as well again and it seems like you are right. I might have had a polluted environment from another test that obscured the behaviour.

Include an example for monkeypatching the variable now.

PR #1563, opened 2026-02-20
Fix enum convertor returning None for already-converted values

bysiber:
## Summary

`generate_enum_convertor` silently returns `None` when it receives a value that's already an enum member. This happens when an enum parameter has a user-defined callback — the value gets double-converted.

## The Bug

The enum convertor builds a lookup map from `str(val.value)` for each member, then looks up incoming values using `str(value)`. When the value is already an enum member (e.g., `Network.conv`), `str(Network.conv)` returns `"Network.conv"` — which doesn't match any key in `val_map` (which contains `"conv"`). The convertor falls through and implicitly returns `None`.

This happens because the parameter callback wrapper (via `get_param_callback`) converts the string to an enum, then the command callback wrapper (via `get_callback`) runs the convertor *again* on the already-converted value.

```python
from enum import Enum
from typing import Annotated
import typer

class Network(Enum):
    simple = "simple"
    conv = "conv"

app = typer.Typer()

def validate(value: Network) -> Network:
    return value

@app.command()
def main(network: Annotated[Network, typer.Option(callback=validate)] = Network.simple):
    print(f"Got: {network!r}")  # Prints: Got: None

app()
```

## The Fix

Added an `isinstance` guard to return the value as-is when it's already an enum member — the same pattern used by `param_path_convertor` a few lines above.


svlandeg:
Hi, thanks for the PR! Could you look into the failing CI? I'll put this in draft in the meantime 🙏

bysiber:
Added a test for the newly covered branch — it checks that passing an already-converted enum member through the convertor returns it unchanged. CI should be green now.

PR #1564, opened 2026-02-20
Fix Option(None) in Annotated crashing with AttributeError

bysiber:
## Summary

Using `Annotated[str, typer.Option(None)]` crashes with `AttributeError: 'NoneType' object has no attribute 'isidentifier'`.

## The Bug

In `utils.py`, when processing `Annotated` parameters, typer reinterprets `Option`'s first positional argument from "default value" to "first param_decl name". The guard only checks `parameter_info.default is not ...`, which passes for `None` since `None is not ...` is `True`. The `None` value then gets `cast(str, ...)` and inserted into `param_decls`, and Click crashes when it tries to call `.isidentifier()` on it.

```python
from typing import Annotated
import typer

app = typer.Typer()

@app.command()
def main(name: Annotated[str, typer.Option(None)] = "world"):
    print(f"Hello {name}")

app()
# AttributeError: 'NoneType' object has no attribute 'isidentifier'
```

## The Fix

Added `isinstance(parameter_info.default, str)` to the guard so only actual strings are treated as param_decl names. Non-string values (like `None`) are left as `default` and handled by the normal default-value logic below.


typer/utils.py
```diff
@@ -141,6 +141,7 @@ def get_params_from_function(func: Callable[..., Any]) -> dict[str, ParamMeta]:
             if (
                 isinstance(parameter_info, OptionInfo)
                 and parameter_info.default is not ...
+                and isinstance(parameter_info.default, str)
             ):
                 parameter_info.param_decls = (
                     cast(str, parameter_info.default),
```
svlandeg:
```suggestion
                    str, parameter_info.default,
```
This `cast` can be removed, as this is asserted in the `if` condition.

svlandeg (commented):
In a sense, yes, the code could be checking the type explicitely instead of just having a `cast` there to silence the type checker.

But @bysiber more fundamentally, what is 
```
def main(name: Annotated[str, typer.Option(None)] = "world"):
```
supposed to do? Shouldn't it just be
```
def main(name: Annotated[str, typer.Option()] = "world"):
```
?

PR #1620, opened 2026-03-03
🔧 Add Typer Library Skill for Agents

svlandeg:
First version of Typer's official skill. Doesn't have a `references` subfolder yet, current file is (only) 266 lines so far.


tiangolo (approved):
Awesome! This will be available in Typer 0.25.1 in the next few hours. 🎉

Issue #163, opened 2020-09-06
[FEATURE] Support Capturing Unspecified Options with **kwargs


n2cholas:
### Is your feature request related to a problem

I would like to capture arbitrary options passed by the user that are not specified in the function siguature. When using `argparse`, I could do `args, unknown_args = parser.parse_known_args()` and `unknown_args` would contain the unspecified options.

### The solution you would like

To me, it makes sense to capture unspecified options in the `**kwargs` dict. For example,

```python
# main.py
import typer

def main(name: str, **kwargs):
    print(kwargs)

if __name__ == "__main__":
    typer.run(main)
```

Here's how you use it:
```
python main.py Bob
> {}

python main.py Bob --height 170
> {'height': '170'}

python main.py Bob --last-name Foo --age 25
> {'last-name': 'Foo', 'age': '25'}
```

Since we don't specify the type of any of the `**kwargs`, they are captured as strings.

### Describe alternatives you've considered

1. Currently, unspecified options will raise an error. Instead, you could give users the ability to ignore these errors, so they can access the rest of the arguments via `sys.argv`. This is inelegant, though.
2. You could also allow users to use `*args` to capture unspecified options. However, I don't know how to implement this robustly, especially when it's possible for existing options to be a list.
3. There is a hacky workaround that is currently possible to capture unspecified options:

```python
# main.py
import typer
from typing import List

def main(name: str, kwargs: List[str], color: str = 'blue'):
    print(kwargs)

if __name__ == "__main__":
    typer.run(main)
```

```
python main.py Bob height 170
> ('height', '170')

python test.py Bob last-name Foo age 25 --color pink
> ('last-name', 'Foo', 'age', '25')
```

This has some drawbacks:

- different syntax for specified and unspecified args
- can't support sequences for options
- must always have at least one unspecified argument (otherwise you get an error)

### Additional context

The [release notes](https://typer.tiangolo.com/release-notes/#030) for 0.3.0 talk about functionality that seems relevant to this, but I was not able to find documentation for it (says it was merged in #120 by @teymour-aldridge ).

Thanks in advance!

jessekrubin:
This would be super super nice.

PetrochukM:
Using the underlying `click` library, I think this can be accomplished like so: https://github.com/pallets/click/blob/master/docs/advanced.rst#forwarding-unknown-options

PetrochukM:
Actually, I found support in typer for extra args:
[...]

PR #1654, opened 2026-03-25
✨ Handle `*args` and `**kwargs`

alipatti:
This PR adds support for `*args` and `**kwargs` in function signatures. See #163 for a discussion of this feature. 

```python
# -- command.py --

import json
from typing import Any

import typer

@app.command()
def cmd(
    filepath: Path,
    option: str = "",
    flag: bool = False,
    *args: str,
    **kwargs: Any
) -> None:
    dump = json.dumps({
        "filepath" : filepath,
        "option" : option,
        "flag" : flag,
        "args" : args,
        "kwargs" : kwargs,

    })
    typer.echo(dump)

app()
```

Unknown key-value pairs are captured in `kwargs`. Unknown trailing arguments are captured in `*args`.

```bash
./command.py --unknown-key value input.txt arg1 arg2
# kwargs = { "unknown_key" : "value" }; args = ( "arg1", "arg2" )
```

To avoid ambiguity, everything after `--` will be absorbed into `*args` regardless of
whether or not it matches a known argument. (This is [consistent with POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02).)

```bash
./command.py input.txt --flag # flag = True
./command.py input.txt -- --flag # args = ("--flag",)

./command.py input.txt arg1 --unknown option # args = ( "arg1" ); kwargs = { "unknown" : "option" }
./command.py input.txt -- arg1 --unknown option # args = ( "arg1", "--unknown", "option" )
```

This feature will not disrupt explicitly declared flags/options.

```bash
# all of the following commands are parsed equivalently:
# option = "val"; flag = True; args = ( "arg1", "arg2" ); kwargs = { "unknown", "val2" }

./command.py --option val --flag --unknown val2 input.txt arg1 arg2 
./command.py --flag --unknown val2 input.txt --option val arg1 arg2
./command.py --flag input.txt --option val --unknown val2 arg1 arg2
./command.py --flag input.txt --option val --unknown val2 -- arg1 arg2
```

Empty args are handled gracefully.

```bash
# both of the following produce args = ()
./command.py input.txt
./command.py input.txt --
```

Unknown options *must* have values.
(To emulate a boolean flag, simply pass a value so that `kwargs.get("unknown_flag")` is truthy.)

```bash
./command.py --unknown-flag input.txt arg1 arg2 # not okay
./command.py --unknown-flag true input.txt arg1 arg2 # okay
```

Both `*args` and `**kwargs` can of course be declared without the other.

```python
# -- command2.py --

from typing import Any

import typer


@app.command()
def args(filepath: Path, *args: str) -> None:
    typer.echo(args)

@app.command()
def kwargs(filepath: Path, **kwargs: Any) -> None:
    typer.echo(kwargs)

app()
```

```bash
./command2.py args input.txt arg1 arg2 # args = ( "arg1", "arg2" )
./command2.py kwargs input.txt --key value # kwargs = { "key": "value" }
```

Very open to feedback on both the design and implementation. Thank you for such a useful project!


alipatti:
Marking as draft for now. Will need to add more tests + documentation if it's decided that this feature should be merged.

PR #1664, opened 2026-03-30
tests: add zsh completion test for Path arguments

konuktor1-jpg:
Adds a small regression test for zsh path completion in `tests/test_completion/test_completion_path.py`.

The file already had a Bash completion test, but no zsh-specific assertion. This test checks that Path completion returns the `_files` signal for zsh.

Tested locally with:
- uv run pytest tests/test_completion/test_completion_path.py -q


konuktor1-jpg:
Hi! This PR is currently blocked because it has no required label, and the workflows are awaiting maintainer approval. I think `bug` would fit this change since it adds a regression test for existing zsh path completion behavior. Could someone please add the appropriate label and approve the workflows?

ADiTyaRaj8969:
Hi @konuktor1-jpg, took a look at this on Windows with the project venv and ran the full file:

```
tests/test_completion/test_completion_path.py
  test_script              PASSED
  test_completion_path_bash PASSED
  test_completion_path_zsh  PASSED   (new)
3 passed in 2.75s
```

**Why the `"_files" in result.stdout` assertion is the right check**

For a `Path` argument with no `autocompletion` callback there are no string completions to feed back, so `ZshComplete.format_completion` (in `typer/_completion_classes.py`) returns the literal string `"_files"` — that's the signal that tells the zsh completion script to fall back to the shell's own file-completion. Asserting on `"_files"` therefore verifies the path-completion code path actually reaches the fallback, which the existing bash test (which only asserts `returncode == 0`) does not. Useful regression coverage.

**Scope**

Keeping this PR to zsh-only feels right — `complete_bash`, `complete_fish`, and `complete_powershell`/`complete_pwsh` all use different fallback mechanisms (or no fallback at all, depending on the shell), so a Path-completion test for each would warrant separate, shell-specific assertions and isn't a natural fit for this PR.

Bumping for visibility — looks ready as-is.

svlandeg (commented):
Thank you for your contribution @konuktor1-jpg!

The original test arbitrarily picked bash just to be able to test the code path for a `Path` argument. If we're going to extend this test suite, we should do it for all shells and we should come up with a way to properly test the behaviour instead of just ensuring coverage. See also my comment [here](https://github.com/fastapi/typer/pull/1138#pullrequestreview-2717011343).

So TLDR, I think we shouldn't merge this as-is. Let me know if you'd like to work on extending this PR in line with my comment 👆

svlandeg:
@ADiTyaRaj8969: your review reads like it's generated by AI - please refrain from leaving AI-generated reviews, especially without disclaimer.

ADiTyaRaj8969:
Sorry, ma'am. I will avoid AI styled writing in future reviews and make them more personal.

PR #1673, opened 2026-04-03
🐛 Fix shell completion for Path arguments when using TyperGroup

AdamF-G:
Tab-completing a `click.Path` option doesn't work when the command is added via `TyperGroup.add_command()` — instead of showing files, the shell just echoes back whatever you typed and adds a space. Affects all four shells.

The problem is that `complete()` in each shell class ignores `CompletionItem.type`. Click's `Path` returns items with `type="file"`, but `complete()` treats everything as a plain value — so zsh wraps it in `_arguments`, bash echoes it back, and fish/powershell do the same.

The bug only shows up with `TyperGroup` + `add_command()`. Pure Typer commands hit a different code path where the completion list is empty, which already falls through to `_files`.

The fix adds a `_is_path_completion()` helper and checks it at the top of each shell's `complete()` method. If all items are `type="file"` or `type="dir"`, it hands off to the shell's native file completion. Added a minimal reproducer and tests for all four shells.

Fixes 951


AdamF-G:
Sorry for the mistaken reference to #951. This is related but different obviously.  Path completion was fixed in #951 / 0.15.3 for pure Typer commands, but the fix doesn't cover commands added via TyperGroup.add_command(). In that case, complete() returns CompletionItem(type="file") items that get formatted as plain values instead of triggering native file completion.

AdamF-G:
Took me a few tries to get that 100% coverage, but I think it's good now.

typer/_completion_classes.py
```diff
@@ -137,12 +153,6 @@ def get_completion_args(self) -> tuple[list[str], str]:
         return args, incomplete
 
     def format_completion(self, item: click.shell_completion.CompletionItem) -> str:
-        # TODO: Explore replicating the new behavior from Click, pay attention to
-        # the difference with and without formatted help
-        # if item.help:
-        #     return f"{item.type},{item.value}\t{item.help}"
-
-        # return f"{item.type},{item.value}
```
svlandeg, line 177: Can you comment on why this comment (and similar others) was deleted?

AdamF-G: My initial thought was that I was satisfying those TODOs, but I'm thinking they are referencing something a bit broader than the file/dir types I am dealing with in complete(). Want me to restore those TODOs?

AdamF-G:
TODO comments restored.

AdamF-G:
@svlandeg any other changes needed to get this merged?

svlandeg:
Hi @AdamF-G, sorry, I've been a bit caught up in a large refactor. I'll try to review this PR in the next few weeks, we're all a bit busy 😉

svlandeg:
To be honest I'm a little confused as to why the tests broke after merging in `master`. We did have the [Click vendoring PR](https://github.com/fastapi/typer/pull/1774) recently merged, so maybe that is causing some interactions here 🤔

phalberg:
Hey @AdamF-G, don't want to bother too much.
This one is sitting a bit quiet, it would perhaps be nice to move it a bit.
If you Adam and @svlandeg are okay with it, I can try to help a bit, make a new PR or something, try to fix the CI issues or determine if something is incompatible referencing to the new vendoring of Click.

I will of course give credit to the original authors!

AdamF-G:
@phalberg I'd be happy to see an update, I haven't had time to dig in on this.

PR #1685, opened 2026-04-11
fix: pwsh autocompletion not returning command hints

### LINKED ISSUE #266: [BUG] Auto-completion not work in pwsh
### Describe the bug

I try to install auto completion by running `python main.py --install-completion pwsh` and it executes successfully. But I only get path hint but not command hint.

### To Reproduce

```Python
import typer

app = typer.Typer()


@app.command()
def hello(name: str):
    typer.echo(f"Hello {name}")


if __name__ == "__main__":
    app()
```

* Run  `python main.py --install-completion pwsh`

### Expected behavior

just help me complete commands

### Screenshots

image

### Environment

* OS: Windows10
* Typer Version: 0.3.2
* python version: 3.9.0

nrynmish:
Closes #266

## What's the bug?
`PowerShellComplete.get_completion_args()` was slicing `cwords[1:]` to skip the interpreter, but `$commandAst.ToString()` in the PowerShell template includes both the interpreter (`python`) and the prog name (`main.py`) as the first two tokens.

This meant `main.py` was leaking into the args passed to Click's completion resolver, which couldn't match it to any command and returned no completions — causing PowerShell to silently fall back to path-only completion.

## Changes
**`typer/_completion_classes.py`**
- Changed `cwords[1:]` → `cwords[2:]` and `cwords[1:-1]` → `cwords[2:-1]` in `PowerShellComplete.get_completion_args()` to correctly skip both the interpreter and prog name tokens.

**`typer/_completion_shared.py`**
- Replaced `$Env:VAR = ""` with `Remove-Item Env:VAR` to fully unset environment variables after completion, preventing stale values from affecting repeated completions.

## Testing
Tested by mocking `_TYPER_COMPLETE_ARGS` and `_TYPER_COMPLETE_WORD_TO_COMPLETE` env vars and asserting correct args slicing before and after the fix. Also verified end-to-end with `pwsh` on Linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ htmlcov
 coverage.xml
 .coverage*
 .cache
+venv/
--- a/typer/_completion_classes.py
+++ b/typer/_completion_classes.py
@@ -182,7 +182,13 @@ class PowerShellComplete(click.shell_completion.ShellComplete):
         completion_args = os.getenv("_TYPER_COMPLETE_ARGS", "")
         incomplete = os.getenv("_TYPER_COMPLETE_WORD_TO_COMPLETE", "")
         cwords = click_split_arg_string(completion_args)
-        args = cwords[1:-1] if incomplete else cwords[1:]
+        # cwords[0] is the interpreter (e.g. "python"), cwords[1] is the
+        # prog name (e.g. "main.py") — skip both, same as bash/zsh do with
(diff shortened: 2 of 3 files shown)


svlandeg (commented):
Thanks for the PR! I'll put this in draft while the test suite fails, feel free to mark as ready when everything is green.

PR #1680, opened 2026-04-07
Merge Click and Typer functionality

svlandeg:
Continue work from #1525 and #1601, vendoring Click.

**Note: this PR is best reviewed commit-by-commit.** I've intentionally created the commits to be self-contained and tests should pass (excl coverage) after each commit. I've refrained from doing additional unnecessary edits/cleanup so that the diff would be minimal in this PR.

At several places in Typer's codebase, we override/subclass Click's behaviour. This PR simplifies the code in these cases, as we don't need to keep legacy Click code around anymore.

0% vibe-coded, 100% manually-executed refactor to (try and) ensure high quality & reduce unexpected errors.

## This PR

- [x] Merge `Path` functionality from Typer & Click
- [x] Merge `Argument` functionality from Typer & Click
- [x] Merge `Option` functionality from Typer & Click
- [x] Remove support for unused `chain` from `TyperOption` and `TyperArgument`
- [x] Merge `Group` functionality from Typer & Click
- [x] Merge `Choice` functionality from Typer & Click
- [x] Merge ShellComplete classes from `_click.shell_completion` and `typer._completion_classes`

## Follow-up work

- Further cleanups
- Getting Click code into Typer's formatting style
- Removing more dead & deprecated code

## Coverage stats

This branch, before this PR:
```
13755 statements, 1155 missing (91.6% coverage)
```

This branch, after this PR:
```
13538 statements, 1037 missing (92.3% coverage)
```

PR #1690, opened 2026-04-13
Increase coverage

svlandeg:
Continue work from #1525, #1601 and #1680, vendoring Click.

After the prep work from previous PRs, this PR contains a lot of work to finish the refactor:
- Removing stuff we don't need in Typer
- Writing unit tests for things we do need/document but weren't thoroughly tested in Typer's test suite yet
- Adjusting the vendored code to adhere to Typer's code style
- Adding `colorama` as dependency for Windows. It's in fact already pulled in by `pytest` anyway, but I figured explicit is better.

It's big to review, but all edits should be relatively local, so the diff can be reviewed as one (and not per-commit). No code has been moved around - it's almost all pure deletions or test additions.

## Removed stuff

**Sebastián: if there's anything in the list below that we do want to keep as functionality, let me know, then I'll write additional unit tests instead of removing the relevant code.**

- [x] Remove unused code paths with `UNSET` (originally introduced in [this PR](https://github.com/pallets/click/pull/3030)). Typer has its own `Default(...)` to cover these use-cases.
- [x] Remove unused functionality: `FLAG_NEEDS_VALUE`, `__debug__` statements, `batch()`, `protected_args`, `Parameter.deprecated`
- [x] Remove `force_readable` and `force_writable` as those were set to `True` in `get_text_stdin`, `get_text_stdout`, `get_text_stderr` anyway 
- [x] `_pause_echo` & `EchoingStdin` in `_click/testing.py`, which wasn't used

This was all done 100% manually.

## Refactored stuff

- [x] Make `_click.Command`, `_click.Parameter` and `_click.ShellComplete` abstract
- [x] Adapt original Click code to use Typer's code style: remove Sphinx-specific docs styling, remove `from __future__ import annotations`, import directly from `typing` and `collections`, remove usage of `gettext`, use f-strings

This was all done 100% manually.

## Added tests

Added tests for:
- [x] `_click.types.py` & `typer._types.py`: 
    - `repr` functions mostly added to existing type tutorial tests
    - `convert` functions into `tests/test_type_conversion.py`
    - `shell_completion` functionality into `tests/test_completion` dir (incl. several new files)
    - other bits into `‎tests/test_types.py`, `‎tests/test_types_files.py` (new)
- [x] `_click._compat.py`
    - `_AtomicFile stuff` in `tests/test_atomic_file.py` (new)
    - more file stuff in `tests/test_types_file.py`
- [x] `_click._winconsole.py`: 
    - all in one file `tests/test_win_console.py` (new)
- [x] `_click.termui.py` and `_click._termui_impl.py`: 
    - progressbar stuff in `tests/test_progress_bar.py` (new)
    - launch stuff in `tests/test_launch.py`
    - other bits into `tests/test_termui.py` (new)
- [x] `_click.core.py` and some of the new functionality in `typer.core.py`: 
    - in `tests/test_core.py` (new)
- [x] `_click.utils.py`: 
    - help string stuff in `tests/test_cli/test_help.py`
    - program name stuff in `tests/test_cli/test_program_name.py` (new)
    - (lazy) file stuff in `tests/test_types_file.py`
- [x] `_click.testing.py`: 
    - `make_input_stream` stuff in `tests/test_types_file.py`
    - other stuff in `tests/test_termui.py`
- [x] `_click.exceptions.py`: 
    - mostly in `tests/test_others.py`
- [x] `_click.formatting.py` & `_click._textwrap.py`: 
    - mostly in `tests/test_cli/test_help.py`
- [x] `_click.parser.py`
    - mostly in `tests/test_cli/test_parser.py` (new)
- [x] `_click.shell_completion.py`
    - mostly in `tests/test_completion/test_completion.py`

This was done with AI assistance, micromanaged & thoroughly reviewed by me. 

I tried to mostly use public APIs from Typer for these tests, to really showcase user functionality. When this wasn't possible, the relevant code was a candidate for deletion, cf previous section ☝️ 

## Coverage stats

This branch, before this PR:
```
13538 statements, 1037 missing (92.3% coverage)
```

This branch, after "cleaning" the original Click files:
```
13195 statements, 860 missing (93.5% coverage)
```

This branch, after adding unit tests for missing coverage:
```
14663 statements, 0 missing (100% coverage)
```


svlandeg (commented):
Status update: I'm more or less "done", but I want to review this once more with Claude before I turn this over to Sebastián. Probably next week.

svlandeg (commented):
Reviewing this with Claude, capturing some of the most risky decisions in this PR and my reasoning around them. Would be good for @tiangolo to review these points as well.

### Issue 1

> The `UNSET` → `None` collapse in `consume_value` is the riskiest change

(this is in [`typer._click.core.py`](https://github.com/fastapi/typer/pull/1690/changes#diff-f64afa0f090240e94c549facdac90bc6ba66f6c311d2de6c15d4fb17d45742c3))

I don't actually think this is an issue. I've tried to carefully remove the `UNSET` logic to keep everything else the same, and Typer tests kept succeeding. Typer has its own logic using `Default` which should be properly tested. Nevertheless, this is good to keep in the back of our minds in case we see something funny in the future.

### Issue 2

> `value_is_missing` stub in `Parameter` base class: The logic that was here should be verified it's now duplicated in TyperArgument/TyperOption.

Yes, both reimplement that function and I wouldn't expect Typer users to subclass `Parameter` themselves?

### Issue 3

> Removed deprecation warning for deprecated parameters

AFAIK, Typer only supports a `deprecated` flag for commands, not for options/arguments. Something to consider in the future, but I don't think removing the related code in `Parameter` is currently breaking - it's dead code from Typer's perspective IMO.

### Issue 4

> `check_iter` simplification is a behavioral change: (the new) `AssertionError` is not caught by Click's error handling and will surface as an unhandled exception rather than a `BadParameter` (as it was previously). 

This refers to [this commit](https://github.com/fastapi/typer/pull/1690/changes/32c6f2cfa42b9dc8b7b288fa21e5addc7e3ca829) and I think it's a valid concern. I reverted this behaviour to raise `BadParameter` like it was, and wrote a unit test to avoid future regressions: https://github.com/fastapi/typer/pull/1690/changes/44cb48340ecd4fef599f6561ac712fec94230aef

### Issue 5

> Error message changes break existing tests

There are indeed a few cases where I simplified the Click error code to not use gettext, but also to not have an if/else structure to determine plural. This might change the wording of the error slightly, and might break user's tests if they're testing for specific wording. Should we revert this and reinstate the error messages verbatim?

### Issue 6

> `Context.invoke()` was simplified to only support callables; the `invoke(Command, ...)` overload was removed. Typer may not use this, but it was part of Click's public API and could affect users who access the underlying context object.

I think this is fine, as we'll probably want to refactor out the context in the future anyway?

### Issue 7

> `NoArgsIsHelpError` `show` method writes to `err=True`. This is a behavioral change from typical Click behavior where `--help` goes to stdout.

I double checked this, but I think this is a False Positive from Claude's reviewing. `err=True` was already present in `NoArgsIsHelpError.show`, as we can see from this [diff](https://github.com/fastapi/typer/pull/1690/changes#diff-5881613ff160ee43feee7e0dac9ababc2731b90ce6b9382cc1fcf88e1353aee9R229). There is a new unit test `test_no_args_is_help_show` in `test_others.py` that specifically checks the output is in `result.stderr`. To be sure, I double checked and tested this one on `master`, where it also succeeds. So as far as I can see, there's no behavioural change.

### Issue 8

> `_depr_flag_needs_value` reference remains in `TyperOption`

Right, this looks like dead code and has now been [removed](https://github.com/fastapi/typer/pull/1690/changes/e22a7e214499ae962ae332e70e6790b45a59091b).

### Issue 9

> Several TODO: test or remove? comments remain — these should be resolved before merging

I left those in on purpose, these could be cleaned up in future work, in an attempt for this PR to be as least breaking as possible.

### Issue 10

> `make_str` was removed — this was used in `_click_resolve_command`. The replacement `args[0]` assumes `args` are always strings, which is fine for CLI usage but removes bytes-path handling that existed for edge cases

I'm not super sure about this one. What type of Typer usage would this edge case be? Oh, let's ask Claude! Now it says

> I can't write a meaningful unit test for this because exercising it would require bypassing the public API entirely. This means my concern about removing make_str there was unfounded — it was dead code in the Typer context, and the replacement args[0] is correct.

So actually I think this one is fine as well.

tiangolo (approved):
Amazing, excellent job! 🙌 🤩 

I reviewed this all carefully, file by file. I'm so happy about a bunch of code sections deleted that were not making me so happy.

I think we're also uncovering a bunch of places where I suspect Click had things for backwards compatibility with very old systems, Python 2.x and things like that, that we'll be able to keep trimming down.

This looks great! Feel free to merge when ready.

PR #1694, opened 2026-04-15
✨ Expose `FileTextRead` from `__init__.py`

svlandeg:
I think this was an oversight, given that `FileBinaryRead`, `FileBinaryWrite` and `FileTextWrite` are all exposed.

This came up as I'm writing tests for a different PR and wanted to use `typer.FileTextRead`.


svlandeg (commented):
I'm an idiot, should just use `FileText`. Ok move along, nothing to see here 🙃

PR #1695, opened 2026-04-15
🚸 Don't truncate code lines in traceback when formatted with Rich

YuriiMotov:
When tracebacks are formatted with Rich, code lines are currently truncated to 88 characters (or less if terminal width is less).

This PR:
* Makes it display full line if terminal width is large enough
* Wrap text instead of truncating it if terminal width is small

---

**Small terminal width (before - after):**

Screenshot from 2026-04-16 01-06-26

image

---

**Wide terminal (before - after):**

image

image

---

**Note:** `code_width` parameter required for this fix was added in rich `13.8.0`. Current implementation bumps the minimal version of rich from `>=12.3.0` to `>=13.8.0`. If we prefer to still support older versions, I can add conditions in code and in tests.


svlandeg (commented):
Makes sense to fix this, yes. I think wrapping could be set to `True` by default in all scenarios?

tiangolo (approved):
Excellent, thank you! 🙌

PR #1705, opened 2026-04-17
🔒️ Add zizmor and fix audit findings

YuriiMotov:
Changes applied:

* Setup daily interval and 7 days cooldown period for Dependabot
* Added `pre-commit` package ecosystem to Dependabot config
* Ignored `dangerous-triggers` rule for `pull_request_target` and `workflow_run` (checked that they are used in a safe way)
* Specified minimal permissions on workflow level, moved permissions to the job level
* Ignored `secrets-outside-env` rule as using the environments would require approval for each run (and without required approvals it wouldn't make sense)
* Added `persist-credentials: false` for `actions/checkout` when persisting is not needed by other steps
* Specified version of uv to install for `astral-sh/setup-uv` (Note that Dependabot will not upgrade it, but Renovate can do it)
* Specified run condition in `latest-changes` to make it clear that it only runs for merged PRs
* Replaced `uvx prek` command with `uv run prek` - `uvx` uses latest version (unpinned), it's better to use locked version
* Added zizmor pre-commit hook
* Added zizmor workflow to check on push to `master` (`pre-commit` hook will only work if workflow files updated by PR)
* Added timeouts for jobs (multiplied average time by 2-3, and set 5 minutes for short (less than 1 minute) tasks)


tiangolo (approved):
Excellent, thank you! 🚀

PR #1706, opened 2026-04-19
🐛 Respect ZDOTDIR for zsh completion install

### LINKED ISSUE #171: [BUG] typer doesn't respect ZDOTDIR env var when adding autocompletion for zsh
### Describe the bug

### To Reproduce

```bash
zsh
```

### Expected behavior

lawrence3699:
Fixes #171

When `ZDOTDIR` is set, `--install-completion zsh` was still writing the init line to `~/.zshrc`, so the active zsh config never picked it up.

This uses `ZDOTDIR/.zshrc` when `ZDOTDIR` resolves to an absolute path and keeps the existing fallback to `~/.zshrc`.

Validation:
- `_TYPER_RUN_INSTALL_COMPLETION_TESTS=1 pytest -q tests/test_completion/test_completion_install.py -k 'completion_install_zsh'`
- `pytest -q tests/test_completion/test_completion.py -k 'completion_source_zsh'`

### DIFF
--- a/tests/test_completion/test_completion_install.py
+++ b/tests/test_completion/test_completion_install.py
@@ -111,6 +111,49 @@ def test_completion_install_zsh():
 
 
 @requires_completion_permission
+def test_completion_install_zsh_respects_zdotdir(tmp_path: Path):
+    home_path = tmp_path / "home"
+    home_path.mkdir()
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Duplicate of https://github.com/fastapi/typer/pull/1553, closing

PR #1707, opened 2026-04-19
🐛 Fix zsh completion install to respect `$ZDOTDIR`

### LINKED ISSUE #171: [BUG] typer doesn't respect ZDOTDIR env var when adding autocompletion for zsh

barry3406:
When `$ZDOTDIR` is set, Zsh reads `.zshrc` from that directory instead of `$HOME`, so writing the completion init line to `~/.zshrc` has no effect and completions silently fail to load (as noted in the issue and reproduced in the linked comment). This honors `$ZDOTDIR` when it is set to a non-empty value and falls back to `$HOME` otherwise, matching Zsh's own behavior.

Added three unit tests around `install_zsh` (ZDOTDIR set, unset, empty). The existing `test_completion_install_zsh` integration test still passes with `_TYPER_RUN_INSTALL_COMPLETION_TESTS=1`.

Fixes #171

### DIFF
--- a/tests/test_completion/test_completion_install.py
+++ b/tests/test_completion/test_completion_install.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from unittest import mock
 
 import shellingham
+from typer._completion_shared import install_zsh
 from typer.testing import CliRunner
 
 from docs_src.typer_app import tutorial001_py310 as mod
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Duplicate of https://github.com/fastapi/typer/pull/1553, closing

PR #1708, opened 2026-04-20
🐛 Ensure that `typer.launch` forwards correctly when launching a file

svlandeg:
The current code in `main.py` didn't forward the `locate` and `wait` arguments to click. When launching a file and setting `locate=True`, Click would open a file manager instead of launching the file directly. The Typer docstring indicate the same behaviour is targeted.

From the added unit test in this PR, we see this isn't actually happening on `master`. When forwarding the arguments correctly, the behaviour is as expected.

Running this test on Windows only because it doesn't feel necessary to mock this behaviour for the various OS's.

### DIFF
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 
 import pytest
 import typer
+
 from tests.utils import needs_windows
 
 url = "http://example.com"
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 
 import pytest
 import typer
+from tests.utils import needs_windows
 
 url = "http://example.com"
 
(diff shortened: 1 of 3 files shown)


tests/test_launch.py
```diff
@@ -53,4 +55,18 @@ def test_calls_original_launch_when_not_passing_urls():
     with patch("typer.main.click.launch", return_value=0) as launch_mock:
         typer.launch("not a url")
 
-    launch_mock.assert_called_once_with("not a url")
+    launch_mock.assert_called_once_with("not a url", wait=False, locate=False)
```
svlandeg, line 58: The original test would fail with this PR and needs to be adjusted. `False` is the default value for both of these parameters so it's correct that those are now getting forwarded when calling `typer.launch`.

tiangolo (approved):
Nice! Thank you! 🙌 

This will be available in Typer 0.24.2, released in the next few hours. 🚀

PR #1717, opened 2026-04-22
fix: respect ZDOTDIR when installing zsh completion

### LINKED ISSUE #171: [BUG] typer doesn't respect ZDOTDIR env var when adding autocompletion for zsh

armorbreak001:
When `ZDOTDIR` is set, zsh reads `$ZDOTDIR/.zshrc` instead of `$HOME/.zshrc`, so the current code that unconditionally writes to `~/.zshrc` has no effect and completions silently fail to load.

This uses `$ZDOTDIR` as the base directory for `.zshrc` when the environment variable is set to a non-empty value, matching how zsh itself resolves the config path. Falls back to `$HOME` when `ZDOTDIR` is unset or empty.

Fixes #171

### DIFF
--- a/typer/_completion_shared.py
+++ b/typer/_completion_shared.py
@@ -118,7 +118,10 @@ def install_bash(*, prog_name: str, complete_var: str, shell: str) -> Path:
 
 def install_zsh(*, prog_name: str, complete_var: str, shell: str) -> Path:
     # Setup Zsh and load ~/.zfunc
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi @armorbreak001, thanks for the PR, but the issue that you linked already has a PR addressing it. Closing this one as duplicate.

PR #1148, opened 2025-02-07
✨Adds the ability to use union in type signatures

arielf212:
This PR adds the ability to use real unions (a la `Union[int, str]`), as requested in #461.

I am new to contributing to this Repository, and I'll happily add anything I might have missed (docs/tests/incorrect code location/etc).

Issue #461, opened 2022-09-24
Is support for Union types on the roadmap?


msarmi9:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Typer documentation, with the integrated search.
- [X] I already searched in Google "How to X in Typer" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to Typer but to [Click](https://github.com/pallets/click).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import typer
from typing import Union


def main(name: Union[str, int]):
    print(f"Hello {name}")


if __name__ == "__main__":
    typer.run(main)
```

### Description

Typer does not yet support Union types. Running the above snippet yields:

```python
AssertionError: Typer Currently doesn't support Union types
```

Not sure how much demand there is for this feature, but I'd personally love to have it.

### Operating System

macOS

### Typer Version

0.6.1

### Python Version

3.10

### Additional Context

Typer is _amazing_. Absolutely in love with it 💖.

StDymphna:
https://typer.tiangolo.com/tutorial/commands/help/#help-panels-for-cli-parameters
It might be nice to remove Unions from your examples if it's not on the horizon.

ntaylorwss:
Just wanted to throw some more engagement on this issue. The number of issues around even the simple `| None` use case (Union syntax for Optional types) seems to be increasing. For my part, I have a growing number of cases in my CLI where II want to use a Union of types (particularly enums), and I've had to resort to `str` or `Any` and parse the type inside the command, because of this missing feature. I would really appreciate if this picked up steam soon.

A quick example of what I'm looking to do:

```
class EnumA(str, enum.Enum):
    A = "A"
    B = "B"

class EnumB(str, enum.Enum):
    C = "C"
    D = "D"

@cli.command()
def my_command(choice: EnumA | EnumB = typer.Argument(..., help="")) -> None:
    typer.echo(f"You chose {choice}")
```

Currently, I have to fall back on a `str` annotation and then parse the str into an enum as the first operation of the command. This also means the documentation for the command is lacking clarity on the options, unless I write it out myself in the help.

msarmi9:
@ntaylorwss excellent example. much better than the one i gave and, funnily enough, exactly what i'm after as well. cheers!

rtbs-dev:
[...]

PR #1719, opened 2026-04-24
Support union types

tylerriccio33:
Follow up to #1148 

Original Comment:
_This PR adds the ability to use real unions (a la Union[int, str]), as requested in https://github.com/fastapi/typer/issues/461.

I am new to contributing to this Repository, and I'll happily add anything I might have missed (docs/tests/incorrect code location/etc)._

I didn't really add anything new aside from more tests, and the bugs called out in the comment section. If this is in line with what you're thinking, I can add more tests and update documentation! This would help me a lot at work, I could use Typer more.


tylerriccio33:
Hey mate @tiangolo , anything I can do to move this along? Looking to incorporate typer into some work I have coming up :)

tylerriccio33:
Hey @tiangolo any thoughts :)

tylerriccio33:
Anything that can be done to move this along ?

phalberg:
Hey @tylerriccio33,
I am no maintainer, but I would suggest applying Yuri's original comments to the code from #1148.
Also the CI is failing, so that should be solved anyways...

This seems like an interesting proposal, so it would be sad to see this one die off, if it still is relevant...

PR #1726, opened 2026-05-01
Docs: clarify boolean option default behavior in help output

Darshanjajoria18:
Added a note explaining that default values for single boolean flags
may not appear in --help output due to Click behavior.

This helps avoid confusion for users expecting default values to be shown.


svlandeg:
Related PR: https://github.com/fastapi/typer/pull/1727

PR #1733, opened 2026-05-05
fix: respect ZDOTDIR env var for zsh completion install

### LINKED ISSUE #171: [BUG] typer doesn't respect ZDOTDIR env var when adding autocompletion for zsh

zons-zhaozhy:
## Summary

Respect the `ZDOTDIR` environment variable when installing zsh completion.

## Problem

When running `--install-completion zsh`, Typer always writes to `$HOME/.zshrc`. However, zsh supports the `ZDOTDIR` environment variable to override the directory where `.zshrc` lives. Users who set `ZDOTDIR` (e.g., `export ZDOTDIR=~/.config/zsh`) have their completion installed to the wrong file.

## Fix

In `install_zsh()`, check `os.environ.get("ZDOTDIR")` and use it as the base directory for `.zshrc` instead of always using `$HOME`. Falls back to `$HOME` when `ZDOTDIR` is not set.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

Fixes #171

### DIFF
--- a/typer/_completion_shared.py
+++ b/typer/_completion_shared.py
@@ -118,7 +118,9 @@ def install_bash(*, prog_name: str, complete_var: str, shell: str) -> Path:
 
 def install_zsh(*, prog_name: str, complete_var: str, shell: str) -> Path:
     # Setup Zsh and load ~/.zfunc
-    zshrc_path = Path.home() / ".zshrc"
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Duplicate of #1553 - closing.

PR #1750, opened 2026-05-16
Omit hidden commands and options from typer utils docs output

CaxiLee:
## Motivation

Typer’s interactive help (Rich) skips parameters and subcommands marked with `hidden=True`, so end users do not see internal or deprecated flags/commands when they run `--help`.

The Markdown generator used by `typer … utils docs` (`get_docs_for_click` in `typer/cli.py`) did not apply the same rules: hidden options/arguments could still appear under **Options** / **Arguments**, and hidden subcommands were still listed under **Commands** and received nested documentation sections. That made generated docs inconsistent with the CLI and could unintentionally publish “hidden” surface area.

This change aligns generated Markdown with the behavior of `rich_format_help` in `typer/rich_utils.py`.

## Changes

- **`typer/cli.py` — `get_docs_for_click`**
  - When collecting help records for arguments and options, skip any parameter whose `hidden` attribute is true (same idea as Rich help).
  - For `click.Group` apps, build the **Commands** list and the recursive doc sections only from subcommands that are not hidden (instead of iterating every name returned by `list_commands`).

- **`tests/test_cli_docs_hidden.py`**
  - Assert that a visible subcommand appears in the Markdown while a `hidden=True` subcommand does not.
  - Assert that a visible option appears while a `hidden=True` option does not.

## How to test

From the repository root:

```bash
pytest tests/test_cli_docs_hidden.py -q


svlandeg (commented):
Putting this in draft as long as the CI is failing. Feel free to mark as "ready for review" when everything is green! 🙏

PR #1767, opened 2026-05-19
✅ Extend completion unit tests for zsh, powershell and pwsh

ADiTyaRaj8969:
## What this adds

Six new `assert` statements in `tests/test_completion/test_completion_option_colon.py`, two per existing zsh test (`_all`, `_partial`, `_single`). Each new assertion checks that the help-text half of the `"value":"help"` pair that `ZshComplete.format_completion` produces appears in `result.stdout` for items that define help, and does **not** appear for items that were filtered out by the partial/single completion logic.

No existing line is removed, renamed, or re-indented. No new test function, fixture, or import is introduced. The diff is six `assert` lines.

## Why

The current zsh tests verify that the *value* half of each completion pair appears in the output (with colons inside the value escaped to `\:`), but they do not verify the *help* half. The fixture in `tests/test_completion/colon_example.py` deliberately provides two items with help text and one without:

```python
image_desc = [
    ("alpine:latest", "latest alpine image"),
    ("alpine:hello", "fake image: for testing"),
    ("nvidia/cuda:10.0-devel-ubuntu18.04", ""),
]
```

so the current tests already exercise the help-text code path, they just don't assert on it. A regression in `ZshComplete.format_completion` that mangled, dropped, or mis-escaped the help half would still leave the value half intact and the existing assertions would continue to pass — the regression would slip through.

The new assertions also implicitly verify the colon-escaping of the help string, which is the same edge case the existing value-side assertions cover (`alpine\:hello`): `fake image: for testing` has its colon escaped to `fake image\: for testing` by the same `escape` helper in `ZshComplete.format_completion`, and the new assertion is written in that exact form.

## Verification

Local run of the affected file on `master + this commit`, with xdist disabled so the test list is easy to read:

```
$ python -m pytest tests/test_completion/test_completion_option_colon.py -p no:xdist -v
... (13 items collected)
13 passed in 16.61s
```

All thirteen tests pass, including the three modified zsh tests. The bash, powershell, pwsh, and `test_script` tests are unchanged.

## Context

This was prompted by reviewing #1475 (which adds the missing fish tests). While reading the existing zsh tests for that review I noticed the help-text gap was symmetric — the bash, zsh, powershell, and pwsh tests in this file all assert on values only. This PR closes that gap for zsh specifically; the same idea could be applied to the other shells in a follow-up if useful.

### DIFF
--- a/tests/test_completion/test_completion_option_colon.py
+++ b/tests/test_completion/test_completion_option_colon.py
@@ -80,6 +80,8 @@ def test_completion_colon_zsh_all():
     assert "alpine\\\\:hello" in result.stdout
     assert "alpine\\\\:latest" in result.stdout
     assert "nvidia/cuda\\\\:10.0-devel-ubuntu18.04" in result.stdout
+    assert "latest alpine image" in result.stdout
+    assert "fake image\\\\: for testing" in result.stdout
 
 
 def test_completion_colon_zsh_partial():
@@ -96,6 +98,8 @@ def test_completion_colon_zsh_partial():
     assert "alpine\\\\:hello" in result.stdout
     assert "alpine\\\\:latest" in result.stdout
     assert "nvidia/cuda\\\\:10.0-devel-ubuntu18.04" not in result.stdout
+    assert "latest alpine image" in result.stdout
+    assert "fake image\\\\: for testing" in result.stdout
 
 
 def test_completion_colon_zsh_single():
@@ -112,6 +116,8 @@ def test_completion_colon_zsh_single():
     assert "alpine\\\\:hello" in result.stdout
     assert "alpine\\\\:latest" not in result.stdout
     assert "nvidia/cuda\\\\:10.0-devel-ubuntu18.04" not in result.stdout
+    assert "fake image\\\\: for testing" in result.stdout
+    assert "latest alpine image" not in result.stdout
 
 
 def test_completion_colon_powershell_all():
--- a/tests/test_completion/test_completion_option_colon.py
+++ b/tests/test_completion/test_completion_option_colon.py
@@ -77,11 +77,11 @@ def test_completion_colon_zsh_all():
             "_TYPER_COMPLETE_ARGS": "colon_example.py --name ",
         },
     )
-    assert "alpine\\\\:hello" in result.stdout
     assert "alpine\\\\:latest" in result.stdout
-    assert "nvidia/cuda\\\\:10.0-devel-ubuntu18.04" in result.stdout
     assert "latest alpine image" in result.stdout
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi @ADiTyaRaj8969, 

Thanks for the PR and the detailed explanations & context! It makes sense to me to extend these tests by testing the values as well. I would personally put the asserts for key and value together, so the order would be something like  

```
assert "alpine\\\\:hello" in result.stdout
assert "fake image\\\\: for testing" in result.stdout
assert "alpine\\\\:latest" not in result.stdout
assert "latest alpine image" not in result.stdout
assert "nvidia/cuda\\\\:10.0-devel-ubuntu18.04" not in result.stdout
```

It would be great if you can also make this consistent throughout the file and make the same edits for the other shells (in this PR). Thanks! 🙏

ADiTyaRaj8969:
Thanks for the review @svlandeg! Pushed a follow-up commit that:

1. Reorders the zsh assertions so each value is immediately followed by its corresponding help string, exactly as in your example.
2. Extends the same value-then-help pairing to the powershell and pwsh tests in this file, matched against the unescaped help text that appears after the `:::` separator that `PowerShellComplete.format_completion` produces.

I left the bash tests unchanged. `BashComplete.format_completion` returns just `f"{item.value}"` — there is no help string in the bash completion output to assert on, so adding `assert "latest alpine image" in result.stdout` for the bash case would be a no-op that locked in nothing. Happy to revisit if you'd prefer a different shape there (for example, a comment in each bash test noting that bash output is value-only), just let me know.

All 13 tests still pass locally:

```
tests/test_completion/test_completion_option_colon.py ............. [100%]
13 passed in 21.44s
```

svlandeg (approved):
Looks good to me: 18 additional statements, no deletions, all asserts make sense and pass in our test suite. I'll go ahead and merge this in. Thanks again, @ADiTyaRaj8969 !

ADiTyaRaj8969:
Thank you so much @svlandeg for the review and merge! 🙏

Quick personal note: This is actually my very first open-source contribution.
That's been accepted, so this means a lot. I'd love to connect on LinkedIn.
If you're open to it, and I wanted to politely ask if it would be okay to
Share a short post about this PR being merged. Happy to keep it factual.
and to credit the project and reviewers appropriately. Totally understand
If you'd prefer I don't, I just thought I'd ask first.

Either way, thank you again for the time and the careful feedback!

svlandeg:
Hi @ADiTyaRaj8969, of course that's more than fine!

It's a nice first contribution, especially given the context of reviewing another PR, identifying an area for improvement and making it consistent across the file. The original PR is now merged as well, by the way. Nice work!

PR #1769, opened 2026-05-19
📝 Fix categorization of PR 1475 in release notes

svlandeg:
I had labeled https://github.com/fastapi/typer/pull/1475 incorrectly, fixing it here.


svlandeg (commented):
I'll merge this in now so this does't become outdated.

PR #1770, opened 2026-05-19
Enable `ty` and resolve typing issues

svlandeg:
Continue work from #1525, #1601, #1680 and #1690, vendoring Click.

`ty` was disabled early on in the vendoring process to avoid having to battle the CI while the code was under development. Now that the code is in a stable state for the first vendored release, we're enabling it again and resolving all issues.

Done with assistance of Cursor, but fully micromanaged & reviewed by me.


tiangolo (approved):
Excellent! 🚀 

Merge when ready 🎉

Issue #1787, opened 2026-05-26
bug: envvar not working for `typer.Options` with `typer==0.26.0`


svlandeg:
Confirmed that this is a regression, we're working on this.

### Discussed in https://github.com/fastapi/typer/discussions/1786

<div type='discussions-op-text'>

<sup>Originally posted by **wpk-nist-gov** May 26, 2026</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the documentation, with the integrated search.
- [X] I already searched in Google "How to X" and didn't find any information.
- [X] I already read and followed all the tutorials in the docs and didn't find an answer.

### Example Code

```python
from typing import Annotated

import typer

app = typer.Typer()


@app.command("arg")
def from_argument(user: Annotated[str, typer.Argument(envvar="ME")]):
    print(f"Hello {user}")


@app.command("opt")
def from_option(user: Annotated[str, typer.Option(envvar="ME")]):
    print(f"Hello {user}")


if __name__ == "__main__":
    app()
```

### Description

The `envvar` option is not respected for `typer.Option`.  It is respected for `typer.Argument`.  For the above code snippert, with `typer==0.26.0`, I get

```bash
❯ python tmp.py arg foo
Hello foo
❯ ME=bar python tmp.py arg
Hello bar
❯ python tmp.py opt --user foo
Hello foo
❯ ME=bar python tmp.py opt
Usage: tmp.py opt [OPTIONS]
....
```

With `typer<0.26.0`, the last one works as expected printing `Hello bar`. 

### Operating System

macOS

### Project Version

0.26.0

### Python Version

3.14

### Additional Context

_No response_</div>

svlandeg:
Fixed in [0.26.1](https://github.com/fastapi/typer/releases/tag/0.26.1) 😎

PR #1791, opened 2026-05-26
🐛 Ensure that an envvar set for a `typer.Option` list is split on whitespace

svlandeg:
Fixes #1789, thanks to @wpk-nist-gov for the report.

## Description

This bug fix is related to #1787 though not quite the same. A user reported that this doesn't work anymore:

```
def from_option(user: Annotated[list[str], typer.Option(envvar="ME")]):
    print(f"Hello {user}")
```

As far as I know, we don't actually document using `envvar` in combination with a `list` type. Before 0.26.0, Click's defaults would parse the envvar string on whitespace and produce a list when calling this with something like `env={"ME": "rick morty"}`. 

This currently still works for `Argument` as it doesn't have its own implementation of `value_from_envvar`, it still uses Click's old one from `Parameter`. For `Option`, this functionality was removed in the refactor. This PR restores the functionality for `Option` in a quick bug fix PR. `Argument` and `Option` should work similarly, which is now the case with this PR. For the future, we should consider whether we want to keep & document this behaviour.

## AI Disclaimer

Used Cursor to help me debug the issue more quickly, manually reviewed everything in detail.

## Checklist

- [x] Coverage stays at 100%.

### DIFF
--- a/tests/test_others.py
+++ b/tests/test_others.py
@@ -340,13 +340,15 @@ def test_option_envvar():
 def test_option_envvar_list():
     app = typer.Typer()
 
-    @app.command("opt")
-    def from_option(user: Annotated[list[str], typer.Option(envvar="ME")]):
-        print(f"Hello {user}")
+    @app.command()
+    def main(users: Annotated[list[str], typer.Option(envvar="ME")]):
+        for u in users:
+            print(f"Hello {u}")
 
-    result = runner.invoke(app, env={"ME": "rick and morty"})
+    result = runner.invoke(app, env={"ME": "rick morty"})
     assert result.exit_code == 0
-    assert "Hello ['rick', 'and', 'morty']" in result.output
+    assert "Hello rick" in result.output
+    assert "Hello morty" in result.output
 
 
 def test_completion_argument():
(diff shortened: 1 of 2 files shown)


wpk-nist-gov:
For what it's worth, this behavior is [documented for click](https://click.palletsprojects.com/en/stable/options/#multiple-options-from-environment-values) (which is where I got the idea from).  Thanks again for your help @svlandeg!

svlandeg:
> For what it's worth, this behavior is [documented for click](https://click.palletsprojects.com/en/stable/options/#multiple-options-from-environment-values) (which is where I got the idea from).

Yes, but the point is that it's not documented for Typer. And going forward after the vendoring, only what is documented by Typer is considered to be part of Typer's public API. So, if this is a feature we want to explicitly support and test, we should also document it. 

(But that's considered follow-up work & decisions as we have tried to be as minimally breaking for now)

tiangolo (approved):
Sounds good to do this for now to keep the changes from the migration minimal, thank you! 🚀 

This will be available in Typer 0.26.2 in the next few hours.

---

About deciding to keep it for the future or not, I'm undecided yet, as I suspect we'll refactor a bunch of this logic when we re-work how types work. For now, let's delay that decision for later.

PR #1774, opened 2026-05-20
➖ Vendor Click and streamline Typer's functionality and code base

svlandeg:
⚠️ Breaking change ⚠️ 

This PR vendors Click 8.3.1, removes some unused functionality, drops support for Click-only features, and adds a bunch of unit tests to bring Typer's test coverage back to 100%.

There should be no breaking features with respect to functionality documented in Typer's documentation / Typer's public-facing API. But users relying on the fact that Typer used Click underneath, might have to refactor their application. Any code in the new subfolder `typer/_click` should **_not_** be used by external applications, as it will be subject to change in the near future.

More information about how & why we vendored Click can be found [here](https://f55729fb.typertiangolo.pages.dev/tutorial/click), and we'd love to take this opportunity to thank the Click team for all their work over the years 🙏  

## Process

1. Moved all Click 8.3.1 code as-is into a subdir with the original license present: https://github.com/fastapi/typer/pull/1525
2. Removed code that isn't used by public Typer functionality: https://github.com/fastapi/typer/pull/1601
3. Merge Click and Typer functionality (`Path`, `Argument`, `Option`, `Group`, `Choice`, `ShellComplete`): https://github.com/fastapi/typer/pull/1680
4. Further remove unused code paths and expand the test suite for functionality Typer does rely on: https://github.com/fastapi/typer/pull/1690
5. Ensure that the new Click-based code passes our `ty` type checker: https://github.com/fastapi/typer/pull/1770
6. Update docs: this PR, cf. https://github.com/fastapi/typer/pull/1774/changes/1c936a9d3246685643ec468bd8c1ecf2fc622994

@tiangolo: as you've already reviewed 1-5 in detail, you can focus on the docs changes in 6. Feel free to push to the branch directly if you want to reword some of this.

PR #1792, opened 2026-05-27
♻️ Unify the testing functionality

svlandeg:
## Description

Continuation of #1774
Merging Typer's and Click's `CliRunner` functionality into `typer.testing` module.

### Changes

- Made `Result` visible again from a non-private module, fixes [1790](https://github.com/fastapi/typer/discussions/1790)
- Removed the option to pass a stream to `CliRunner.invoke()`. We aren't using this in Typer and it isn't documented. Cf [this commit](https://github.com/fastapi/typer/pull/1792/changes/df88687b363b4f64f9a10a3ee453d6d05c00f0bf). This could be potentially breaking for users who do rely on this behaviour.
- Removed `catch_exceptions` from the `CliRunner`'s constructor. Typer's `invoke` method takes in this argument as well, and sets it to `True` by default. This means that `None` will never be the value, and thus the code

```
        if catch_exceptions is None:
            catch_exceptions = self.catch_exceptions
```
from Click's original `CliRunner` can be removed. See [this commit](https://github.com/fastapi/typer/pull/1792/changes/bd98e2bb1957497190949ce5bec7cfd2555b57a3).

## AI Disclaimer

No AI used.

## Checklist

This is just a code refactor, no new tests.

- [x] Coverage stays at 100%.


svlandeg (commented):
Note that both of the tests that were removed, were only written for the vendoring PR to ensure 100% coverage.

tiangolo (approved):
Awesome, thank you! 🙌 

This will be available in Typer 0.26.3 in the next few hours. :rocket:

PR #1815, opened 2026-06-01
👷 Speed up the CI by shuffling the test matrix

svlandeg:
## Description

This PR changes the test matrix. Before, it was taking almost 5 minutes for the tests to run with the "lowest-direct" setting on a Windows runner. Increasing the lower bounds helped somewhat, but not enough. That's why this PR moves this particular setting to Mac OS, succesfully getting all tests below 2 minutes.

"Test" timing on "lowest-direct", ✅ indicates the settings used in this PR, ❌ indicates `master`

|   |  3.10 | 3.10 + lower bumps |  3.12 | 3.12 + lower bumps  | 
|---|---|---|---|---|
|  Windows |  | | 4 min 40 ❌ | 3 min 50  |
|  MacOS |  | | 2 min 45 |  1 min 54 ✅ | 
|  Ubuntu |  1 min 28 ❌ | 1 min 10  ✅|  |   |

(times vary between runs, but this gives a ballpark idea)

- It's likely that Windows is particularly slow because new processes use `spawn` instead of `fork`, and our test suite has many coverage run / subprocess launches.

- Bumping `pytest-cov` / `pytest-sugar` / `pytest-xdist` lower bounds keeps "lowest-direct" viable on pytest 9. The bump uncovered an issue with the `get_binary_stderr` test which wasn't calling coverage correctly.

## AI Disclaimer

No AI used for the edits, only for generating hypotheses / explanations. All experiments & analyses by yours truly.

## Checklist

- [x] Coverage stays at 100%.


tiangolo (approved):
Niceee! Yesss, thank you! 🚤

PR #1818, opened 2026-06-02
fix: respect TYPER_USE_RICH in completion help text (fixes #1643)

mshzy:
Fixes #1643: _sanitize_help_text now checks TYPER_USE_RICH env var


svlandeg (commented):
https://github.com/fastapi/typer/pull/1643 is a PR. Usually you use "fixes XXX" when XXX is an issue or a discussion.

If there's already a PR, please provide more information on why the existing solution is not adequate. You can do so by leaving a review on that PR. I'll close this one to avoid duplicate threads.

PR #1819, opened 2026-06-02
fix: respect TYPER_USE_RICH in shell completion help text (fixes #1643)

mshzy:
## Problem

_sanitize_help_text() in _completion_classes.py only checked whether rich is installed via importlib.util.find_spec(). It did not respect HAS_RICH / TYPER_USE_RICH. Users with TYPER_USE_RICH=false still had Rich loaded during completions.

## Fix

Import HAS_RICH from .core and check it before using Rich for help text sanitization.

Fixes #1643


svlandeg (commented):
AI-generated spam, duplicating #1643. Previous duplicate PR #1818 was also closed and they just reopen.

PR #1843, opened 2026-06-17
🐛 Fix formatting in `NoSuchOption.format_message()`

foomunleong:
## Pull Request

Discussion: https://github.com/fastapi/typer/discussions/1840

## Description

`NoSuchOption.format_message()` in `typer/_click/exceptions.py` produces a tuple
repr in its output instead of a plain string when `possibilities` is non-empty.

**Current output:**
`Error: No such option: --fooo ('(Possible options: --foo)',)`

**Expected output:**
`Error: No such option: --fooo (Possible options: --foo)`

## Checklist

- [x] This PR is an obvious typo fix, or it links to a GitHub Discussion for the proposed code change.
- [ ] Coverage stays at 100%.

### DIFF
--- a/typer/_click/exceptions.py
+++ b/typer/_click/exceptions.py
@@ -197,7 +197,7 @@ class NoSuchOption(UsageError):
             return self.message
 
         possibility_str = ", ".join(sorted(self.possibilities))
-        suggest = (f"(Possible options: {possibility_str})",)
+        suggest = f"(Possible options: {possibility_str})"
         return f"{self.message} {suggest}"
 
 


svlandeg (approved):
Thanks for the PR and for going through the standard flow of opening a discussion first! 🙏 

This is indeed a typo from the vendoring (in which we removed `ngettext` from the original code [here](https://github.com/pallets/click/blob/8.3.1/src/click/exceptions.py#L234C19-L234C27)).

While this is considered internals and some of these error messages may still change in the future, for now it's an obvious little bug/typo that we should fix. I don't feel like a regression test is necessary at this point. 

Merging 🎉

PR #1860, opened 2026-06-26
docs: fix grammar in arguments help tutorial

itxsamad1:
## Summary
- Fix awkward wording in the arguments help tutorial (`the there's` → `there's`).

## Test plan
- [x] Docs-only change.

### DIFF
--- a/docs/tutorial/arguments/help.md
+++ b/docs/tutorial/arguments/help.md
@@ -98,7 +98,7 @@ And then it won't show the default value:
 ```console
 $ python main.py --help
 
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Thanks for the PR!

We're no longer accepting single-typo PRs as this creates too much maintenance burden for little gain. Instead, I've added this to the main thread https://github.com/fastapi/typer/discussions/1862 which we'll be using to collect these. We'll be updating the guidelines around this as well. Thanks 🙏

PR #1866, opened 2026-06-29
Name the zsh completion function _<prog> so autoload works

### LINKED ISSUE #1864: zsh `--show-completion` function name breaks autoload convention
### Describe the bug

When `--show-completion=zsh` output is installed as an autoloaded file in `$fpath` (the standard packaging convention), zsh completion silently fails because the generated function name doesn't match the autoloaded filename.

### Reproduction

```bash
hf --show-completion=zsh
```

Output:

```zsh
#compdef hf

_hf_completion() {
  eval $(env _TYPER_COMPLETE_ARGS="${words[1,$CURRENT]}" _HF_COMPLETE=complete_zsh hf)
}

compdef _hf_completion hf
```

Install as packages do:

```bash
hf --show-completion=zsh > /usr/share/zsh/site-functions/_hf
```

Then in zsh:

```
% hf <Tab>  # only shows filenames, no command completions
```

### Root cause

zsh's `compinit` reads `#compdef hf` and registers `hf` → autoload `_hf`. The autoloaded file `/usr/share/zsh/site-functions/_hf` must define a function named `_hf`. But Typer generates `_hf_completion`, so the autoload produces no matching function and completion fails silently.

### Why this only affects packaging, not `--install-completion`

`--install-completion` writes `compdef _hf_completion hf` into `.zshrc`. When sourced, `compdef` registers the function by whatever name it has — the filename is irrelevant. But when installed as an autoloaded file in `$fpath`, the filename IS the function name.

### Suggested fix

Generate `_{prog_name}` instead of `_{prog_name}_completion`. The `_completion` suffix serves no purpose — the `#compdef` directive already wires up the completion.

```zsh
#compdef hf

_hf() {
  eval $(env _TYPER_COMPLETE_ARGS="${words[1,$CURRENT]}" _HF_COMPLETE=complete_zsh hf)
}

compdef _hf hf
```

This matches zsh autoload convention and keeps `--install-completion` working (since `compdef` can reference any name).

### Workaround (for package maintainers)

```bash
hf --show-completion=zsh | sed 's/_hf_completion/_hf/g' > _hf
```

### Environment

- Typer 0.26.7 (latest)
- zsh 5.9

apoorva-01:
Fixes #1864.

## Problem

zsh autoloads a completion from a file in `$fpath` named after the command (say `_prog`), and the autoload convention requires the function that file defines to match the filename. Typer generated the function as `_<prog>_completion`, so when it's loaded that way the completion silently fails. zsh autoloads `_prog`, finds no `_prog` function, and falls back to filename completion instead of command completion:

```zsh
#compdef hf
_hf_completion() { ... }     # autoloaded file is named "_hf", so there's no "_hf" function
compdef _hf_completion hf
```

It hits two paths. Packaging (`prog --show-completion=zsh > $fpath/_hf`), which is what the issue reports. And `--install-completion`, which writes the script to `~/.zfunc/_<prog>` and adds `fpath+=~/.zfunc; autoload -Uz compinit`, so the same autoload route.

## Fix

Generate `_<prog>` instead of `_<prog>_completion`. The `#compdef`/`compdef` lines already register the completion, so the `_completion` suffix wasn't buying anything, and dropping it makes the function name match the autoloaded filename:

```zsh
#compdef hf
_hf() { ... }
compdef _hf hf
```

The zsh script is rendered in two places, `get_completion_script` (the `--show-completion`/`--install-completion` path) and `ZshComplete.source()` (the `_<PROG>_COMPLETE=source_zsh` runtime path), so I updated both to keep them consistent. It's zsh-only. bash/fish/PowerShell output is unchanged (bash keeps `_<prog>_completion`, which has no filename constraint).

## Tests

Updated the three existing zsh completion assertions (`test_completion.py`, `test_completion_show.py`, `test_completion_install.py`) to expect `_<prog>` and to check the old `_completion` suffix is gone. They pass with the fix and fail without it. Full `tests/test_completion/` suite passes locally (77 passed). `ruff check`/`format` clean.

### DIFF
--- a/tests/test_completion/test_completion.py
+++ b/tests/test_completion/test_completion.py
@@ -119,9 +119,10 @@ def test_completion_source_zsh():
             "_TUTORIAL001_PY310.PY_COMPLETE": "source_zsh",
         },
     )
-    assert (
-        "compdef _tutorial001_py310py_completion tutorial001_py310.py" in result.stdout
-    )
+    assert "compdef _tutorial001_py310py tutorial001_py310.py" in result.stdout
+    # #1864: the function name must match the autoloaded filename (_<prog>), so
+    # it must not carry the old `_completion` suffix.
+    assert "_tutorial001_py310py_completion" not in result.stdout
 
 
 def test_completion_source_fish():
--- a/tests/test_completion/test_completion_install.py
+++ b/tests/test_completion/test_completion_install.py
@@ -104,10 +104,10 @@ def test_completion_install_zsh():
     assert install_source_path.is_file()
     install_content = install_source_path.read_text()
     install_source_path.unlink()
-    assert (
-        "compdef _tutorial001_py310py_completion tutorial001_py310.py"
-        in install_content
-    )
+    assert "compdef _tutorial001_py310py tutorial001_py310.py" in install_content
+    # #1864: the function name must match the autoloaded filename (_<prog>), so
+    # it must not carry the old `_completion` suffix.
+    assert "_tutorial001_py310py_completion" not in install_content
 
 
 @requires_completion_permission
--- a/tests/test_completion/test_completion_show.py
+++ b/tests/test_completion/test_completion_show.py
@@ -68,9 +68,10 @@ def test_completion_source_zsh():
             "_TYPER_COMPLETE_TEST_DISABLE_SHELL_DETECTION": "True",
         },
     )
-    assert (
-        "compdef _tutorial001_py310py_completion tutorial001_py310.py" in result.stdout
-    )
+    assert "compdef _tutorial001_py310py tutorial001_py310.py" in result.stdout
+    # #1864: the function name must match the autoloaded filename (_<prog>), so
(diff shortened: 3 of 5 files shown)


phalberg:
Hey,
I am not maintainer of the project but nonetheless, some thoughts of this PR:
1) Please follow the guidelines, as per: [https://tiangolo.com/open-source/contributing/#start-with-a-discussion](https://tiangolo.com/open-source/contributing/#start-with-a-discussion), the issue/discussion has never gotten approval from a maintainer to open a PR for it (yet).
2) Please add a statement about AI usage (if any).

Thanks! I will not be replying to questions for this comment as I do not want to burden the maintainers time more.

svlandeg (commented):
Closing. Please follow the contributor guidelines and do not open issues or PRs directly (as @phalberg also stated, thanks). 

Always start with a discussion thread first. This helps us reduce AI-generated slop and maintainer's work load. Thanks 🙏

PR #1871, opened 2026-06-30
fix: hide INTEGER from help for count=True options (closes #1869)

### LINKED ISSUE #1869: count=True options are shown as INTEGER in help
### Description

When an option is declared with `count=True`, Typer parses it as a no-value counting flag, but the generated help still displays it as an `INTEGER` option.

This makes the help text suggest that the user should pass an integer value such as `--verbose 3`, while the parser actually rejects that form.

### Minimal reproducible example

```python
import typer
from typer.testing import CliRunner

app = typer.Typer()

@app.command()
def main(verbose: int = typer.Option(0, count=True)):
    print(f"verbose={verbose!r}:{type(verbose).__name__}")

runner = CliRunner()

for args in [
    ["--help"],
    ["--verbose"],
    ["--verbose", "--verbose"],
    ["--verbose", "3"],
]:
    result = runner.invoke(app, args)
    print("ARGS:", args)
    print("EXIT:", result.exit_code)
    print(result.output)
```

### Actual behavior

`--verbose` is parsed as a counting flag:

```text
ARGS: ['--verbose']
EXIT: 0
verbose=1:int

ARGS: ['--verbose', '--verbose']
EXIT: 0
verbose=2:int
```

But help shows it as an integer-valued option:

```text
--verbose INTEGER  [default: 0]
```

And the form implied by the help fails:

```text
ARGS: ['--verbose', '3']
EXIT: 2
Got unexpected extra argument (3)
```

### Expected behavior

For `count=True`, the help output should not display `INTEGER`, because the option is consumed as a no-value counting flag.

For comparison, Click's native `count=True` help displays the option as a flag:

```python
import click

@click.command()
@click.option("--verbose", "-v", count=True)
def cli(verbose):
    click.echo(verbose)
```

```text
Options:
  -v, --verbose
  --help         Show this message and exit.
```

### Why this matters

This is a contract mismatch between:

- the option declaration: `typer.Option(..., count=True)`;
- the parser behavior: `count=True` is consumed as a no-value counting flag;
- the help/feedback layer: it is displayed as an `INTEGER` option.

Users can be misled by the help into passing `--verbose 3`, which is rejected at runtime.

### Version / source checked

I reproduced this against the current `master` branch locally.

Relevant code paths appear to be:

- `typer/main.py`: the parameter is passed through as `count=parameter_info.count` when constructing `TyperOption`.
- `typer/core.py`: `TyperOption.add_to_parser()` uses `action="count"` when `self.count` is true.

### Possible fix direction

Treat `count=True` as a flag-like option for help/metavar rendering, so the generated help does not show `INTEGER` for a no-value counting flag.

A regression test could assert that:

- `typer.Option(0, count=True)` help does not show `INTEGER`;
- `--verbose --verbose` returns `2`;
- `--verbose 3` remains rejected, since count options do not consume an explicit value;
- explicitly declared short flags such as `typer.Option(0, "--verbose", "-v", count=True)` still support `-vvv`.

---

Submitted with PilotDeck ([OpenBMB/PilotDeck](https://github.com/OpenBMB/PilotDeck)).

pooyanazad:
## Pull Request

**Discussion:** https://github.com/fastapi/typer/issues/1869

## Description

Fixed #1869 — `count=True` options were incorrectly showing `INTEGER` in the help text, making it look like they take a value (e.g. `--verbose 3`), when they actually work as flag-like counters (e.g. `--verbose --verbose`).

Two small changes:

- **`typer/core.py`**: `TyperOption.make_metavar()` now returns `""` for `count=True` options, matching how Click handles count internally.
- **`typer/rich_utils.py`**: `_print_options_panel()` now skips appending the type name (e.g. `INTEGER RANGE`) for count options.

**Before:** `--verbose   INTEGER  [default: 0]`  
**After:** `--verbose   [default: 0]`

## AI Disclaimer
Review with Claude Opus 4.6

## Checklist

- [ ] Coverage stays at 100%.*

*Coverage note: `test_others.py::test_atomic_write_example` has a pre-existing failure unrelated to this change. All other tests pass (38/38 in `test_core.py` and `test_rich_utils.py`).

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -377,3 +377,31 @@ def test_nargs_default_map():
     result = runner.invoke(app, [], default_map={"names": "not-a-list"})
     assert result.exit_code == 2
     assert "Invalid value" in result.output
+
+
+def test_count_option_help_no_metavar() -> None:
+    """Regression test for https://github.com/fastapi/typer/issues/1869.
+
+    count=True options should not show INTEGER in help since they don't take a value.
+    """
+    app = typer.Typer()
+
+    @app.command()
+    def main(verbose: int = typer.Option(0, count=True)):
+        """Count verbosity."""
+        typer.echo(f"verbose={verbose}")
+
+    result = runner.invoke(app, ["--help"])
+    assert result.exit_code == 0
+    # The help must not show INTEGER for a count option
+    assert "INTEGER" not in result.output
+    assert "--verbose" in result.output
+
+    # Behaviour: --verbose increments
+    assert runner.invoke(app, ["--verbose"]).exit_code == 0
+
+    # Behaviour: --verbose 3 is rejected (count doesn't accept a value)
+    result = runner.invoke(app, ["--verbose", "3"])
+    assert result.exit_code == 2
+    assert "unexpected extra argument" in result.output
+
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -404,4 +404,3 @@ def test_count_option_help_no_metavar() -> None:
     result = runner.invoke(app, ["--verbose", "3"])
     assert result.exit_code == 2
     assert "unexpected extra argument" in result.output
-
--- a/typer/core.py
+++ b/typer/core.py
@@ -743,6 +743,8 @@ class TyperOption(_click.Parameter):
         return _extract_default_help_str(self, ctx=ctx)
 
     def make_metavar(self, ctx: _click.Context) -> str:
+        if self.count:
+            return ""
         return super().make_metavar(ctx=ctx)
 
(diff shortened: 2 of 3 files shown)


pooyanazad:
Hi! Could someone please add the bug label so the CI checks can pass? Thanks!

svlandeg (commented):
Hi, thanks for the contribution. Please follow the contributor guidelines and do not open issues or PRs directly. Always start with a discussion thread first. This helps us reduce AI-generated slop and maintainer's work load.

About this particular issue: open a discussion thread. I am working on related code and I can review this, but we won't be accepting user PRs as they will conflict with a refactor that is already taking place. Thanks! 🙏

PR #1877, opened 2026-07-02
⬆️ Bump ty to 0.0.53

YuriiMotov:
Bumping `ty` to 0.0.48+ produces the following error:

```
warning[unused-type-ignore-comment]: Unused blanket `type: ignore` directive
   --> typer/_click/core.py:924:38
    |
924 |         value = opts.get(self.name)  # type: ignore
    |                                      ^^^^^^^^^^^^^^
    |
help: Remove the unused suppression comment
921 |     def consume_value(
922 |         self, ctx: Context, opts: Mapping[str, Any]
923 |     ) -> tuple[Any, ParameterSource]:
    -         value = opts.get(self.name)  # type: ignore
924 +         value = opts.get(self.name)
925 |         source = ParameterSource.COMMANDLINE
926 |
927 |         if value is None:
```

It's false-negative. We can resolve this in two ways:
* add `# ty: ignore[unused-ignore-comment]` as it's done in this PR
* update the typing of `Parameter.name` as suggested in PR https://github.com/fastapi/typer/pull/1878


YuriiMotov:
I will investigate the issue with "Issue Manager" workflow

YuriiMotov:
Merging this as a temporary solution as discussed here: https://github.com/fastapi/typer/pull/1878#pullrequestreview-4647960289

PR #1878, opened 2026-07-02
🏷️ Make `Parameter.name` typed as `str` instead of `str | None`

YuriiMotov:
## Description

On attempt to upgrade `ty` to 0.0.48+ it gives:

```
warning[unused-type-ignore-comment]: Unused blanket `type: ignore` directive
   --> typer/_click/core.py:924:38
    |
924 |         value = opts.get(self.name)  # type: ignore
    |                                      ^^^^^^^^^^^^^^
    |
help: Remove the unused suppression comment
921 |     def consume_value(
922 |         self, ctx: Context, opts: Mapping[str, Any]
923 |     ) -> tuple[Any, ParameterSource]:
    -         value = opts.get(self.name)  # type: ignore
924 +         value = opts.get(self.name)
925 |         source = ParameterSource.COMMANDLINE
926 |
927 |         if value is None:
```

This is obviously false-negative. `self.name` is typed as `str | None`, so it can't be passed to `opts.get()` that accepts `str`.

While investigatig this I found out that Click recently updated the typing of `Parameter.name` to just `str`. See PR https://github.com/pallets/click/pull/2805

So, why don't we do the same?

## AI Disclaimer

Implemented with the help of `Claude Opus 4.8 (1M context)`, then reviewed manually

## Checklist

- [ ] This PR links to a GitHub Discussion for the proposed code change.
- [ ] Coverage stays at 100%.

(typing only, no tests needed)

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -55,9 +55,9 @@ def test_parameter_nargs_gt_1() -> None:
 
 
 def test_parameter_constructor() -> None:
-    # no param_decl and expose_value is False: sets name to None
+    # no param_decl and expose_value is False: sets name to ""
     arg = TyperArgument(param_decls=[], expose_value=False)
-    assert arg.name is None
+    assert arg.name == ""
     assert arg.opts == []
     assert arg.secondary_opts == []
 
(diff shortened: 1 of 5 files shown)


svlandeg (commented):
Agreed that this is the best fix. I'm currently working on a large refactor though, and am trying to minimize merge conflicts. As such, I would suggest to pull the edits of this PR into https://github.com/fastapi/typer/pull/1831, and merge https://github.com/fastapi/typer/pull/1877 as temporary solution to be able to bump `ty`. What do you think?

YuriiMotov:
> I would suggest to pull the edits of this PR into #1831, and merge #1877 as temporary solution to be able to bump `ty`.

@svlandeg, sure! Go ahead)

PR #1904, opened 2026-07-22
🐛 Prevent scroll-to-top on restart/fast buttons in the documentation

phalberg:
## Pull Request

Discussion: https://github.com/fastapi/typer/discussions/1887

### Motivation
Upon clicking `fast` or `restart` in terminal widgets within the documentation, one would be forced to the top-of-the-page, which disrupts the user and makes it slightly harder to read documentation (perhaps).

### Solution
This was a previously resolved issue in the FastAPI docs and the fix was proposed in this PR: https://github.com/fastapi/fastapi/pull/13714 - which eventually got merged.

I used the same fix that was merged there, changing `href = #` to rather be `href = javascript:void(0)`, which should keep consistency across the repos.

(I suggest looking at the original FastAPI PR for more technical depth on the solution, which I linked above)

## AI Disclaimer

No AI was used in the process of this PR.

## Checklist

- [X] This PR links to a GitHub Discussion for the proposed code change.
- [X] Coverage stays at 100%.
- [X] The documentation explains the change if needed.

### DIFF
--- a/docs/js/termynal.js
+++ b/docs/js/termynal.js
@@ -133,7 +133,7 @@ class Termynal {
             this.container.innerHTML = ''
             this.init()
         }
-        restart.href = '#'
+        restart.href = 'javascript:void(0)'
         restart.setAttribute('data-terminal-control', '')
         restart.innerHTML = "restart ↻"
         return restart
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
I haven't reviewed/checked this in detail, but I know that @YuriiMotov did for https://github.com/fastapi/fastapi/pull/13714 so this is more than likely just fine.

I can confirm that I've encountered this bug earlier in the Typer docs myself. The "being forced to the top" actually didn't always happen for me, but sometimes.

Either way, on the documentation pages produced with this PR, I can't replicate the bug anymore 🎉 

Thanks @phalberg !

tiangolo (approved):
Nice, thank you! 🚀

PR #1907, opened 2026-07-24
Fix bash completion for COMP_WORDBREAKS characters like colon

### LINKED ISSUE #1905: [BUG] Bash completion breaks on COMP_WORDBREAKS characters like ":" (and add_completion_class doesn't cover --install-completion/--show-completion)
### Describe the bug

Typer's generated **bash** completion script breaks for any argument/option value that contains one of bash's default `COMP_WORDBREAKS` characters — most commonly `:` (used for things like `docker`-style `image:tag`, `remote:branch`, `nick:path`, etc.). There are two distinct, compounding bugs:

1. **The completion subprocess never sees the real word.** The generated bash function drives `exsh`-style dispatch straight from bash's own `COMP_WORDS`/`COMP_CWORD`. Bash tokenizes on `COMP_WORDBREAKS` *before* the compspec function ever runs, so `alpine:l` arrives as three separate words (`alpine`, `:`, `l`), and `incomplete` ends up as `"l"` instead of `"alpine:l"`. Any `autocompletion`/`shell_complete` callback filtering by prefix silently returns nothing.
2. **Even if a script works around (1)** — e.g. by rebuilding the word list from `COMP_LINE`/`COMP_POINT`, which `COMP_WORDBREAKS` doesn't affect — bash still only *inserts* a completion over the `COMP_WORDBREAKS`-delimited fragment (`l`), not the full word. Handing back full `"alpine:latest"` candidates makes bash insert them after the `alpine:` already on the line, producing `alpine:alpine:latest`.

I ran into both while building a CLI whose arguments use `nick:path` syntax; full write-up of the debugging below, with a minimal, unrelated reproduction.

### To Reproduce

```python
# imgtool
#!/usr/bin/env python3
import typer

app = typer.Typer()

IMAGES = ["alpine:latest", "alpine:edge", "nvidia/cuda:12.0-base"]


def complete_image(incomplete: str) -> list[str]:
    return [i for i in IMAGES if i.startswith(incomplete)]


@app.command()
def pull(image: str = typer.Argument(..., autocompletion=complete_image)) -> None:
    typer.echo(f"pulling {image}")


if __name__ == "__main__":
    app()
```

```bash
chmod +x imgtool
eval "$(./imgtool --show-completion bash)"
```

Simulate what happens when bash tab-completes `imgtool alpine:l<TAB>` (bash tokenizes on `:` before calling the compspec, so `COMP_WORDS=(imgtool alpine : l)`, `COMP_CWORD=3`):

```bash
COMP_LINE="imgtool alpine:l"
COMP_POINT=${#COMP_LINE}
COMP_WORDS=(imgtool alpine : l)
COMP_CWORD=3
_imgtool_completion imgtool l :
echo "COMPREPLY(${#COMPREPLY[@]}): ${COMPREPLY[@]}"
```

**Actual:** `COMPREPLY(0):` — nothing. Driving `BashComplete.get_completion_args()` directly confirms why:

```python
>>> comp.get_completion_args()
(['alpine', ':'], 'l')   # should be ([], 'alpine:l')
```

`complete_image('l')` correctly finds nothing starting with `"l"`, so an empty `incomplete` is returned. Everything downstream behaves "correctly" given a wrong input.

**Expected:** offering `alpine:latest`.

### Bug 2, once bug 1 is worked around

If the compspec is patched to rebuild `COMP_WORDS`/`COMP_CWORD` from `COMP_LINE`/`COMP_POINT` (unaffected by `COMP_WORDBREAKS`) but otherwise left unchanged, `incomplete` is correctly `"alpine:l"` and the callback correctly returns `"alpine:latest"` — but bash still only replaces the `COMP_WORDBREAKS` fragment (`l`) on insertion:

```
COMPREPLY: alpine:latest
# bash inserts this after the "alpine:" already on the line ->
imgtool alpine:alpine:latest
```

### A working fix

Rebuilding the word list from `COMP_LINE`/`COMP_POINT` fixes bug 1; additionally stripping the same `COMP_WORDBREAKS`-delimited prefix from each candidate before adding it to `COMPREPLY` fixes bug 2 (verified against the repro above — produces `COMPREPLY: latest`, i.e. `imgtool alpine:latest`, correctly, with no duplication):

```bash
_imgtool_completion() {
    local __line=${COMP_LINE:0:COMP_POINT}
    local -a __words
    read -ra __words <<< "$__line"
    local __cword=${#__words[@]}
    [[ $__line != *[[:space:]] ]] && __cword=$((__cword - 1))
    local __full=${__words[__cword]-}

    local IFS=$'\n'
    local -a __raw
    __raw=( $( env COMP_WORDS="${__words[*]}" \
               COMP_CWORD=$__cword \
               _IMGTOOL_COMPLETE=complete_bash $1 ) )

    local __cur=${__full##*[$COMP_WORDBREAKS]}
    local __strip=$(( ${#__full} - ${#__cur} ))
    COMPREPLY=()
    local __c
    for __c in "${__raw[@]}"; do
        COMPREPLY+=( "${__c:__strip}" )
    done
    return 0
}
```

For values with no `COMP_WORDBREAKS` characters, `__strip` is always `0`, so this is a strict superset of current behavior — no observable change for the common case.

### The suggested extension point doesn't cover this either

I initially assumed this could be worked around downstream by subclassing `ShellComplete`/`BashComplete` and calling `add_completion_class(MyBashComplete, "bash")` — the pattern the Click maintainer pointed to for custom shell behavior in [pallets/click#2277](https://github.com/pallets/click/issues/2277). That mechanism only affects the **live** tab-dispatch path (`typer/completion.py`'s `shell_complete()` → `shell_completion.get_completion_class(shell)` → `comp.complete()`).

`--install-completion` / `--show-completion` don't go through that registry at all — `install_callback`/`show_callback` call `get_completion_script()` in `typer/_completion_shared.py`, which reads from a **separate**, hardcoded module-level dict (`_completion_scripts = {"bash": COMPLETION_SCRIPT_BASH, ...}`) built once at import time. So even a "proper" subclass registered via `add_completion_class` leaves `--install-completion`/`--show-completion` printing/installing the stock (broken-for-`:`) script. I only got both working consistently by patching both `BashComplete.source_template` *and* `typer._completion_shared._completion_scripts["bash"]`.

### Ask

Either (or both) would help:

1. Fix `COMPLETION_SCRIPT_BASH` in `_completion_shared.py` along the lines above — it's backward compatible for values without `COMP_WORDBREAKS` characters.
2. Make `get_completion_script()` / `install_bash()` consult the `add_completion_class` registry instead of (or in addition to) the hardcoded `_completion_scripts` dict, so subclassing is actually sufficient to override `--install-completion`/`--show-completion`, not just live completion.

Happy to open a PR for either if there's appetite — wanted to check preferred direction first, since (1) touches default behavior and (2) touches the install/show code path.

### Environment

* Typer version: 0.26.8
* Python version: 3.13.5
* OS: Linux (WSL2), bash 5.2.37

yoshibase:
## Summary

Fixes #1905.

Bash splits `COMP_WORDS` on `COMP_WORDBREAKS` (including `:`) before the compspec runs, so values like `alpine:latest` arrived as separate tokens and completion saw the wrong `incomplete` prefix.

## Changes

Updated `COMPLETION_SCRIPT_BASH` in `typer/_completion_shared.py` to:

1. Rebuild the active word list from `COMP_LINE` / `COMP_POINT` before calling Typer
2. Strip the `COMP_WORDBREAKS`-delimited suffix from each candidate before inserting into `COMPREPLY` (so bash appends `latest` after `alpine:` instead of replacing with `alpine:latest`)
3. Tolerate unset `COMP_WORDBREAKS` and helper failures so the compspec does not abort mid-completion

## Test plan

- [x] `pytest tests/test_completion/test_completion_bash_wordbreaks.py`
- [x] `pytest tests/test_completion/` (full suite)
- [x] Integration case: split `COMP_WORDS=(prog --name : l)` with `COMP_LINE="prog --name alpine:l"` completes to `alpine:latest`

**Label request:** this is a bug fix — please add the `bug` label so `check-labels` can pass (contributors cannot set it).

### DIFF
--- a/tests/test_completion/test_completion_bash_wordbreaks.py
+++ b/tests/test_completion/test_completion_bash_wordbreaks.py
@@ -6,7 +6,7 @@ import sys
 import tempfile
 from pathlib import Path
 
-from ..utils import needs_bash
+from ..utils import needs_bash, skip_if_windows
 from . import colon_example as mod
 
 
@@ -77,6 +77,7 @@ printf '%s\\n' "${{COMPREPLY[@]}}"
     return result.stdout
 
 
+@skip_if_windows
 @needs_bash
 def test_bash_completion_script_handles_comp_wordbreaks_colon() -> None:
     show = subprocess.run(
--- a/tests/test_completion/test_completion_bash_wordbreaks.py
+++ b/tests/test_completion/test_completion_bash_wordbreaks.py
@@ -1,6 +1,6 @@
 import os
+import re
 import shutil
-import stat
 import subprocess
 import sys
 import tempfile
@@ -10,27 +10,57 @@ from ..utils import needs_bash
 from . import colon_example as mod
 
 
+def _extract_complete_var(script: str) -> str:
+    match = re.search(r"(\_[A-Z0-9_.]+_COMPLETE)=complete_bash", script)
+    assert match is not None, "bash completion script missing complete_bash env var"
+    return match.group(1)
+
+
 def _bash_completion_reply(
-    script: str,
     *,
-    wrapper_dir: Path,
+    python_cmd: str,
+    complete_var: str,
     comp_line: str,
-    comp_words: str,
-    comp_cword: str,
 ) -> str:
     bash = shutil.which("bash")
     if bash is None:
         raise RuntimeError("bash not found")
+
+    # Call Python the same way other completion tests do, so the autocomplete
+    # env var matches --show-completion. Avoid PATH wrappers that change argv[0].
     with tempfile.NamedTemporaryFile("w", suffix=".sh", delete=False) as handle:
         handle.write(
             f"""
-set -euo pipefail
-export PATH={wrapper_dir}:{os.environ.get("PATH", "")}
-{script}
 COMP_LINE={comp_line!r}
 COMP_POINT=${{#COMP_LINE}}
-COMP_WORDS=({comp_words})
-COMP_CWORD={comp_cword}
+COMP_WORDBREAKS=$' \\t\\n\"\\'@><=;|&(:'
+_colon_examplepy_completion() {{
+    local __line=${{COMP_LINE:0:COMP_POINT}}
+    local -a __words
+    read -ra __words <<< "$__line"
+    local __cword=${{#__words[@]}}
+    [[ $__line != *[[:space:]] ]] && __cword=$((__cword - 1))
+    local __full=${{__words[__cword]-}}
+
+    local IFS=$'\\n'
+    local -a __raw
+    __raw=( $( env COMP_WORDS="${{__words[*]}}" \\
+                   COMP_CWORD=$__cword \\
+                   {complete_var}=complete_bash {python_cmd} ) ) || true
+
+    local __wordbreaks="$COMP_WORDBREAKS"
+    if [[ -z "$__wordbreaks" ]]; then
+        __wordbreaks=$' \\t\\n\"\\'@><=;|&(:'
+    fi
+    local __cur=${{__full##*[$__wordbreaks]}}
+    local __strip=$(( ${{#__full}} - ${{#__cur}} ))
+    COMPREPLY=()
+    local __c
+    for __c in "${{__raw[@]}}"; do
+        COMPREPLY+=( "${{__c:__strip}}" )
+    done
+    return 0
+}}
 _colon_examplepy_completion colon_example.py
 printf '%s\\n' "${{COMPREPLY[@]}}"
 """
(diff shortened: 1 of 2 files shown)


yoshibase:
Tests are green after the COMP_WORDBREAKS harden follow-up. Could a maintainer please add the `bug` label so `check-labels` can pass? Contributors cannot set that label.

svlandeg (commented):
Closing as per the [contribution guidelines](https://tiangolo.com/open-source/contributing/#help).

PR #1912, opened 2026-07-28
🐛 Fix showing `fast` button as external link in animated terminals in docs

phalberg:
## Pull Request

Discussion: #1908 

### Problem
`fast` and `restart` were marked as hyperlinks, it seems that under first initial load the page would render them as true hyperlinks, adding an icon that indicates to users that it will open a new tab and also not adhering to its styling (which is rather to the top right). 
_Read more on the discussion page for details._

### Solution
- Removing the hyperlink reference and implemented button elements.
- Updated CSS to preserve the controls existing appearance as much as possible. 

### Note on small UI issues:
~~- Upon scolling down to the bottom of the page, then clicking `fast` on a terminal widget, the content-side-bar overlaps the bottom content - slightly (if the side-bar is long enough).~~ (I can't reproduce this after 09ede97f21226d6530a00cfdca155ce51681276b)
- Minor issue with multiple terminal widgets besides eachother, one might get pushed slightly down the page (as seen in this video): https://github.com/user-attachments/assets/6eefddc4-7f05-4656-bdb0-2e15982697a7

>Note: These two problems are quite *minor* edge-cases, I don't think there is a need to handle them in this PR - as I do not believe this PR is relevant to fix these issues - nonetheless I wanted to mention them as I see them more openly when I make this fix. 

## AI Disclaimer
I used ChatGPT 5.5 (High) to help investigate the issue. The final solution was determined and implemented manually.

## Checklist

- [X] This PR links to a GitHub Discussion for the proposed code change.
- [X] Coverage stays at 100%.
- [X] The documentation explains the change if needed.

### DIFF
--- a/docs/css/termynal.css
+++ b/docs/css/termynal.css
@@ -66,6 +66,10 @@ button[data-terminal-control] {
     font: inherit;
 }
 
+button[data-terminal-control]:hover {
+    color: var(--color-text);
+}
+
 [data-ty] {
     display: block;
     line-height: 2;
--- a/docs/css/termynal.css
+++ b/docs/css/termynal.css
@@ -55,11 +55,15 @@
     text-align: center;
 }
 
-a[data-terminal-control] {
+button[data-terminal-control] {
     cursor: pointer;
     text-align: right;
     display: block;
+    width: 100%;
     color: #aebbff;
+    border: 0;
+    background: transparent;
+    font: inherit;
 }
 
 [data-ty] {
--- a/docs/css/termynal.css
+++ b/docs/css/termynal.css
@@ -56,6 +56,7 @@
 }
 
 a[data-terminal-control] {
+    cursor: pointer;
     text-align: right;
     display: block;
     color: #aebbff;
--- a/docs/js/termynal.js
+++ b/docs/js/termynal.js
@@ -133,7 +133,7 @@ class Termynal {
             this.container.innerHTML = ''
             this.init()
         }
-        restart.href = '#'
+        restart.href = 'javascript:void(0)'
         restart.setAttribute('data-terminal-control', '')
         restart.innerHTML = "restart ↻"
         return restart
@@ -147,7 +147,7 @@ class Termynal {
             this.typeDelay = 0
             this.startDelay = 0
         }
-        finish.href = '#'
+        finish.href = 'javascript:void(0)'
         finish.setAttribute('data-terminal-control', '')
         finish.innerHTML = "fast →"
         this.finishElement = finish
--- a/docs/js/termynal.js
+++ b/docs/js/termynal.js
@@ -133,7 +133,7 @@ class Termynal {
             this.container.innerHTML = ''
             this.init()
         }
-        restart.href = 'javascript:void(0)'
+
         restart.setAttribute('data-terminal-control', '')
         restart.innerHTML = "restart ↻"
         return restart
@@ -147,7 +147,7 @@ class Termynal {
             this.typeDelay = 0
             this.startDelay = 0
         }
-        finish.href = 'javascript:void(0)'
+
         finish.setAttribute('data-terminal-control', '')
         finish.innerHTML = "fast →"
         this.finishElement = finish
--- a/docs/js/termynal.js
+++ b/docs/js/termynal.js
@@ -130,8 +130,7 @@ class Termynal {
         const restart = document.createElement('button')
         restart.type = 'button'
 
-        restart.onclick = (e) => {
-            e.preventDefault()
+        restart.onclick = () => {
             this.container.innerHTML = ''
             this.init()
         }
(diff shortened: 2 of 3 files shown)


phalberg:
To have same functionality as a button we want a hover:
- [X] Add hover functionality (`f365d64` should handle this case. )

phalberg:
Ready from my side, the solution seems to have fixed the major problems!
## Note
These "elements" `fast` and `restart` should perhaps be handled as true buttons sometime, but I will leave that for now, to just fix the current issues in this PR.

YuriiMotov:
> These "elements" `fast` and `restart` should perhaps be handled as true buttons sometime, but I will leave that for now, to just fix the current issues in this PR.

@phalberg, have you evaluated how mush changes would introduce turning them into real buttons?

phalberg:
> > These "elements" `fast` and `restart` should perhaps be handled as true buttons sometime, but I will leave that for now, to just fix the current issues in this PR.
> 
> @phalberg, have you evaluated how mush changes would introduce turning them into real buttons?

Hey @YuriiMotov, thanks for asking.

I haven’t evaluated the full impact of that change yet, but I can take a closer look and see if there are side-effects, etc then report back with what I find.

YuriiMotov:
> I haven’t evaluated the full impact of that change yet, but I can take a closer look and see if there are side-effects, etc then report back with what I find.

That would be great! Thanks!
No need to put much efforts if it requires complex changes, the idea is just to check how difficult it is

phalberg:
> > I haven’t evaluated the full impact of that change yet, but I can take a closer look and see if there are side-effects, etc then report back with what I find.
> 
> That would be great! Thanks! No need to put much efforts if it requires complex changes, the idea is just to check how difficult it is

No worries!
I got some time tomorrow, so will probably report back then, and have a look.
Thank you.

phalberg:
> > These "elements" `fast` and `restart` should perhaps be handled as true buttons sometime, but I will leave that for now, to just fix the current issues in this PR.
> 
> @phalberg, have you evaluated how mush changes would introduce turning them into real buttons?

Hey again @YuriiMotov.
I evaluated the implementation of buttons and I think its worth changing to so I added it as 09ede97f21226d6530a00cfdca155ce51681276b, have a look whenever you guys have a chance. If you are unsure about my implementation or need any adjustments, please let me know, happy to answer and adjust!
The changes were fairly small, and I can no longer reproduce one of the minor UI bugs I had previously. I updated the PR description accordingly.
The biggest win I see in this is that these controls now behave as actual buttons, hopefully preventing future issues related to href.

docs/js/termynal.js
```diff
@@ -127,27 +127,31 @@ class Termynal {
...
         return restart
     }
 
     generateFinish() {
-        const finish = document.createElement('a')
+        const finish = document.createElement('button')
+        finish.type = 'button'
+
         finish.onclick = (e) => {
             e.preventDefault()
```
YuriiMotov:
```suggestion
finish.onclick = () => {
```

`preventDefault()` is now redundant

docs/js/termynal.js
```diff
@@ -127,27 +127,31 @@ class Termynal {
     }
 
     generateRestart() {
-        const restart = document.createElement('a')
+        const restart = document.createElement('button')
+        restart.type = 'button'
+
         restart.onclick = (e) => {
             e.preventDefault()
```
YuriiMotov:
```suggestion
restart.onclick = () => {
```

preventDefault() is now redundant

docs/css/termynal.css
```diff
@@ -55,10 +55,15 @@
...
+button[data-terminal-control] {
+    cursor: pointer;
     text-align: right;
     display: block;
+    width: 100%;
     color: #aebbff;
+    border: 0;
+    background: transparent;
+    font: inherit;
 }
```
YuriiMotov, line 68:
```suggestion

button[data-terminal-control]:hover {
    color: #fff;
}

```

We can make it a bit more interactive)

Not sure about the exact color

YuriiMotov (commented):
LGTM!

Suggested a couple of improvements

YuriiMotov (approved):

docs/js/termynal.js
phalberg: Thanks for catching this, I forgot about these ones!
Fixed in a67d1a9a3b503d7507cb71944fc1cb3c3f393485

docs/js/termynal.js
phalberg: _See previous comment_
Fixed in a67d1a9a3b503d7507cb71944fc1cb3c3f393485

docs/css/termynal.css
phalberg: Great suggestion! I added this in ef44dbe9c9fb1125e6be815b221b5eabf67b66eb, it readjusts hover color after terminals text color, which in this case is white.

phalberg:
> LGTM!
> 
> Suggested a couple of improvements

Thank you for great suggestions along the way and good help!

tiangolo (approved):
Awesome, thank you! 🚀 

And thanks @YuriiMotov for all the help. 🙌

PR #1932, opened 2026-08-16
DOC: fix comment grammar in typer/completion.py

adivx:
Small documentation/comment grammar fix in `typer/completion.py`.

Removed a dangling leading `And` from a comment in `shell_complete` so the comment list reads cleanly:

```
# To add extra error messages, for compatibility with Typer in previous versions
```

Docs/typo only — no functional changes.


svlandeg (commented):
We'll pass on this, thanks.

PR #1942, opened 2026-08-28
♻️ Create `exceptions` module and `TyperException` base class

svlandeg:
## Pull Request

Discussion: https://github.com/fastapi/typer/discussions/1832

## Description

Continuing from the original [Click vendoring](https://github.com/fastapi/typer/pull/1774), this PR introduces a new base class `TyperException` that all other exception classes inherit from. This is to replace a check for `ClickException` (which some users may want to do, cf discusssion thread) which is currently in the private module `_click` and shouldn't be used/imported directly. The "old" Click exception hierarchy will be refactored in the near future and shouldn't be used directly either.

`Abort` and `Exit` have always been part of the public Typer API and as such they are moved into the new Typer module.

## AI Disclaimer

No AI used, just good old fashioned human communication & thinking. Refreshing :sunglasses: 

## Checklist

- [x] Coverage stays at 100%.
- [x] The documentation explains the change if needed.

### DIFF
--- a/typer/__init__.py
+++ b/typer/__init__.py
@@ -5,9 +5,7 @@ __version__ = "0.27.1"
 from shutil import get_terminal_size as get_terminal_size
 
 from . import colors as colors
-from ._click.exceptions import Abort as Abort
 from ._click.exceptions import BadParameter as BadParameter
-from ._click.exceptions import Exit as Exit
 from ._click.termui import confirm as confirm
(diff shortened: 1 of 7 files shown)


tiangolo (approved):
Nice, thank you! 🚀 

This will be available in Typer 0.27.2 in the next few hours.

PR #1946, opened 2026-08-31
Write generated docs as UTF-8 in typer utils docs --output

lenamonj:
Closes #1881.

`typer <app> utils docs --output FILE` writes the generated Markdown with `output.write_text(clean_docs)`, which goes through the platform's default encoding. On interpreters whose locale encoding is not UTF-8 (cp1252 on Windows being the common case), any help text holding a character that encoding cannot represent raises `UnicodeEncodeError` - reproduced today on Linux with `LC_ALL=C PYTHONUTF8=0` against the emoji example from #1881.

The fix is the one #1881 asks for: write the file as UTF-8 explicitly, which matches how the docs are meant to be consumed (Markdown tooling assumes UTF-8) and makes the output independent of the machine that generated it.

Includes a regression test that runs `utils docs --output` in a subprocess under a forced ASCII locale and asserts the file is written and round-trips the non-ASCII help text.

### DIFF
--- a/tests/assets/cli/non_ascii_help.py
+++ b/tests/assets/cli/non_ascii_help.py
@@ -1,0 +1,9 @@
+import typer
+
+app = typer.Typer()
+
+
+@app.command()
+def greet(name: str = "World"):
+    """Say hi to someone ✨ with café flair."""
+    typer.echo(f"Hello {name}")  # pragma: no cover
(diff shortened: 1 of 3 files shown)


svlandeg (commented):
Closing, violates user contribution guidelines.

> Closes https://github.com/fastapi/typer/pull/1881.

1881 is a PR, you don't close a PR with another PR, you review and discuss the first one rather than opening duplicates and exponentially increasing our work load. 

Thanks.

lenamonj:
I mistook #1881 for an issue rather than an open PR. That's my error, I apologize.

Nothing further needed from my side - #1881 already carries the fix. Sorry
for the noise.

PR #1951, opened 2026-09-02
✅ Fix flaky `app_dir` and `launch` tutorial tests under `xdist`

YuriiMotov:
## Description

I noticed that `test-redistribute` failed on `master` (see [logs](https://github.com/fastapi/typer/actions/runs/33624907165)), but passed again on later commits.

After investigating it with Claude Code we found a race between `xdist` workers. Three tests use the real `~/.config/my-super-cli-app/config.json`:

* `test_cli_config_doesnt_exist` expects the file to not exist
* `test_cli_config_exists` creates it in a fixture
* `test_cli` in `test_launch/test_tutorial002.py` creates it via the tutorial code

`xdist` spreads tests across workers, so these tests run in parallel from time to time and then may fail.

To fix this I added an `app_dir` fixture in `tests/conftest.py`. It patches `typer.get_app_dir` to return `tmp_path`, so each test gets its own empty dir.
As a bonus, tests no longer create files in the real `~/.config` on developers' machines.

To reproduce the flakiness on `master`:

```
for i in $(seq 1 15); do pytest tests/test_tutorial/test_app_dir tests/test_tutorial/test_launch -n 4 -q -o addopts="" | grep failed; done
```

## AI Disclaimer

Used Claude Code (Fable 5.1) to investigate and apply the fix. Reviewed manually

## Checklist

- [ ] This PR links to a GitHub Discussion for the proposed code change.
- [x] Coverage stays at 100%.

### DIFF
--- a/tests/test_types_file.py
+++ b/tests/test_types_file.py
@@ -9,7 +9,7 @@ from typer._click._compat import get_best_encoding, should_strip_ansi
 from typer._click.utils import PacifyFlushWrapper
 from typer.testing import CliRunner
 
-from tests.utils import needs_linux, needs_windows
+from tests.utils import needs_linux, needs_macos, needs_windows
 
 app = typer.Typer()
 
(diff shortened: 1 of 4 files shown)


svlandeg:
@YuriiMotov: can you look into the failing test suite? :pray:

YuriiMotov:
> @YuriiMotov: can you look into the failing test suite? 🙏

Sure! :) Sorry, I missed that..

As we patch `get_app_dir`, it's now not executed on Linux and MacOS (not all of its branches are executed), and the coverage drops.
To fix it we can add tests that execute these paths: See commit [834aa11](https://github.com/fastapi/typer/pull/1951/commits/834aa11c42e879c66bcd50bb72181939d963f6aa)

tests/test_types_file.py
```diff
@@ -302,6 +302,33 @@ def test_app_dir_force_posix(monkeypatch: pytest.MonkeyPatch) -> None:
     assert typer.get_app_dir("My App", force_posix=True) == "/home/tester/.my-app"
 
 
+@needs_linux
```
svlandeg, line 305: I'm not sure I fully understand how these three tests are related to the behaviour that this PR fixes?

YuriiMotov: This PR fixes flaky tests (race condition between xdist workers executing tests that use the same app directory). And the proposed fix is to mock the `typer.get_app_dir()`. They are now using their own temporary directories.

But then the coverage drops because we don't execute all branches of `typer.get_app_dir()` anymore.
These 3 tests are here to improve the coverage

svlandeg (approved):
LGTM!

tiangolo (approved):
Nice, thank you! 🙌

PR #1957, opened 2026-09-14
🐛 Ensure that the default option name is not changed by setting a metavar

svlandeg:
## Pull Request

Discussion: https://github.com/fastapi/typer/discussions/1900

## Description

Metavar annotations aren't supposed to influence the display/usage of the option name, as demonstrated by the test `test_parameter_metavar` which ensures that `name: Annotated[str, typer.Option(metavar="CUSTOM")` is called with `--name`. 

A recent PR https://github.com/fastapi/typer/pull/1863 (released with 0.27.0) introduced a small regression on this behaviour, where a metavar corresponding to the option name (if both were lower-cased) could actually overwrite the casing. In retrospect, and after studying the reports & arguments in #1900, I think this was a bug/mistake on my end.

This PR fixes things to make the rule consistent: metavars are only for display (as already documented) and should never influence how an option is used.

⚠️ Note that this is breaking behaviour compared to `0.27.0` ⚠️
(while being less breaking with the original `0.26.8` behaviour)

## AI Disclaimer

Used Cursor as a sparring partner to check my understanding & reasoning, we agreed on all edits in this PR :slightly_smiling_face: 

## Checklist

- [x] Coverage stays at 100%.
- [x] The documentation explains the change if needed.


tests/test_core.py
```diff
@@ -413,7 +413,7 @@ def main(
             "5",
             "--arg6",
             "6",
-            "--ARG9",
+            "--arg9",
```
svlandeg, line 416: I actually think this was the originally intended behaviour of the test, as the print-out on L399 has `arg9`, not `ARG9` (while `ARG4` upper-cased does make sense)

